PR #40, opened 2021-07-16
Initial Stargate support (wip)

### LINKED ISSUE #5: Create initial CRD and controller for Stargate
For this ticket we want to introduce a minimal CRD and initial controller. Assume a Stargate instance per CassandraDatacenter by default for now.

We can worry about filling out the CRD in follow up tickets.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-581) by [Unito](https://www.unito.io)

adutra:
**What this PR does**:
Adds initial support for Stargate CRD and controller.

Fixes #5 

**Checklist**
- [x] Changes manually tested


api/v1alpha1/stargate_types.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// ContainerImage is the image characteristics to use for Stargate containers. Leave nil to use
+	// a default image.
+	// +optional
+	ContainerImage *ContainerImage `json:"containerImage,omitempty"`
+
+	// InitContainerImage is the image characteristics to use for Stargate init containers. Init
+	// containers are used to wait for Cassandra. Leave nil to use a default image.
+	// +optional
+	InitContainerImage *ContainerImage `json:"initContainerImage,omitempty"`
```
jsanda, line 59: Keep in mind that a pod can have multiple init containers. If this is for the init container we use in the Helm, then I would include that in the field name. With that said, we won't need that init container :) The purpose of the init container we deploy with the chart is to wait to deploy Stargate until the Cassandra cluster is up. The K8ssandraCluster controller can do this. It will be the one that creates the Stargate and Reaper objects. It should wait to deploy them until the Datacenter is up. Of course, both Stargate and Reaper can be deployed without a K8ssandraCluster, so I would include logic in the Stargate controller to check that the datacenter is ready before deploying Stargate.

api/v1alpha1/stargate_types.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// ReadinessProbe sets the Stargate readiness probe. Leave nil to use defaults.
+	// +optional
+	ReadinessProbe *corev1.Probe `json:"readinessProbe,omitempty"`
+
+	// Affinity is the affinity to apply to the Stargate pods. See
+	// https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+	// for background.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
```
jsanda, line 88: Based on prior discussions with @adejanovski my understanding is that we want to deploy Stargate nodes in each rack. Affinity for Cassandra pods is configured implicitly via racks. I think we will want piggy back off of that. By default, I think we should look at the rack configuration of the CassandraDatacenter and configure affinity such that we will get Stargate pods distributed evenly among racks (as best we can). This goes beyond the scope of this PR, but I wanted to at least point it out here.

jsanda (commented):
I haven't made it through all the code yet but it looks good!

controllers/stargate_controller.go
```diff
@@ -0,0 +1,340 @@
...
+	err := r.Get(ctx, req.NamespacedName, stargate)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Stargate resource not found. Ignoring since object must be deleted")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Failed to get Stargate")
+		return ctrl.Result{}, err
+	}
+
```
jsanda, line 68: Before doing anything else you should do `stargate = stargate.DeepCopy()`. Reads, i.e., `Get()` hit a shared cached. We don't want to modify objects in the cache.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,340 @@
...
+	err = r.Get(ctx, types.NamespacedName{Name: stargate.Name, Namespace: stargate.Namespace}, found)
+	if err != nil && errors.IsNotFound(err) {
+
+		// FIXME is this correct?
+		cassdc := &cassdcapi.CassandraDatacenter{}
+		dcKey := client.ObjectKey{Namespace: req.Namespace, Name: stargate.Spec.DatacenterName}
+		err := r.Get(ctx, dcKey, cassdc)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				logger.Info("CassandraDatacenter resource not found. Ignoring since object must be deleted")
```
jsanda, line 80: It could also be that the CassandraDatacenter has not been created.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		pullPolicy = corev1.PullIfNotPresent
+	} else {
+		image = fmt.Sprintf("%s/%s:%s", containerImage.Registry, containerImage.Repository, containerImage.Tag)
+		pullPolicy = containerImage.PullPolicy
+	}
+
+	var initImage string
+	var initPullPolicy corev1.PullPolicy
+	initContainerImage := stargate.Spec.InitContainerImage
```
jsanda, line 151: As I mentioned in an earlier comment, we won't need this init container. The Stargate controller can wait to create the Stargate Deployment until the CassandraDatacenter is ready.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,340 @@
...
+			if errors.IsNotFound(err) {
+				logger.Info("CassandraDatacenter resource not found. Ignoring since object must be deleted")
+				return ctrl.Result{}, nil
+			}
+			logger.Error(err, "Failed to get CassandraDatacenter")
+			return ctrl.Result{}, err
+		}
+
+		deployment := r.newStargateDeployment(stargate, cassdc)
+		logger.Info("Creating a new Stargate Deployment", "Deployment.Namespace", deployment.Namespace, "Deployment.Name", deployment.Name)
```
jsanda, line 88: A common pattern is to create the object, compute the hash, and store it in an annotation. Then when the object already exists, you can simply compare hashes to see if there are changes. You can find an example of this in k8ssandracluster_controller.go. You will also find this in cass-operator with the StatefulSets that it creates.

jsanda (changes requested):

api/v1alpha1/stargate_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// queries.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=1
+	Size int32 `json:"size"`
+
+	// ServerVersion is the version of Stargate to deploy.
+	// +kubebuilder:validation:Pattern=\d+\.\d+\.\d+
+	// +kubebuilder:default="1.0.29"
+	// FIXME can we use type container.Version or version.Version here?
+	ServerVersion string `json:"serverVersion"`
```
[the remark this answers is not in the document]
adutra: You are right that this field is only relevant if `ContainerImage` is nil, which does feel a bit like a code smell. I agree to remove it.

controllers/stargate_controller.go
adutra: Ok I will. That said, the controller doesn't change the `Stargate` object, it only updates the `Deployment` object. But that might change in the future I guess, so better safe than sorry.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,335 @@
...
+		return ctrl.Result{}, err
+	}
+	cassdc = cassdc.DeepCopy()
+
+	// Wait until the DC is ready
+	if !cassandra.DatacenterReady(cassdc) {
+		logger.Info("Waiting for datacenter to become ready", "CassandraDatacenter", dcKey)
+		return ctrl.Result{RequeueAfter: 15 * time.Second}, nil
+	}
+
```
jsanda, line 90: We also need to create the Service object for Stargate.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,335 @@
...
+	var livenessProbe *corev1.Probe
+	if stargate.Spec.LivenessProbe == nil {
+		livenessProbe = &corev1.Probe{
+			TimeoutSeconds:      10,
+			InitialDelaySeconds: 30,
+			FailureThreshold:    5,
+		}
+	} else {
+		livenessProbe = stargate.Spec.LivenessProbe.DeepCopy()
```
jsanda, line 216: I don't think we need the DeepCopy `stargate` is already a copy.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,335 @@
...
+
+	var readinessProbe *corev1.Probe
+	if stargate.Spec.ReadinessProbe == nil {
+		readinessProbe = &corev1.Probe{
+			TimeoutSeconds:      10,
+			InitialDelaySeconds: 30,
+			FailureThreshold:    5,
+		}
+	} else {
+		readinessProbe = stargate.Spec.ReadinessProbe.DeepCopy()
```
jsanda, line 227: I don't think we need the DeepCopy `stargate` is already a copy.

controllers/stargate_controller.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+func deploymentName(cassdc *cassdcapi.CassandraDatacenter) string {
+	// FIXME is this correct?
+	//  {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
+	return cassdc.Spec.ClusterName + "-" + cassdc.Name + "-stargate"
+}
+
+// newStargateDeployment returns a Deployment object for Stargate
+func (r *StargateReconciler) newStargateDeployment(stargate *api.Stargate, cassdc *cassdcapi.CassandraDatacenter) *appsv1.Deployment {
```
jsanda, line 150: I would make `newStargateDeployment` a stand alone function instead of a method of StargateReconciler. This will make it a lot easier for writing unit tests.

test/framework/e2e_framework.go
```diff
@@ -269,6 +270,38 @@ func (f *E2eFramework) DeleteDatacenters(namespace string, timeout, interval tim
 	})
 }
 
+// DeleteStargates deletes all Stargates in namespace in all remote clusters.
+// This function blocks until all pods from all Stargates have terminated.
+func (f *E2eFramework) DeleteStargates(namespace string, timeout, interval time.Duration) error {
+	f.logger.Info("deleting all Stargates", "Namespace", namespace)
+
+	for _, remoteClient := range f.remoteClients {
```
jsanda, line 278: We are starting to see a pattern across multiple functions that involve performing some action across cluster. Then poll across all clusters until condition is met. Maybe we can extract some of the control flow parts into separate functions. Doesn't have to be done here in this PR. Just thinking out loud...

test/e2e/suite_test.go
```diff
@@ -22,8 +22,8 @@ func TestOperator(t *testing.T) {
 
 	ctx := context.Background()
 
-	//t.Run("CreateSingleDatacenterCluster", e2eTest(ctx, "single-dc", createSingleDatacenterCluster))
-	t.Run("CreateMultiDatacenterCluster", e2eTest(ctx, "multi-dc", createMultiDatacenterCluster))
+	t.Run("CreateSingleDatacenterCluster", e2eTest(ctx, "single-dc", createSingleDatacenterCluster))
```
jsanda, line 25: After thinking about it some more I do think it would be worthwhile to have some e2e test coverage for deployment Stargate on its own without a parent K8ssandraCluster. We still need a CassandraDatacenter though. It allows us to get the full test coverage of the Stargate controller without the complexity and overhead of the multi-cluster stuff. These tests would only require a single k8s cluster and could be run with the free tier GH Actions runners, which is a big plus.

wdyt?

jsanda (changes requested):

test/e2e/suite_test.go
jsanda: Of course, we should have no trouble running a multi-cluster env with only a single cluster with the free tier runner in GH Actions.

jsanda (changes requested):
I rebased your branch on top of my branch for #50 to do some testing. My work is here https://github.com/jsanda/k8ssandra-operator/tree/stargate-test. I am seeing this error after the first Stargate instance is deployed:

`2021-07-26T02:04:26.344Z	ERROR	controller-runtime.manager.controller.k8ssandracluster	Reconciler error	{"reconciler group": "k8ssandra.io", "reconciler kind": "K8ssandraCluster", "name": "test", "namespace": "multi-dc-tns82l", "error": "stargates.k8ssandra.io \"test-dc1-stargate\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update"}`

It also looks like the resource annotation on the Stargate object is lost along the way.

controllers/k8ssandracluster_controller.go
```diff
@@ -141,6 +143,59 @@ func (r *K8ssandraClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 					return ctrl.Result{}, err
 				}
 			}
+
+			if dcTemplate.Stargate != nil {
```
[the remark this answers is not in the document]
adutra: Ah right, I forgot that possibility. OK to leave it up to another PR.

adutra:
> I am seeing this error after the first Stargate instance is deployed [...] metadata.resourceVersion: Invalid value: 0x0: must be specified for an update

I have not seen this error in my own branch, but I admit that the error pops up in yours. My understanding is that these two lines are maybe doing more harm than good:

```
actualStargate = actualStargate.DeepCopy()
desiredStargate.DeepCopyInto(actualStargate)
...
remoteClient.Update(ctx, actualStargate)					
```

I changed this to:

```
desiredStargate.SetResourceVersion(actualStargate.GetResourceVersion())
...
remoteClient.Update(ctx, desiredStargate)					
```

And the error disappears. Do you know what's the preferred idiom to update an existing `actual` resource so that it reflects the `desired` one? 

> It also looks like the resource annotation on the Stargate object is lost along the way.

I haven't been able to reproduce the situation where the resource hash annotation is lost during updates.

jsanda:
>And the error disappears. Do you know what's the preferred idiom to update an existing actual resource so that it reflects the desired one?

Great question. See [here](https://github.com/k8ssandra/cass-operator/blob/master/pkg/reconciliation/reconcile_racks.go#L244) in cass-operator where it uses the `DeepCopyInto` function. It also calls `SetResourceVersion`.

jsanda (approved):
@adutra please feel free to merge when you are ready. And of course I am happy review/test more before merging if you want. Considering we are still in a prototyping phase and that I have done testing with you branch, I think we are good to merge to address changes in follow up PRs.

adutra:
@jsanda I just finished rebasing and re-testing everything, so I'm OK too to merge now. Among the topics we discussed for follow-up PRs:

# Ability to delete Stargate resources from k8ssandracluster controller;
# Revisit / fix Stargate test on multi-cluster scenario.

Should I go ahead and create tickets for the above topics?

PR #54, opened 2021-07-27
test using self-hosted runner

jsanda:
**What this PR does**:

Fixes #<issue number>

**Checklist**

* [ ] Changes manually tested
* [ ] Automated Tests added/updated
* [ ] Documentation added/updated
* [ ] CHANGELOG.md updated (not required for documentation PRs)
* [ ] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

┆Issue is synchronized with this [Jiraserver Task](https://k8ssandra.atlassian.net/browse/K8SSAND-751) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-751
┆Priority: Medium

### DIFF
--- a/.github/workflows/kind_multicluster_e2e_tests.yaml
+++ b/.github/workflows/kind_multicluster_e2e_tests.yaml
@@ -19,4 +19,4 @@ jobs:
     needs: build_image
     steps:
       - name: Provision kind clusters
-        run: ./scripts/setup-kind-multi-cluster.sh --clusters 2
+        run: ./scripts/setup-kind-multicluster.sh --clusters 2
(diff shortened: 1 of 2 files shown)


.github/workflows/kind_multicluster_e2e_tests.yaml
```diff
@@ -0,0 +1,22 @@
+name: Run kind e2e multi-cluster tests
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [ main ]
```
burmanm, line 7: Do we dare to run on PRs without security? Perhaps we should limit self-hosted to main merges only until we have some security checks?

jsanda: Considering we will probably have plenty of PRs for which we will want to run multi-cluster tests I would like to run them on PRs. With that said, security is certainly a big concern. What about manually kicking off the workflow? If the PR is from a maintainer, I think we would be fine with running.

burmanm: Yes, certainly a possibility. Just something to prevent automated run of unknown PRs not from our maintainers.

PR #55, opened 2021-07-28
Client cache initialization during operator startup

burmanm:
**What this PR does**:
Initializes all the remoteClients at the start of the operator instead of dynamically while trying to access them the first time.

Fixes #<issue number>

**Checklist**

* [ ] Changes manually tested
* [x] Automated Tests added/updated
* [ ] Documentation added/updated
* [ ] CHANGELOG.md updated (not required for documentation PRs)
* [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

┆Issue is synchronized with this [Jiraserver Task](https://k8ssandra.atlassian.net/browse/K8SSAND-750) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-750
┆Priority: Medium


main.go
```diff
@@ -89,12 +91,58 @@ func main() {
...
+		setupLog.Error(err, "unable to create manager")
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+
+	// Fetch ClientConfigs and create the clientCache
 	clientCache := clientcache.New(mgr.GetClient(), scheme)
 
+	cConfigs := k8ssandraiov1alpha1.ClientConfigList{}
```
jsanda, line 103: Wouldn't it be simpler to add volume mounts for the secrets? With or without the ClientConfig CRD the user has to provide the secret.

jsanda:
> @jsanda Includes workaround for current e2e-tests init (since I didn't care to rewrite the kind test scripts today..).

Why would `setup-kind-multicuster.sh` need to rewritten? It creates the kind clusters and generates a kubeconfig file that can be stored in a secret. Don't we need this whether we initialize client connections at start up or at K8ssandraCluster creation?

api/v1alpha1/clientconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ClientConfigSpec defines the desired state of KubeConfig
+type ClientConfigSpec struct {
+	// ContextName allows to override the object name for context-name. If not set, the ClientConfig.Name is used as context name
+	// +kubebuilder:validation:optional
+	ContextName string `json:"contextName,omitempty"`
```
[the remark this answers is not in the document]
burmanm: ContextName in this case refers to the "clusterName", which is not given by anything in the Kubernetes. What I wanted was to allow renaming the cluster for the user, to give them some better known names. "backup", "primary" and then set the DC names to touch them.

main.go
burmanm: Yes, it's simpler if you only wish to mount one secret and nothing else. I don't actually want to build kubeconfigs to always use single secret, but multiple secrets. Also, the idea here was that the same CRD would be usable in the future if we have dynamic cluster selection - one could simply add a new ClientConfig (bad name, I know) item and it could even be reconciled.

api/v1alpha1/clientconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ClientConfigSpec defines the desired state of KubeConfig
+type ClientConfigSpec struct {
```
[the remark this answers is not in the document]
burmanm: To allow defining the secret**s** that we wish to use and to allow some additional parameters (I only added cluster name override, but I guess other connection parameters could be required).

burmanm:
To rewrite it to make a single secret per kubeconfig context and to create those config items.

jsanda:
>To rewrite it to make a single secret per kubeconfig context and to create those config items.

Why do we need/want a secret per context? That seems like extra, unnecessary work.

api/v1alpha1/clientconfig_types.go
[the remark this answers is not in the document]
burmanm: For example, I could imagine as setting parameters such as "AllocatableMemory", CPU restrictions etc which could limit what you can deploy to remote clusters. Or give annotations / labels, such as "aws-region", "oceanic-datacenter". Or why not more restrictions like affinity / tolerations, but to pods. "If dc deployed to us-east-1, don't allow deploying second copy to us-east-2, only us-west-n".

burmanm:
For testing? Yes - unless we want to test how user would use them. Imagine a user wanting to add a single new target cluster, or remove a target cluster. Why would they have to recreate a new kubeconfig file with just those datacenters? How would they even do it? The current setup-kind-multicluster.sh uses ``yq`` to merge multiple files together - doesn't seem very intuitive. Also, what if the operator himsefl can't create access to a cluster, but they would have to be created somewhere else (inside the AWS like we talked about expiring tokens - perhaps creating inside the cloud allows to create permanent tokens?). In that case, one would create the token inside something and the other one somewhere else. 

Unless we provide tools to manage a single secret somewhat sanely, is that really what we want? Does kubectl allow me to bind them together from external and internal sources and not all my defined contexts? What if there are multiple operators? Do they need to download the existing secret and parse it before they can add/remove a new item?

I'm not suggesting we have this thing thought out, but I would imagine a single maintainable secret is going to cause some headaches.

burmanm:
Now reading again a bit.. obviously, you don't have to have multiple secrets - we can parse a single secret for multiple target clusters (perhaps the ClientConfig should be called "TargetCluster" or something like that instead), but I think we should have one TargetCluster/ClientConfig for each target cluster. They just all have "KubeConfigSecret.Name: k8s-contexts" for e2e test.

api/v1alpha1/clientconfig_types.go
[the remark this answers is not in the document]
jsanda: In general I like the idea of additional parameters, but I am not sure about the examples. Wouldn't we want to specify things like affinity and tolerations in the K8ssandraCluster object?

api/v1alpha1/clientconfig_types.go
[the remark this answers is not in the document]
jsanda: >ContextName in this case refers to the "clusterName", which is not given by anything in the Kubernetes.

Makes sense, but the user can also easily set or change the name of a kubeconfig context with `kubectl config rename-context`.

main.go
jsanda: >Yes, it's simpler if you only wish to mount one secret and nothing else.

We can mount multiple secrets with a [projected volume](https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/). 

>I don't actually want to build kubeconfigs to always use single secret, but multiple secrets.

It would be cumbersome to have to create a separate kubeconfig and secret for each cluster. It is easy enough to create a secret with a given kubeconfig. With that said, from a security standpoint I think it would be best if the kubeconfig file includes only those clusters that are intended to be used with the operator. We can provide a script to do this for users. Pass it the name of the contexts you want and it can create a new kubeconfig. We could even have options for renaming the contexts.

jsanda:
> For testing? Yes - unless we want to test how user would use them. Imagine a user wanting to add a single new target cluster, or remove a target cluster. Why would they have to recreate a new kubeconfig file with just those datacenters? How would they even do it? The current setup-kind-multicluster.sh uses `yq` to merge multiple files together - doesn't seem very intuitive. 

I would want and encourage the user to provide a kubeconfig file that only includes those clusters that are intended to be used by the operator. Including additional clusters would be insecure IMO.

`setup-kind-multicluster.sh` merges files together because I found that approach easy enough at the time. We could probably refactor it to just extract the contexts out of `~/.kube/config` (or another kubeconfig file that you specify).

>Also, what if the operator himsefl can't create access to a cluster, but they would have to be created somewhere else (inside the AWS like we talked about expiring tokens - perhaps creating inside the cloud allows to create permanent tokens?). In that case, one would create the token inside something and the other one somewhere else.

In-cluster authentication is normally done via a service account token which is mounted in every pod. I don't think this is any different whether you are running in EKS, GKE, etc.

Out-of-cluster authentication is generally done via a kubeconfig file. A kubeconfig can have a list of user objects. Each user map. to a particular cluster. There can be cloud provider specific auth in those user objects. client-go have various auth plugins for handle those details.

> Unless we provide tools to manage a single secret somewhat sanely, is that really what we want? Does kubectl allow me to bind them together from external and internal sources and not all my defined contexts? What if there are multiple operators? Do they need to download the existing secret and parse it before they can add/remove a new item?

I am definitely not suggesting or advocating that we manage only a single secret. If the user wants to add credentials for additional clusters, he should be able to update an existing secret or create a new secret.

api/v1alpha1/clientconfig_types.go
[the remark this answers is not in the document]
burmanm: There could be multiple K8ssandraCluster objects targetting the same clusters. Lets say there's an allocation for 8 nodes in the target cluster. You could have 2 different K8ssandraClusters which could build a DC there, both taking 3 nodes for example. But a third K8ssandraCluster with 3 nodes could no longer use that datacenter since it's full.

The other defaults could be something like cluster specific storageClass settings. For example, cluster-1 would use storageClass: premium-ssd ; cluster-2 would use storageClass: ultra. When you create the K8ssandraCluster, you could only define "400GB" and let the storageClass be modified by the target cluster's definitions. Obviously could be defined in the K8ssandraCluster each time, but we could simplify things for user (or if there's RBAC issues to make sure that some people are not allowed to define what disks they can use..)

.. or, Ingress/Egress settings if we're not using a single network in the future. Additional labels / annotations per cluster?

main.go
burmanm: > It would be cumbersome to have to create a separate kubeconfig and secret for each cluster.

Is it? Inside GCP it would be something like ``gcloud container clusters get-credentials cluster-name``. 

https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#generate_kubeconfig_entry

I don't really see creating a single file with only the clusters you want any simpler than single file per cluster (which then can be pushed using kubectl create secret --from-file). Given that the user must define somehow the exported secrets instead of everything.

In any case, the the current implementation allows both single one as well as multiple secrets.

api/v1alpha1/clientconfig_types.go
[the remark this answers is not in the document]
burmanm: But also, do we potentially want to keep status of the target cluster? For example to indicate if the connection is down.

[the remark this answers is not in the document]
jsanda: >do we potentially want to keep status of the target cluster?

I don't know. If we do at any point, we probably ought look into https://github.com/kubernetes-sigs/cluster-api.

To close the loop on the discussion around the `ClientConfig` type, you addressed my immediate questions. The rest of the discussion probably should live outside of the PR, maybe the forum ;-)

main.go
jsanda: Thanks for pointing out the `gcloud` command. I looked briefly for EKS and did not find anything similar, but hopefully there is as well as for AKS.

>In any case, the the current implementation allows both single one as well as multiple secrets.

👍

main.go
```diff
@@ -89,12 +90,50 @@ func main() {
...
+	err = mgr.GetClient().List(ctx, &cConfigs)
```
jsanda, line 103: It would be good to put this code and the following for loop into some functions/methods, maybe in clientcache so that we can write some integration tests for it. I am fine doing it in a follow up PR though if need be rather. I want to get this merged sooner rather than later.

main.go
```diff
@@ -89,12 +90,50 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to fetch cluster connections")
+		os.Exit(1)
+	}
+
+	for _, cCfg := range cConfigs.Items {
+		// Create clients and add them to the client cache
+		cfg, err := clientCache.GetRestConfig(&cCfg)
```
jsanda, line 111: As I mentioned in comments elsewhere in the PR, I would to still like the ability to create multiple client connections from a single kubeconfig. I am fine with and actually prefer to address in a follow up PR. Or maybe we find it isn't necessary.

test/framework/e2e_framework.go
```diff
@@ -253,6 +254,42 @@ func (f *E2eFramework) DeployK8sContextsSecret(namespace string) error {
...
+		// Deploy Secrets from here
+		cCfg := api.ClientConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ctx.Cluster,
+				Namespace: namespace,
+			},
+			Spec: api.ClientConfigSpec{
+				ContextName: ctx.Cluster,
+				KubeConfigSecret: corev1.LocalObjectReference{
+					Name: "k8s-contexts",
```
jsanda, line 280: If we are creating a secret per kubeconfig, doesn't `Name` need to be distinct here and match the name of the secrets? I think you need to update the `generateContextsKustomization` function. 

Note that one of the intermediate steps of `setup-kind-multicluster.sh` is to write out the individual kubeconfig files. You want the updated ones intended for in-cluster use which are located at `build/kubeconfigs/updated`.

main.go
```diff
@@ -89,12 +90,50 @@ func main() {
...
+		if err != nil {
+			setupLog.Error(err, "unable to setup cluster connections")
+			os.Exit(1)
+		}
+
+		clientCache.CreateClient(cCfg.GetContextName(), cfg)
```
jsanda, line 117: unhandled error here

test/e2e/suite_test.go
```diff
@@ -101,6 +102,11 @@ func beforeTest(t *testing.T, namespace, fixtureDir string, f *framework.E2eFram
 		return err
 	}
 
+	if err := f.DeployK8sClientConfigs(namespace); err != nil {
```
jsanda, line 105: `f.DeployK8sClientConfigs` fails because the CRDs are deployed in `f.DeployK8ssandraOperator`. You'll need to deploy the CRDs before this call. I would also move up this block:

```go
if err := f.WaitForCrdsToBecomeActive(); err != nil {
		t.Log("failed waiting for CRDs to become active")
		return err
}
```

main.go
```diff
@@ -89,12 +90,50 @@ func main() {
...
+		err = mgr.Add(c)
+		if err != nil {
+			setupLog.Error(err, "unable to add cluster to manager")
+			os.Exit(1)
+		}
+	}
+
+	// Create the reconciler and start it
+
 	if err = (&controllers.K8ssandraClusterReconciler{
```
jsanda, line 137: With these changes in the client cache I think we can get rid of the `SetupMultiClusterWithManager` function and move the watches setup into `SetupWithManager`.

jsanda (changes requested):
We are still going to need a controller to handle events for ClientConfigs, but let's address that in a future ticket/PR.

main.go
burmanm: This PR can still do that. You only need multiple ClientConfigs, but they all can point to the same Secret (which has a single kubeconfig with multiple contexts).

test/framework/e2e_framework.go
burmanm: In this case, the same Secret is used for all the ClientConfigs, that's why the name was hardcoded for e2e-tests.

burmanm:
Addresses some comments, rebased, squashed and what not..

main.go
jsanda: excellent! 👌

jsanda (approved):

main.go
jsanda: >Is it? Inside GCP it would be something like `gcloud container clusters get-credentials cluster-name`.

@burmanm I wanted to circle back to this since I am working on some docs. I don't know why I failed to remember at the time, but that command only adds an entry in your kubeconfig file.

PR #77, opened 2021-08-11
Minor fixes

adutra:
**What this PR does**:
Various fixes ranging from typos to renaming stuff for clarity and changing the types of a few struct fields.

None.

**Checklist**
- [x] Changes manually tested


pkg/clientcache/cache.go
```diff
@@ -85,24 +85,24 @@ func (c *ClientCache) CreateClient(contextName string, restConfig *rest.Config)
...
-	apiConfig, err := c.extractClientCmdFromSecret(namespacedName)
+	apiConfig, err := c.extractClientCmdApiConfigFromSecret(secretKey)
 	if err != nil {
 		return err
 	}
 
-	for _, ctx := range apiConfig.Contexts {
-		clientCfg := clientcmd.NewNonInteractiveClientConfig(*apiConfig, ctx.Cluster, &clientcmd.ConfigOverrides{}, nil)
-		cfg, err := clientCfg.ClientConfig()
+	for ctx, _ := range apiConfig.Contexts {
```
adutra, line 98: I think it's more correct to use the context name and not the cluster name, since that's what `NewNonInteractiveClientConfig` expects to receive. But I might be wrong.

pkg/clientcache/cache.go
```diff
@@ -85,24 +85,24 @@ func (c *ClientCache) CreateClient(contextName string, restConfig *rest.Config)
 }
 
 // CreateRemoteClientsFromSecret is a convenience method for testing purposes
-func (c *ClientCache) CreateRemoteClientsFromSecret(namespacedName types.NamespacedName) error {
+func (c *ClientCache) CreateRemoteClientsFromSecret(secretKey types.NamespacedName) error {
```
adutra, line 88: This method doesn't look in use right now.

pkg/clientcache/cache.go
```diff
@@ -112,19 +112,25 @@ func (c *ClientCache) CreateRemoteClientsFromSecret(namespacedName types.Namespa
...
+	apiConfig, err := c.extractClientCmdApiConfigFromSecret(secretKey)
 	if err != nil {
 		return nil, err
 	}
 
-	clientCfg := clientcmd.NewNonInteractiveClientConfig(*apiConfig, assistCfg.Name, &clientcmd.ConfigOverrides{}, nil)
-	return clientCfg.ClientConfig()
+	clientCmdCfg := clientcmd.NewNonInteractiveClientConfig(
+		*apiConfig,
+		assistCfg.GetContextName(),
```
adutra, line 123: I think it's more correct to call `assistCfg.GetContextName()` instead of accessing `assistCfg.Name`.

pkg/clientcache/cache.go
jsanda: Yeah, I think it can be removed.

pkg/clientcache/cache.go
jsanda: Good catch. I think we got lucky because the context names and cluster names that we use in tests are the same.

pkg/clientcache/cache.go
jsanda: Agreed

jsanda (approved):

PR #85, opened 2021-08-16
Create Stargate deployments per rack

### LINKED ISSUE #17: Create Stargate instance per rack by default
k8ssandra-operator should create a Stargate instance per rack by default. Configuring Stargate should be addressed in other tickets. 

The focus of this ticket should be on:

* Deploying Stargate
* Adding/updating tests
* Updating the status of the K8ssandraCluster

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-600) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-600
┆Priority: Medium

adutra:
**What this PR does**:
Adds support for deploying Stargate in different racks. The Stargate reconciler now creates one deployment per rack, instead of one single deployment for the whole DC.

Fixes #17 

**Checklist**
- [x] Changes manually tested


pkg/cassandra/rack.go
```diff
@@ -0,0 +1,94 @@
...
+	cassdcapi "github.com/k8ssandra/cass-operator/operator/pkg/apis/cassandra/v1beta1"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sort"
+)
+
+const zoneLabel = "failure-domain.beta.kubernetes.io/zone"
+
+func BuildRackAffinity(dc *cassdcapi.CassandraDatacenter, rackName string) *corev1.Affinity {
```
adutra, line 13: This is borrowed from cassdc api for now.

pkg/cassandra/rack.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+	}
+}
+
+func rackAffinityLabels(dc *cassdcapi.CassandraDatacenter, rackName string) map[string]string {
+	var labels map[string]string
+	racks := dc.GetRacks()
+	for _, rack := range racks {
+		if rack.Name == rackName {
+			labels = utils.MergeMap(rack.NodeAffinityLabels, dc.Spec.NodeAffinityLabels)
```
adutra, line 58: Wanted to mention that this comes from cassdc api. As is, DC spec overrides rack spec. This seems in contradiction with K8ssandra's hierarchical principles but I didn't want to change the logic.

pkg/stargate/stargate.go
```diff
@@ -248,18 +32,14 @@ func NewService(sg *api.Stargate, dc *cassdcapi.CassandraDatacenter) *corev1.Ser
 				{Port: 9042, Name: "cassandra"},
 			},
 			Selector: map[string]string{
-				api.StargateLabel: serviceName,
+				api.StargateLabel: stargate.Name,
```
adutra, line 35: I believe this was wrong since the beginning. The selector should match all pods with `StargateLabel` = `stargate.Name`, not `serviceName`.

pkg/stargate/deployments.go
```diff
@@ -0,0 +1,323 @@
...
+
+							Resources: resources,
+
+							Env: []corev1.EnvVar{
+								{Name: "JAVA_OPTS", Value: jvmOptions},
+								{Name: "CLUSTER_NAME", Value: dc.Spec.ClusterName},
+								{Name: "CLUSTER_VERSION", Value: clusterVersion},
+								{Name: "SEED", Value: seedService},
+								{Name: "DATACENTER_NAME", Value: dc.Name},
+								{Name: "RACK_NAME", Value: rack.Name},
```
adutra, line 127: Rack name was previously hard-coded to `default`.

pkg/cassandra/rack.go
jsanda: That's a good point. Since it is the cluster controller who is creating the CassandraDatacenters, we can avoid using the `.Spec.NodeAffinityLabels` property. There is one caveat though. At some point we will need to adding an existing CassandraDatacenter to a K8ssandraCluster, particularly for migration scenarios. Maybe we should raise an issue in cass-operator for changing the behavior.

pkg/stargate/deployments.go
```diff
@@ -0,0 +1,324 @@
...
+	racks := dc.GetRacks()
+
+	averageReplicasPerRack := int32(math.Ceil(float64(stargate.Spec.Size) / float64(len(racks))))
+	replicasRemaining := stargate.Spec.Size
+
+	var deployments = make(map[string]appsv1.Deployment)
+	for _, rack := range racks {
+
+		if replicasRemaining == 0 {
+			// No more room for more Stargate pods, some racks won't have any Stargate pod
```
jsanda, line 42: I feel like we should do something here to let the user know that some racks won't have Stargate pods. Putting something in the operator logs is fine but it doesn't have good visibility. 

There is a broader question that we should consider which is - should we even allow uneven racks?

controllers/stargate_controller.go
```diff
@@ -122,88 +122,120 @@ func (r *StargateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		}
 	}
 
 	// Update status to reflect deployment status
-	if stargate.Status.Replicas != actualDeployment.Status.Replicas ||
-		stargate.Status.ReadyReplicas != actualDeployment.Status.ReadyReplicas ||
-		stargate.Status.UpdatedReplicas != actualDeployment.Status.UpdatedReplicas ||
-		stargate.Status.AvailableReplicas != actualDeployment.Status.AvailableReplicas {
-		ratio := fmt.Sprintf("%v/%v", actualDeployment.Status.ReadyReplicas, stargate.Spec.Size)
+	if stargate.Status.Replicas != replicas ||
```
jsanda, line 221: Now that a Stargate can have multiple Deployments, should the status be changed in some way to reflect that?

I haven't formed an opinion yet but wanted to go ahead and raise the question.

pkg/stargate/stargate.go
adutra: Confirmed that this is now working as expected.

api/v1alpha1/stargate_types.go
```diff
@@ -65,26 +58,85 @@ type StargateTemplate struct {
...
+}
+
+// MergeWith merges this StargateDatacenterTemplate with the given StargateClusterTemplate.
+func (in *StargateDatacenterTemplate) MergeWith(clusterTemplate *StargateClusterTemplate) *StargateDatacenterTemplate {
+	if in == nil && clusterTemplate == nil {
+		return nil
+	} else if in == nil {
+		return &StargateDatacenterTemplate{StargateClusterTemplate: *clusterTemplate}
+	} else {
+		return in
```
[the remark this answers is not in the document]
adutra: It's more of an override than a merge here: if the dc-level template is non nil it will be used, otherwise if the cluster-level template is non nil it will be used, otherwise Stargate won't be deployed at all. 

As discussed offline, this has the merit of being simple and straightforward. But I would see a case for merging templates down; consider the following definition:

```yaml
kind: Stargate
metadata:
  name: s1
spec:
  datacenterRef:
    name: dc1
  size: 3
  heapSize: 512Mi
  nodeSelector:
      - foo: bar
  racks:
    - name: rack1
      nodeSelector:
        - foo: qix
```

Here, rack1 has its own template with a different node selector than the DC-level one. However, I would expect rack1 to also inherit `heapSize:512Mi`, but that's not how it works currently: in rack1, the heap size would be the default one (`256Mi`).

api/v1alpha1/stargate_types.go
```diff
@@ -65,26 +58,85 @@ type StargateTemplate struct {
...
+}
+
+// MergeWith merges this StargateRackTemplate the given StargateDatacenterTemplate.
+func (in *StargateRackTemplate) MergeWith(dcTemplate *StargateDatacenterTemplate) *StargateTemplate {
+	if in == nil && dcTemplate == nil {
+		return nil
+	} else if in == nil {
+		return &dcTemplate.StargateTemplate
+	} else {
+		return &in.StargateTemplate
```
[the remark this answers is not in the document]
adutra: See my other comment here:
https://github.com/k8ssandra/k8ssandra-operator/pull/85#discussion_r692080158

api/v1alpha1/stargate_types.go
```diff
@@ -65,26 +58,85 @@ type StargateTemplate struct {
...
 	// The map should have a key named cassandra_yaml.
 	// +optional
-	CassandraConfigMap *corev1.LocalObjectReference `json:"cassandraConfigMap,omitempty"`
+	CassandraConfigMapRef *corev1.LocalObjectReference `json:"cassandraConfigMap,omitempty"`
+}
+
+// StargateClusterTemplate defines global rules to apply to all Stargate pods in all datacenters in the cluster.
+// These rules will be merged with rules defined at datacenter level in a StargateDatacenterTemplate; dc-level rules
+// have precedence over cluster-level ones.
+type StargateClusterTemplate struct {
```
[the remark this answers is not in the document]
adutra: I don't see any other property that would make sense at cluster level currently. Maybe later we could add some sort of filter on the DC name/labels to limit the Stargate deployments to only the DCs that match?

api/v1alpha1/stargate_types.go
```diff
@@ -65,26 +58,85 @@ type StargateTemplate struct {
...
+	// Size is the number of Stargate instances to deploy in each datacenter. They will be spread evenly across racks.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=1
+	Size int32 `json:"size"`
+}
+
+// StargateDatacenterTemplate defines rules to apply to all Stargate pods in a given datacenter.
+// These rules will be merged with rules defined at rack level in a StargateRackTemplate; rack-level rules
+// have precedence over datacenter-level ones.
+type StargateDatacenterTemplate struct {
```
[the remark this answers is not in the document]
adutra: We could go as far as exposing most of the properties defined in `cassdcapi.CassandraDatacenter`... but I'd suggest that we add them progressively, wdyt?

pkg/stargate/affinity.go
```diff
@@ -0,0 +1,107 @@
...
+
+// computeNodeAffinityLabels computes a set of labels to select worker nodes for the Stargate pods.
+// We use the same labels that were used to select worker nodes for the data pods.
+// See cass-operator, pkg/reconciliation/construct_podtemplatespec.go
+func computeNodeAffinityLabels(dc *cassdcapi.CassandraDatacenter, rackName string) map[string]string {
+	var labels map[string]string
+	racks := dc.GetRacks()
+	for _, rack := range racks {
+		if rack.Name == rackName {
+			labels = utils.MergeMap(rack.NodeAffinityLabels, dc.Spec.NodeAffinityLabels)
```
adutra, line 23: @jsanda my comment got deleted by a force-push so I'm reposting so that we can resume the conversation:

> Wanted to mention that this comes from cassdc api. As is, DC spec overrides rack spec. This seems in contradiction with K8ssandra's hierarchical principles but I didn't want to change the logic.

From @jsanda :

> That's a good point. Since it is the cluster controller who is creating the CassandraDatacenters, we can avoid using the .Spec.NodeAffinityLabels property. There is one caveat though. At some point we will need to adding an existing CassandraDatacenter to a K8ssandraCluster, particularly for migration scenarios. Maybe we should raise an issue in cass-operator for changing the behavior.

I agree to raise an issue, I will take care of it.

But now that you mention this, I wonder if we shouldn't include the following properties in `CassandraDatacenterTemplateSpec`:

* `NodeAffinityLabels map[string]string `
* `NodeSelector map[string]string`
* `Tolerations []corev1.Toleration`
* `AllowMultipleNodesPerWorker bool`

My understanding is that the above properties greatly influence pod scheduling, and I guess that most users would eventually like to see them in the K8ssandraCluster API.

Note: there is also the field `PodTemplateSpec *corev1.PodTemplateSpec` in `CassandraDatacenter` – but atm I think we can avoid exposing it.

controllers/stargate_controller.go
adutra: I was asking myself that question too. I suggest that we take care of it in a follow-up PR, as I would like to revisit the whole logic for updating the Stargate status, and adopt an approach similar to the one you did for `K8ssandraClusterReconciler`.

controllers/stargate_controller.go
```diff
@@ -122,88 +122,120 @@ func (r *StargateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		// Set Stargate instance as the owner and controller
-		if err := ctrl.SetControllerReference(stargate, actualDeployment, r.Scheme); err != nil {
-			logger.Error(err, "Failed to set controller reference on updated Stargate Deployment", "Deployment", deploymentKey)
-			return ctrl.Result{}, err
-		} else if err := r.Update(ctx, actualDeployment); err != nil {
-			logger.Error(err, "Failed to update Stargate Deployment", "Deployment", deploymentKey)
+		if err := ctrl.SetControllerReference(stargate, &desiredDeployment, r.Scheme); err != nil {
+			logger.Error(err, "Failed to set controller reference on new Stargate Deployment", "Deployment", deploymentKey)
 			return ctrl.Result{}, err
+		} else if err := r.Create(ctx, &desiredDeployment); err != nil {
```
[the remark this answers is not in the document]
adutra: Good idea, will do.

adutra:
@k8ssandra/maintainers this is now officially ready to review :-)

scripts/setup-kind-multicluster.sh
```diff
@@ -47,7 +47,17 @@ containerdConfigPatches:
...
-$(for ((i=0; i<$num_workers; i++)); do echo "- role: worker"; done)
+$(for ((i=1; i<=$num_workers; i++)); do
+cat << EOF2
+- role: worker
+  kubeadmConfigPatches:
+    - |
+      kind: JoinConfiguration
+      nodeRegistration:
+        kubeletExtraArgs:
+          node-labels: "rack=rack$i"
```
adutra, line 58: Please note: this change will require you to upgrade kind to its latest version, 0.11.1.

pkg/stargate/deployments.go
adutra: I will add a warning for now, not sure about what we could do more.

> should we even allow uneven racks?

I checked the cass-operator code and it does seem to allow uneven racks, see function `cassdcapi.SplitRacks`. I was also able to execute the following test:

```
func TestUnevenRacks(t *testing.T) {
	rc, _, _ := setupTest()
	rc.Datacenter.Spec.Racks = []api.Rack{{Name: "rack0"}, {Name: "rack1"}, {Name: "rack2"}}
	rc.Datacenter.Spec.Size = 4
	err := rc.CalculateRackInformation()
	assert.NoError(t, err)
	assert.Equal(t, 2, rc.desiredRackInformation[0].NodeCount)
	assert.Equal(t, 1, rc.desiredRackInformation[1].NodeCount)
	assert.Equal(t, 1, rc.desiredRackInformation[2].NodeCount)
}
```

So my suggestion would be that uneven rack should be allowed in Stargate as well.

api/v1alpha1/constants.go
```diff
@@ -3,9 +3,14 @@ package v1alpha1
 const (
 	ResourceHashAnnotation = "k8ssandra.io/resource-hash"
 
-	// StargateLabel is the distinctive label for objects created by the Stargate controller.
+	// StargateLabel is the distinctive label for all objects created by the Stargate controller. The label value is
+	// the Stargate resource name.
```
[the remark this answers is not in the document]
adutra: Thank you! I will add the missing ones.

api/v1alpha1/constants.go
```diff
@@ -3,14 +3,32 @@ package v1alpha1
...
 
+	CreatedByLabel                                = "app.kubernetes.io/created-by"
+	CreatedByLabelValueK8ssandraClusterController = "k8ssandracluster-controller"
+	CreatedByLabelValueStargateController         = "stargate-controller"
+
+	PartOfLabel      = "app.kubernetes.io/part-of"
 	PartOfLabelValue = "k8ssandra"
 
-	K8ssandraClusterLabel = "io.k8ssandra/cluster"
+	K8ssandraClusterLabel = "k8ssandra.io/cluster"
```
adutra, line 23: I noticed this label had an inverted prefix: `io.k8ssandra` instead of `k8ssandra.io`. I fixed it.

api/v1alpha1/constants.go
```diff
@@ -3,14 +3,32 @@ package v1alpha1
 const (
 	ResourceHashAnnotation = "k8ssandra.io/resource-hash"
 
-	// StargateLabel is the distinctive label for objects created by the Stargate controller.
-	StargateLabel = "k8ssandra.io/stargate"
+	NameLabel      = "app.kubernetes.io/name"
+	NameLabelValue = "k8ssandra-operator"
 
-	DefaultStargateVersion = "1.0.30"
+	InstanceLabel = "app.kubernetes.io/instance"
```
adutra, line 9: These 2 labels are declared but not used for now. I'm not sure how to compute a sensible value for them, and this goes beyond the scope of this PR>

controllers/stargate_controller.go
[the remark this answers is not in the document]
adutra: I like the overall idea, but how would you deal with the aggregated errors? Is it ok to return, say, the first one?

pkg/stargate/deployments.go
bradfordcp: IIRC I thought we blocked uneven racks in the validating webhook

api/v1alpha1/stargate_types.go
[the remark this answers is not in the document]
jsanda: >I'd suggest that we add them progressively, wdyt?

agreed

controllers/stargate_controller.go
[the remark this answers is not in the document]
jsanda: >how would you deal with the aggregated errors? 

I think we would need a custom error type whose `Error()` method provides a summary of all the errors. As each error occurs we should log it as well.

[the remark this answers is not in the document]
adutra: Thinking a bit more about this idea, I think It might not be a good idea. Looking at cass-operator code, every error encountered along the (very long) reconciliation process stops the reconcile loop immediately – even when inside a  range loop, for example in `CheckRackStoppedState`.

I would suggest that, in a follow-up PR, we implement something similar to cass-operator's `ReconcileResult` interface. This would greatly simplify the code as we would be able to easily split the reconciliation logic into many small, independent functions.

Once this new utility is available, then maybe we could add support for accumulating multiple errors until we reach the next checkpoint (a point in the reconcile loop where we must stop if there is any error so far). Does that make sense to you?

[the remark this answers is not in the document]
jsanda: @adutra can you clarify what you think might be a bad idea? I am not sure if you are referring to the cass-operator process of failing fast or trying to fail later.

I am certainly in favor of using `ReconcileResult` or something similar if it will help simplify the reconciliation logic.

And I am definitely fine with making the change in a follow-up PR as it is more of a enhancement than anything else.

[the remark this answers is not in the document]
adutra: I'm sorry, what I meant is that it would be a bad idea _if done within this PR_. I do think your idea is good, but I would have to introduce at least some support for accumulating errors and merging `ctrl.Result` instances. What I meant to say is just that it would be better to do it in a follow-up PR. I'd be happy to bootstrap this work as soon as this PR is merged.

scripts/setup-kind-multicluster.sh
```diff
@@ -47,7 +47,17 @@ containerdConfigPatches:
...
-$(for ((i=0; i<$num_workers; i++)); do echo "- role: worker"; done)
+$(for ((i=1; i<=$num_workers; i++)); do
+cat << EOF2
+- role: worker
+  kubeadmConfigPatches:
+    - |
+      kind: JoinConfiguration
+      nodeRegistration:
+        kubeletExtraArgs:
+          node-labels: "topology.kubernetes.io/zone=rack$i"
```
jsanda, line 58: nice addition 👍

pkg/stargate/affinity.go
adutra: Here is the issue: https://github.com/k8ssandra/cass-operator/issues/162.

api/v1alpha1/constants.go
jsanda: Yeah, definitely agree it is outside the scope of the PR. With that said, we do need to firm up our labels.

api/v1alpha1/constants.go
jsanda: Java package naming conventions surfacing from my subconscious 😄

adutra:
Heads up: I rebased on top of main to incorporate the latest changes there.

pkg/stargate/deployments.go
adutra: I don't see anything in cass-operator webhooks preventing uneven racks, but maybe I'm not looking at the right place. In any case, I'd vote for allowing uneven racks in Stargate deployments: it doesn't change much our code and the impacts of uneven Stargate pod distributions seems low to me. I added a message in the logs already if that happens.

api/v1alpha1/constants.go
jsanda: I think every object that the operator creates should have the `K8ssandraClusterLabel` label. @adutra wdyt?

api/v1alpha1/stargate_types.go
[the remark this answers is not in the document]
jsanda: I am in favor of sticking with what we have for now, getting this PR merged, and then start a discussion on the forum to solicit input as to whether merge behavior would be preferred.

api/v1alpha1/stargate_types_test.go
```diff
@@ -0,0 +1,243 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"testing"
+)
+
+var (
+	quantity256Mi = resource.MustParse("256Mi")
+	quantity512Mi = resource.MustParse("512Mi")
+)
+
```
jsanda, line 14: I think some of these tests could be written more succinctly as table-tests.

api/v1alpha1/constants.go
adutra: I agree and I think it is already the case for `CassandraDatacenter` and `Stargate` objects. For `Deployment` and `Service` objects created by the Stargate controller, I think we should not set the label on every object, but only when the corresponding `Stargate` object also has the label. I will give it a try. Wdyt?

api/v1alpha1/stargate_types.go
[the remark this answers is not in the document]
adutra: Sounds good. Once this PR is merged I will compile a list of follow-up actions, which will include creating a few GH issues and hitting the public forum for input on this topic.

controllers/stargate_controller.go
jsanda: Sounds good. I should mention though that I am not sure I am happy with what I did in `K8ssandraClusterReconciler` 😬

controllers/stargate_controller_test.go
```diff
@@ -96,50 +106,245 @@ func testCreate(t *testing.T, testClient client.Client) {
...
+	}, timeout, interval)
+
+	t.Log("check that the Stargate resource is fully reconciled")
+	require.Eventually(t, func() bool {
+		err := testClient.Get(ctx, stargateKey, stargate)
+		return err == nil && stargate.Status.Progress == api.StargateProgressRunning
+	}, timeout, interval)
+
+	t.Log("check Stargate status")
+	require.EqualValues(t, 1, stargate.Status.Replicas, "expected to find 1 replica for Stargate")
```
jsanda, line 153: minor nit - consider the `assert` equivalent functions that would allow for multiple verifications before failing.

pkg/stargate/affinity.go
jsanda: Those properties are definitely going to be added in `CassandraDatacenterTemplateSpec`. I expect some or all of that to be done as part of #87. It

Thanks for creating the ticket!

jsanda (commented):
@adutra I added some more comments but I am good with merging provided the tests pass. Thanks!

api/v1alpha1/constants.go
jsanda: My thinking was that every object created by the operator _is_ part of the K8ssandraCluster by adding the cluster label a user or admin can easily and conveniently query with a simple label selector to get a full inventory of all objects created by the operator for a given K8ssandraCluster. This is something on which we can iterate on and further discuss. I don't want it to hold up the PR.

pkg/stargate/deployments.go
```diff
@@ -138,6 +138,9 @@ func NewDeployments(stargate *api.Stargate, dc *cassdcapi.CassandraDatacenter) m
 				},
 			},
 		}
+		if klusterName, found := stargate.Labels[api.K8ssandraClusterLabel]; found {
```
jsanda, line 141: I like the variable name 😄

adutra: Speaking of which, have you considered adding a short name to the `k8ssandracluster` CRD? Maybe `kluster`?

adutra:
@jsanda I addressed your last comments but I can't merge without at least one approval.

PR #89, opened 2021-08-20
Script to generate kubeconfig secret and ClientConfig

### LINKED ISSUE #83: Add support for service account token authentication to remote clusters
In #82 I reported authentication issues with GKE. The OAuth token expires every hour and Google Cloud Tools is needed to renew it. We will face similar problems with other cloud providers. An alternative solution that is cloud provider-agnostic is to use service account tokens from the remote clusters. An approach was described in this [article](https://goteleport.com/blog/kubectl-gke/).

I propose that we create a script that does the following:

* Get the service account token from the k8ssandra-operator service account in the remote cluster
* Get the CA cert from the k8ssandra-operator service account in the remote cluster
* Create a kubeconfig for the remote cluster
* Create a kubeconfig secret in the control plane cluster
* Create a ClientConfig in the control plane cluster. It should reference the secret.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-799) by [Unito](https://www.unito.io)
┆Epic: Multi-Cluster Deployment
┆Issue Number: K8SSAND-799
┆Priority: Medium

jsanda:
**What this PR does**:
* Gets the service account token from the remote cluster
* Gets the CA cert from the remote cluster
* Creates a kubeconfig file using the token and CA cert
* Create a secret that contains the kubeconfig
* Create a ClientConfig that references the secret

Fixes #83

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/scripts/create-client-config.sh
+++ b/scripts/create-client-config.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+
+set -e
+
+set -o xtrace
+
+OPTS=$(getopt -o h --long src-context:,src-kubeconfig:,dest-context:,dest-kubeconfig:,namespace:,serviceaccount,output-dir:,help -n 'create-client-config' -- "$@")
+
+eval set -- "$OPTS"
+
+function help() {
+    echo
+    echo "Syntax: create-client-config.sh [options]"
+    echo "Options:"
+    echo "src-context     The context for the source cluster that contains the service account."
+    echo "src-kubeconfig   The kubeconfig for the source cluster that contains the service account."
+    echo "dest-context    The context for the cluster where the ClientConfig will be created."
+    echo "dest-kubeconfig  The kubeconfig for the cluster where the ClientConfig will be created."
+    echo "namespace       The namespace in which the service account exists and where the ClientConfig will be created."
+    echo "serviceaccount  The name of the service account from which the ClientConfig will be created. Defaults to k8ssandra-operator."
+    echo "output-dir      The directory where generated artifacts are written. If not specified a temp directory is created."
+}
+
+src_context=""
+src_kubeconfig=""
+dest_context=""
+dest_kubeconfig=""
+service_account="k8ssandra-operator"
+namespace=""
+output_dir=""
+
+while true; do
+  case "$1" in
+    --src-context ) src_context="$2"; shift 2 ;;
+    --src_kubeconfig ) src_kubeconfig="$2"; shift 2 ;;
+    --dest-context ) dest_context="$2"; shift 2 ;;
+    --dest-kubeconfig ) dest_kubeconfig="$2"; shift 2 ;;
+    --namespace ) namespace="$2"; shift 2 ;;
+    --serviceaccount ) service_account="$2"; shift 2 ;;
+    --output-dir ) output_dir="$2"; shift 2 ;;
+    -h | --help ) help; exit;;
+    -- ) shift; break ;;
+    * ) break ;;
+  esac
+done
+
+if [ -z "$src_context" ] && [ -z "$src_kubeconfig" ]; then
+  echo "At least one of the --src-context or --src-kubeconfig options must be specified"
+  exit 1
+fi
+
+src_context_opt=""
+src_kubeconfig_opt=""
+namespace_opt=""
+
+if [ -z "$src_context" ]; then
+  src_context=$(kubectl $src_kubeconfig config current-context)
+else
+  src_kubeconfig=""
+  src_context_opt="--context $src_context"
+fi
+
+if [ ! -z "$src_kubeconfig" ]; then
+  src_kubeconfig_opt="--kubeconfig $src_kubeconfig"
+fi
+
+if [ -z "$output_dir" ]; then
+  output_dir=$(mktemp -d)
+fi
+
+if [ ! -z "$namespace"]; then
+  namespace_opt="-n $namespace"
+fi
+
+sa_secret=$(kubectl $src_kubeconfig_opt $src_context_opt $namespace_opt get serviceaccount $service_account -o jsonpath='{.secrets[0].name}')
+sa_token=$(kubectl $src_kubeconfig_opt $src_context_opt $namespace_opt get secret $sa_secret -o jsonpath='{.data.token}' | base64 -d)
+ca_cert=$(kubectl $src_kubeconfig_opt $src_context_opt $namespace_opt get secret $sa_secret -o jsonpath="{.data['ca\.crt']}")
+
+cluster=$(kubectl $src_kubeconfig_opt config view -o jsonpath="{.contexts[?(@.name == \"$src_context\"})].context.cluster}")
+cluster_addr=$(kubectl $src_kubeconfig_opt config view -o jsonpath="{.clusters[?(@.name == \"$cluster\"})].cluster.server}")
+
+output_kubeconfig="$output_dir/kubeconfig"
+echo "Creating $output_kubeconfig"
+cat > $output_kubeconfig <<EOF
+apiVersion: v1
+clusters:
+- cluster:
+    certificate-authority-data: ${ca_cert}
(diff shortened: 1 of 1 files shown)


scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
...
+
+OPTS=$(getopt -o h --long src-context:,src-kubeconfig:,dest-context:,dest-kubeconfig:,namespace:,serviceaccount,output-dir:,help -n 'create-client-config' -- "$@")
+
+eval set -- "$OPTS"
+
+function help() {
+    echo
+    echo "Syntax: create-client-config.sh [options]"
+    echo "Options:"
+    echo "src-context     The context for the source cluster that contains the service account. This or the src-kubeconfig option must be set."
```
adutra, line 27: Very minor: for a prettier output you could align the start of each description text.

scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
+#
+# k8ssandra-operator should be installed in the remote clusters prior to running this. The
+# script fetches the k8ssandra-operator service account from the remote cluster and
+# extracts the token and CA cert which are then added to a kubeconfig file. The script then
+# creates a secret with the contents of the kubeconfig file. Lastly, the script creates a
+# ClientObject that references the secret.
```
adutra, line 7:
```suggestion
# ClientConfig object that references the secret.
```

scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
...
+  src_kubeconfig_opt="--kubeconfig $src_kubeconfig"
+fi
+
+if [ -z "$output_dir" ]; then
+  output_dir=$(mktemp -d)
+else
+  mkdir -p $output_dir
+fi
+
+if [ ! -z "$namespace"]; then
```
adutra, line 85:
```suggestion
if [ ! -z "$namespace" ]; then
```

scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
...
+  namespace_opt="-n $namespace"
+fi
+
+sa_secret=$(kubectl $src_kubeconfig_opt $src_context_opt $namespace_opt get serviceaccount $service_account -o jsonpath='{.secrets[0].name}')
```
adutra, line 89: Is this procedure supposed to work on any cluster? On my local kind cluster, the service account exists, but has no secrets:

```
kubectl get sa k8ssandra-operator -o json
{
    "apiVersion": "v1",
    "kind": "ServiceAccount",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"ServiceAccount\",\"metadata\":{\"annotations\":{},\"name\":\"k8ssandra-operator\",\"namespace\":\"default\"}}\n"
        },
        "creationTimestamp": "2021-08-20T15:37:14Z",
        "name": "k8ssandra-operator",
        "namespace": "default",
        "resourceVersion": "34052",
        "uid": "982796e5-0a00-4d7d-8a1c-8dd9067d9a13"
    }
}
```

adutra (changes requested):

scripts/create-client-config.sh
jsanda: Good question. I assumed it would but I only tested against GKE so far. My understanding from the [docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) is that the secret is automatically added to the service account. I think that behavior is controlled by an auth plugin. Maybe it is turned off for kind. I'll investigate.

jsanda: @adutra I just deployed the operator in my kind cluster as follows:

`kustomize build config/default | k apply -f -`

And here is the service account:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{},"name":"k8ssandra-operator","namespace":"default"}}
  creationTimestamp: "2021-08-20T16:19:01Z"
  name: k8ssandra-operator
  namespace: default
  resourceVersion: "28259"
  uid: a9ab7ac2-6405-4d13-b824-919589d3c170
secrets:
- name: k8ssandra-operator-token-snqp9
```
I am curious why we are seeing different behavior.

adutra: I did the same. But I just deleted the cluster and re-created everything from scratch, and now the service account has a token. Strange. But let's disregard this incident for now.

adutra: I confirmed this morning that, once the syntax error line 85 is fixed, the script works and the generated secret is OK.

scripts/create-client-config.sh
jsanda: I tried a couple different approaches for formatting/wrapping the text into fixed width columns. Any suggestions on how best to do it?

scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
...
+    echo "Syntax: create-client-config.sh [options]"
+    echo "Options:"
+    echo "src-context     The context for the source cluster that contains the service account. This or the src-kubeconfig option must be set."
+    echo "src-kubeconfig   The kubeconfig for the source cluster that contains the service account. This or the src-context option must be set."
+    echo "dest-context    The context for the cluster where the ClientConfig will be created."
+    echo "dest-kubeconfig  The kubeconfig for the cluster where the ClientConfig will be created."
+    echo "namespace       The namespace in which the service account exists and where the ClientConfig will be created."
+    echo "serviceaccount  The name of the service account from which the ClientConfig will be created. Defaults to k8ssandra-operator."
+    echo "output-dir      The directory where generated artifacts are written. If not specified a temp directory is created."
+}
```
adutra, line 34:
```suggestion
function help() {
cat << EOF
Syntax: create-client-config.sh [options]
Options:
  --src-context <ctx>      The context for the source cluster that contains the service account.
                           This or the src-kubeconfig option must be set.
  --src-kubeconfig <cfg>   The kubeconfig for the source cluster that contains the service account.
                           This or the src-context option must be set.
  --dest-context <ctx>     The context for the cluster where the ClientConfig will be created.
  --dest-kubeconfig <cfg>  The kubeconfig for the cluster where the ClientConfig will be created.
  --namespace <ns>         The namespace in which the service account exists and
                           where the ClientConfig will be created.
  --serviceaccount <name>  The name of the service account from which the ClientConfig will be created.
                           Defaults to k8ssandra-operator.
  --output-dir <path>      The directory where generated artifacts are written.
                           If not specified a temp directory is created.
  --help                   Displays this help message.
EOF
}
```

scripts/create-client-config.sh
adutra: I'm no specialist in these things but I added a suggestion in another comment below:

https://github.com/k8ssandra/k8ssandra-operator/pull/89/files#r693825451

scripts/create-client-config.sh
```diff
@@ -0,0 +1,138 @@
...
+dest_kubeconfig=""
+service_account="k8ssandra-operator"
+namespace=""
+output_dir=""
+
+while true; do
+  case "$1" in
+    --src-context ) src_context="$2"; shift 2 ;;
+    --src_kubeconfig ) src_kubeconfig="$2"; shift 2 ;;
+    --dest-context ) dest_context="$2"; shift 2 ;;
```
adutra, line 48: I don't see `dest_context` or `dest_kubeconfig` being used anywhere below.

scripts/create-client-config.sh
jsanda: 👍

scripts/create-client-config.sh
jsanda: I updated the script.

adutra (approved):

PR #94, opened 2021-08-26
Cassandra configuration [K8SSAND-822][K8SSAND-796]

### LINKED ISSUE #87: Initial C* configuration exposure
**What is missing?**
Building off of the conversation that happened in https://github.com/k8ssandra/k8ssandra-operator/issues/10 translate the discussed approach into an initial implementation exposing a basic set of C* configuration options.

We can be initially restrictive with what can be exposed - the ideal state would be parity with what is available in k8ssandra-1.x if possible.

**Why do we need it?**
To allow end-users (either via kustomize or eventually helm) to be able to configure various C* features.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-822) by [Unito](https://www.unito.io)
┆Epic: Exposing Cassandra Configuration
┆Issue Number: K8SSAND-822
┆Priority: Medium

jsanda:
**What this PR does**:
This PR introduce a framework for configuring CassandraDatacenters and for configuring Cassandra itself. Properties are exposed both at the cluster and dc levels with the latter taking precedence. 

Not all properties and C* settings are exposed yet. Those can be added later in separate PRs. The important thing is having the framework in place for how the properties should be exposed and managed.

For Cassandra configs, I want to introduce merge semantics but we can tackle that in a follow up PR.

Fixes #87

**Checklist**
- [ ] Changes manually tested


api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -129,6 +140,53 @@ type EmbeddedObjectMeta struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
+// TODO Implement Stringer interface. It will helpful for debugging and testing.
+type CassandraConfig struct {
```
adutra, line 144: I'd note that we are handling Cassandra config through an embedded struct, while Stargate config is currently an independent ConfigMap. I don't have any preference, but shouldn't we at least harmonize both to use the same approach?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -129,6 +140,53 @@ type EmbeddedObjectMeta struct {
...
+	CassandraYaml *CassandraYaml `json:"cassandraYaml,omitempty"`
+
+	JvmOptions *JvmOptions `json:"jvmOptions,omitempty"`
+}
+
+type CassandraYaml struct {
```
adutra, line 150: I wonder how future-proof this will be? There might be differences already between C* 3.11 and C* 4.0, and who knows what surprises C* 4.1 and higher will bring wrt to config options... How do you think we would handle these discrepancies?

pkg/cassandra/datacenter.go
```diff
@@ -0,0 +1,141 @@
...
+	Size              int32
+	Resources         *corev1.ResourceRequirements
+	SystemReplication SystemReplication
+	StorageConfig     *cassdcapi.StorageConfig
+	Racks             []cassdcapi.Rack
+	CassandraConfig   *api.CassandraConfig
+	AdditionalSeeds   []string
+}
+
+func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig) (*cassdcapi.CassandraDatacenter, error) {
```
adutra, line 42: I like the variable name ;-)

jsanda:
@burmanm can you please review this. I made the mistake of thinking #111 would have a quick, easy fix and tried to tackle it in this PR. That hasn't been the case so I will get things cleaned up but would like to get wrapped up.

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -76,42 +76,77 @@ type K8ssandraClusterList struct {
...
+	// ServerVersion is the Cassandra version.
+	// +kubebuilder:validation:Pattern=(3\.11\.\d+)|(4\.0\.\d+)
+	ServerVersion string `json:"serverVersion,omitempty"`
+
+	// Resources is the cpu and memory resources for the cassandra container.
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// SystemLoggerResources is the cpu and memory resources for the server-system-logger
+	// container.
+	SystemLoggerResources *corev1.ResourceRequirements `json:"systemLoggerResources,omitempty"`
```
burmanm, line 99: Is this really something we want to make so configurable? Are we going to expose all the init containers and their resource configurations also?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -76,42 +76,77 @@ type K8ssandraClusterList struct {
...
-	// Size is the number of data replicas to deploy in this datacenter.
+	// Size is the number Cassandra pods to deploy in this datacenter.
 	// This number does not include Stargate instances.
 	// +kubebuilder:validation:Minimum=1
 	Size int32 `json:"size"`
 
-	ServerVersion string `json:"serverVersion"`
+	// ServerVersion is the Cassandra version.
+	// +kubebuilder:validation:Pattern=(3\.11\.\d+)|(4\.0\.\d+)
+	ServerVersion string `json:"serverVersion,omitempty"`
```
burmanm, line 131: So if I set ServerVersion to 3.11 here and CassandraClusterSpec to 4.0.0, would I get a cluster with one DC having 3.11 and one with 4.0.0 ?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -76,42 +76,77 @@ type K8ssandraClusterList struct {
...
+	// Cluster is the name of the cluster. This corresponds to cluster_name in
+	// cassandra.yaml.
+	// +kubebuilder:validation:MinLength=2
 	Cluster string `json:"cluster,omitempty"`
 
-	Datacenters []CassandraDatacenterTemplateSpec `json:"datacenters,omitempty"`
+	// ServerImage is the image for the cassandra container. Note that this should be a
+	// management-api image. If left empty the operator will choose a default image based
+	// on ServerVersion.
+	ServerImage string `json:"serverImage,omitempty"`
```
burmanm, line 88: Should probably be +optional (cass-operator's choice of image should be the newest one released for that ServerVersion)

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -30,7 +30,7 @@ import (
 type K8ssandraClusterSpec struct {
 	K8sContextsSecret string `json:"k8sContextsSecret,omitempty"`
```
burmanm, line 31: Shouldn't this be removed?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -76,42 +76,77 @@ type K8ssandraClusterList struct {
...
+	// CassandraConfig is configuration settings that are applied to cassandra.yaml and
+	// jvm-options for 3.11.x or jvm-server-options for 4.x.
+	CassandraConfig *CassandraConfig `json:"config,omitempty"`
+
+	// StorageConfig is the persistent storage requirements for each Cassandra pod. This
+	// includes everything under /var/lib/cassandra, namely the commit log and data
+	// directories.
+	StorageConfig *cassdcapi.StorageConfig `json:"storageConfig,omitempty"`
+
+	Racks []cassdcapi.Rack `json:"racks,omitempty"`
```
burmanm, line 110: StorageConfig is pointer, this isn't a pointer (and both are +optional?). I'd like to see consistency to which one we use.

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -76,42 +76,77 @@ type K8ssandraClusterList struct {
...
+	SystemLoggerResources *corev1.ResourceRequirements `json:"systemLoggerResources,omitempty"`
+
+	// CassandraConfig is configuration settings that are applied to cassandra.yaml and
+	// jvm-options for 3.11.x or jvm-server-options for 4.x.
+	CassandraConfig *CassandraConfig `json:"config,omitempty"`
+
+	// StorageConfig is the persistent storage requirements for each Cassandra pod. This
+	// includes everything under /var/lib/cassandra, namely the commit log and data
+	// directories.
+	StorageConfig *cassdcapi.StorageConfig `json:"storageConfig,omitempty"`
```
burmanm, line 108: If StorageConfig is left out (omit empty), what's the intended behaviour?

api/v1alpha1/k8ssandracluster_types.go
burmanm: I think this is more generalised issue with the config-builder problem we have. And it's not just the CassandraYaml section, but also JvmOptions. JvmOptions depend on the used JDK version and the selected GC.  Those properties do not make much sense when used with ZGC or Shenandoah (both of which we'll want to promote with 4.0)

pkg/cassandra/config.go
```diff
@@ -2,52 +2,162 @@ package cassandra
...
+		//
+		//if len(config.CassandraYaml.Authorizer) > 0 {
+		//	cassandraYaml["authorizer"] = config.CassandraYaml.Authorizer
+		//}
+		//
+		//if len(config.CassandraYaml.RoleManager) > 0 {
+		//
+		//}
+
+		if config.CassandraYaml.ConcurrentReads != nil {
```
burmanm, line 67: Why isn't this just using yaml encoder?

pkg/cassandra/datacenter.go
```diff
@@ -0,0 +1,141 @@
...
+	namespace := template.Meta.Namespace
+	if len(namespace) == 0 {
```
burmanm, line 44: if namespace == "" ?

pkg/cassandra/datacenter.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// TODO Add validation check to ensure StorageConfig is set at the cluster or DC level
+	if dcTemplate.StorageConfig == nil {
+		dcConfig.StorageConfig = clusterTemplate.StorageConfig
+	} else {
+		dcConfig.StorageConfig = dcTemplate.StorageConfig
+	}
+
+	// TODO Do we want merge vs override?
```
burmanm, line 133: Merge.

test/testdata/fixtures/multi-dc/k8ssandra.yaml
```diff
@@ -6,106 +6,77 @@ spec:
   k8sContextsSecret: k8s-contexts
   cassandra:
     cluster: test
+    serverVersion: "3.11.10"
```
burmanm, line 9: 3.11.10 and 3.11.11 image?

api/v1alpha1/k8ssandracluster_types.go
jsanda: Cassandra configs are set in the `CassandraDatacenter.Spec.Config` property. The value is serialized into json and set in an environment variable in the config init container. cass-config-builder is responsible for generating the actual config files. While changing this has been discussed and something we will probably pursue at some point, it is definitely out of scope for the first alpha. This rules providing a ConfigMap for Cassandra configs for the time being.

@adutra @burmanm do we want to have Stargate also use cass-config-builder? On the one hand it gives us consistency. On the other hand, I would argue that it is low priority. Happy to discuss though.

api/v1alpha1/k8ssandracluster_types.go
jsanda: >I wonder how future-proof this will be?

Is it not sufficient to something like this:

```
if version is 3.11.x 
    // apply 3.11.x specific settings
else 
    // apply 4.x specific settings
```

Can't we do something similar for JVM options? If we know the image that is being used then we should know which JVM is being used and should be able to apply settings based on that knowledge.

Right now the goal is just to provide parity with what we offer in the Helm charts.

api/v1alpha1/k8ssandracluster_types.go
jsanda: Yes. This also means you could have one DC at 4.0.0 and one at 4.0.1 and so on.

api/v1alpha1/k8ssandracluster_types.go
jsanda: >StorageConfig is pointer, this isn't a pointer

I am confused by this statement.  The StorageConfig fields in both the CassandraClusterTemplate and the CassandraDatacenterTemplate are pointer so we can easily determine if they are unset.

api/v1alpha1/k8ssandracluster_types.go
jsanda: Validation is needed to ensure that it is set in either the CassandraClusterTemplate or the CassandraDatacenterTemplate (or both).

pkg/cassandra/datacenter.go
jsanda: What is the question? Are you asking why I am checking length vs `namespace == ""`? If so both are idiomatic approaches.

burmanm: With the exception that this one doesn't catch diff between []string and string

pkg/cassandra/datacenter.go
jsanda: I will create a separate ticket to do this.

api/v1alpha1/k8ssandracluster_types.go
jsanda: Good question. When I started on this PR I picked some arbitrary properties to expose in order to help me get started with the code changes. I definitely think we can avoid exposing the system logger resources and probably for server-config-init as well. 

I do think we will need to expose configs for other containers like for backup/restore.

api/v1alpha1/k8ssandracluster_types.go
burmanm: Usage point of view. Now whoever uses the types has both value and references in their code instead of consistently getting pointers for example ([]*cassdcapi.Rack).

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -138,6 +189,7 @@ type CassandraDatacenterTemplateSpec struct {
 }
 
 type EmbeddedObjectMeta struct {
+	// +optional
```
burmanm, line 192: in which case would this be optional?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -147,6 +199,81 @@ type EmbeddedObjectMeta struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
```
burmanm, line 199: Annotations, but no labels?

api/v1alpha1/k8ssandracluster_types.go
```diff
@@ -147,6 +199,81 @@ type EmbeddedObjectMeta struct {
...
+}
+
+type Auth struct {
+	Enabled bool `json:"enabled,omitempty"`
+
+	CacheValidityPeriodMillis *int64 `json:"cacheValidityPeriodMillis,omitempty"`
+
+	CacheUpdateIntervalMillis *int64 `json:"cacheUpdateIntervalMillis,omitempty"`
+
+	SuperUserSecretName string `json:"SuperUserSecretName,omitempty"`
```
burmanm, line 218: Capital S in the json field name?

controllers/k8ssandracluster_controller.go
```diff
@@ -284,46 +292,25 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
...
-	namespace := template.Meta.Namespace
-	if len(namespace) == 0 {
-		namespace = k8ssandraKey.Namespace
-	}
+func (r *K8ssandraClusterReconciler) resolveSeedEndpoints(ctx context.Context, dc *cassdcapi.CassandraDatacenter, remoteClient client.Client) ([]string, error) {
+	podList := &corev1.PodList{}
+	labels := client.MatchingLabels{cassdcapi.DatacenterLabel: dc.Name}
 
-	config, err := cassandra.GetMergedConfig(template.Config, dcNames, systemDistributedRF, template.ServerVersion)
+	err := remoteClient.List(ctx, podList, labels)
```
burmanm, line 299: Is my code editor playing funny or do we not have rights to do this cluster-wide (no namespace restriction)? This would require ClusterRole rights to pods.

api/v1alpha1/k8ssandracluster_types.go
jsanda: Oh, are you referring to `Racks` _not_ being a pointer?

jsanda: For slices, strings, and maps, it is easy to determine if the value is _empty_. For structs and primitive types a pointer makes it possible to distinguish whether the property has been set.

api/v1alpha1/k8ssandracluster_types.go
jsanda: I am going to leave this as is for now. I created #120 to come back and address. If there is consensus to remove that's fine with me.

api/v1alpha1/k8ssandracluster_types.go
jsanda: I created #121 to address validation checks.

api/v1alpha1/k8ssandracluster_types.go
jsanda: Labels are there as well.

api/v1alpha1/k8ssandracluster_types.go
jsanda: fixed

controllers/k8ssandracluster_controller.go
jsanda: It's not your editor ;-) Interestingly this code works. Regardless, I will add the namespace filter.

pkg/cassandra/datacenter.go
jsanda: I created #122

api/v1alpha1/k8ssandracluster_types.go
jsanda: The DC's namespace is set to the namespace of the K8ssandraCluster.

burmanm:
I guess something went wrong with the rebase since Stargate is now failing in e2e tests

burmanm (approved):

jsanda:
> I guess something went wrong with the rebase since Stargate is now failing in e2e tests

I just checked the logs on each of the tests and it looks like they are failing in the cleanup with the replicated secret.

PR #100, opened 2021-08-31
Create data_endpoint_auth keyspace before deploying Stargate (fixes #57)

### LINKED ISSUE #57: Multi-cluster deployment of Stargate resources isn't deploying as expected
┆Issue is synchronized with this [Jiraserver Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-777) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-777
┆Priority: Medium

adutra:
**What this PR does**:
This PR fixes Stargate multi-DC deployment issues detected previously by creating the data_endpoint_auth keyspace _before_ Stargate is deployed, with the appropriate replication config and strategy.

Fixes #57

**Checklist**
- [x] Changes manually tested


adutra:
Note to reviewers: there is a Stargate issue that would allow a Stargate node to create the keyspace with appropriate config: https://github.com/stargate/stargate/issues/739.

But for now we need this, and conceptually speaking, I think it's the k8ssandracontroller responsibility to create that keyspace, since it knows how many dcs are going to be created and where.

One consequence of this is that the keyspace is _not_ created when Stargate is created in standalone mode. I think it's fair to assume that if the Stargate resource is created directly by the user, then it's their responsibility to create the keyspace themselves beforehand.

jsanda:
>But for now we need this, and conceptually speaking, I think it's the k8ssandracontroller responsibility to create that keyspace, since it knows how many dcs are going to be created and where.

agreed

>One consequence of this is that the keyspace is not created when Stargate is created in standalone mode. I think it's fair to assume that if the Stargate resource is created directly by the user, then it's their responsibility to create the keyspace themselves beforehand.

agreed

controllers/k8ssandracluster_controller.go
```diff
@@ -169,113 +172,118 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+	stargateKey := types.NamespacedName{
+		Namespace: actualDc.Namespace,
+		Name:      kc.Name + "-" + actualDc.Name + "-stargate",
+	}
+	actualStargate := &api.Stargate{}
+	logger = logger.WithValues("Stargate", stargateKey)
+
+	if stargateTemplate != nil {
+
+		if !kc.Status.IsStargateAuthKeyspaceCreated() {
```
jsanda, line 213: More and more I am uncertain about looking at the status of an object to determine whether a change should be applied. We have seen a similar approach in medusa-operator cause problems. 

Here is my reasoning. The status describes the current status of the object. When a Reconciler is called it fetches the object to be reconciled. The status of the object at that point might be stale. There was some event that triggered the reconciliation and now as part of the reconciliation we may recompute and change the status.

Let's look at the situation in practice. Suppose after the auth keyspace is created I accidentally drop it. Will the controller recreate it? I think it should. If we rely on the status condition check, then then controller won't recreate it.

jsanda (changes requested):

controllers/k8ssandracluster_controller.go
```diff
@@ -169,113 +172,118 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+	dcTemplate api.CassandraDatacenterTemplateSpec,
+	actualDc *cassdcapi.CassandraDatacenter,
+	logger logr.Logger,
+	remoteClient client.Client,
+) (ctrl.Result, error) {
+
+	stargateTemplate := dcTemplate.Stargate.Coalesce(kc.Spec.Stargate)
+	stargateKey := types.NamespacedName{
+		Namespace: actualDc.Namespace,
+		Name:      kc.Name + "-" + actualDc.Name + "-stargate",
```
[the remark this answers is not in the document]
adutra: Agreed. Also later on I'd suggest that we revisit the names of services and deployments created by the Stargate controller, the suffix `-service` or `-deployment` is not really needed.

controllers/k8ssandracluster_controller.go
adutra: I agree with your reasoning. But how can we check the existence of a keyspace inside the reconcile loop? The management API does not offer such an endpoint. I could use the go-cql driver but I would rather avoid it. I could also use go-cassandra-native-protocol [simple client](https://github.com/datastax/go-cassandra-native-protocol/tree/main/client), but it has been designed for testing purposes. Do you have other options in mind?

pkg/cassandra/management.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (r *defaultManagementApiFacade) CreateKeyspace(
+	ctx context.Context,
+	dc *cassdcapi.CassandraDatacenter,
+	remoteClient client.Client,
+	keyspaceName string,
+	replication map[string]int,
+	logger logr.Logger,
+) error {
+	if managementApiClient, err := r.createManagementApiClient(ctx, dc, remoteClient, logger); err != nil {
```
adutra, line 40: @jsanda do you know if it's possible to store this client as a struct field and reuse it? Or is it a disposable client?

controllers/k8ssandracluster_controller.go
jsanda: Unfortunately I do not have any other options in mind. I suppose one option could be to always do a `CREATE IF NOT EXISTS`.

pkg/cassandra/management.go
jsanda: I think it makes sense to store it in some some of request scope variable  (conceptually speaking since there is no such scope in Go). We shouldn't store it across calls to the Reconciler though.

controllers/k8ssandracluster_controller.go
adutra: Unfortunately the management API does not issue the `CREATE KEYSPACE` statement with the `IF NOT EXISTS` option, see [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/0608d0c32e8f7d1bb51e5d9b62e05eb767474a4f/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java#L354-L357).

adutra: Oh sorry my bad: it does use a `IF NOT EXISTS` option, the right code to look at is [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/0608d0c32e8f7d1bb51e5d9b62e05eb767474a4f/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java#L336-L340). 

So that might be a quick win.

adutra: I suggest the following course of action:

1. As a temporary workaround, I will simply remove the condition I created and call the `CreateKeyspace` endpoint for _every_ Stargate resource that the controller creates. 
2. As a long-term solution, we are going to create a new Management API endpoint to list existing keyspaces, then call the `CreateKeyspace` endpoint only if the Stargate auth keyspace is not present yet.

As I'm stepping out for a few days, @adejanovski offered to resume the work where I left. I think though that we could merge this PR with the temporary workaround, then Alex would open a follow-up PR when the endpoint is available. How does that sound @jsanda @adejanovski ?

jsanda: Sounds good to me! Can you create an issue for the follow up work.

adutra: There you go: https://github.com/k8ssandra/k8ssandra-operator/issues/106.

controllers/k8ssandracluster_controller.go
```diff
@@ -462,6 +472,21 @@ func (r *K8ssandraClusterReconciler) setStatusForStargate(kc *api.K8ssandraClust
 	return nil
 }
 
+func (r *K8ssandraClusterReconciler) ensureStargateAuthKeyspaceExists(
```
jsanda, line 475: We should only call `ensureStargateAuthKeyspaceExists` once, not once per Stargate instance. We can address this in #106.

main.go
```diff
@@ -154,6 +154,7 @@ func main() {
 			Scheme:        mgr.GetScheme(),
 			ClientCache:   clientCache,
 			SeedsResolver: cassandra.NewRemoteSeedsResolver(),
+			ManagementApi: cassandra.NewManagementApiFacade(),
```
jsanda, line 157: I am not a fan of having the `ManagementApi` field in the Reconciler. Every method that we add to the `ManagementApiFacade` interface will have to include connection info. More importantly, we have to initialize the connection on each method invocation. That is overhead we could otherwise avoid. I will create a follow up issue for this.

jsanda (approved):

PR #114, opened 2021-09-03
Create a ReplicatedSecret in startup and superuserSecret when installing CassDc [K8SSAND-846]

### LINKED ISSUE #97: Create replicated superuser secret before creating CassandraDatacenters
We need k8ssandra-operator to create and manage the default superuser secret so that each CassandraDatacenter does not wind up with a different set of credentials.

Now that #53 has been merged the K8ssandraCluster controller can do the following:

* Create a secret with default superuser and random password
  * Or allow the user to specify a secret with credentials
* Create a ReplicatedSecret that references the the superuser secret

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-846) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-846
┆Priority: Medium

burmanm:
**What this PR does**:
Creates a ReplicatedSecret that replicates all the k8ssandra-operator managed secrets to all the available clusters. Also, when creating a new Datacenter, create a superuserSecret and set its name to prevent cass-operator from creating another.

Fixes #97 

**Checklist**
- [ ] Changes manually tested


pkg/secret/replicated.go
```diff
@@ -0,0 +1,127 @@
...
+	return secretName
+}
+
+// VerifySuperuserSecret creates the superUserSecret with propert annotations
+func VerifySuperuserSecret(c client.Client, secretName, namespace string) error {
+	if secretName == "" {
+		return fmt.Errorf("secretName is required")
+	}
+	currentSec := &corev1.Secret{}
+	err := c.Get(context.Background(), types.NamespacedName{Name: secretName, Namespace: namespace}, currentSec)
```
jsanda, line 52: We should using a context from the caller, not `context.Background()`.

pkg/secret/replicated.go
```diff
@@ -0,0 +1,127 @@
...
+				Type:       "Opaque",
+				// Immutable feature is only available from 1.21 and up (beta in 1.19 and up)
+				// Immutable:  true,
+				Data: map[string][]byte{
+					"username": []byte(secretName),
+					"password": password,
+				},
+			}
+
+			return c.Create(context.Background(), sec)
```
jsanda, line 71: We should using a context from the caller, not `context.Background()`.

pkg/secret/replicated.go
```diff
@@ -0,0 +1,127 @@
...
+			ReplicationTargets: replicationTargets,
+		},
+	}
+}
+
+func getManagedObjectMeta(name, namespace string) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      name,
+		Namespace: namespace,
+		Labels: map[string]string{
```
jsanda, line 123: I think we should include additional labels, namely `k8ssandra.io/cluster`. I raised the same point for some of the objects created by the stargate controller. Here is the rationale. If every object that the operator creates includes a common label, then a simple label selector query can return an inventory of all objects created on behalf of a K8ssandraCluster. I think this could be useful from a security standpoint.

pkg/secret/replicated.go
```diff
@@ -0,0 +1,127 @@
...
+			return c.Create(context.Background(), sec)
+		}
+	}
+
+	// It exists or was created, we won't modify it anymore
+	return nil
+}
+
+// VerifyReplicatedSecret ensures that the correct replicatedSecret for all managed secrets is created
+func VerifyReplicatedSecret(c client.Client, leaderID, namespace string, targetContexts []string) error {
```
jsanda, line 80: Not sure I understand the purpose of this.

controllers/secret_controller.go
```diff
@@ -113,7 +113,7 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request)
 					}
 					for _, deleteKey := range secretsToDelete {
 						err = remoteClient.Delete(ctx, deleteKey)
-						if err != nil {
+						if err != nil && !errors.IsNotFound(err) {
```
jsanda, line 116: How come you removed this `IsNotFound` check?

jsanda (changes requested):

pkg/secret/replicated.go
burmanm: The function? Without it, there's no replication of that superUserSecret - it would only remain in the local cluster.

controllers/secret_controller.go
burmanm: I added it? ;)

controllers/k8ssandracluster_controller.go
```diff
@@ -133,6 +140,13 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
 
 			if actualHash, found := actualDc.Annotations[api.ResourceHashAnnotation]; !(found && actualHash == desiredDcHash) {
 				logger.Info("Updating datacenter")
+
+				if actualDc.Spec.SuperuserSecretName != desiredDc.Spec.SuperuserSecretName {
```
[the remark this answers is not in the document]
burmanm: It's trickier than that. cass-operator's validation will reject any changes to the superUserSecretName. Thus, if we've diverged - the actualDc's value is the one that should be used instead of desired. But if that's different per cluster, then essentially we're pretty screwed.

Thus, whatever you set to K8ssandraCluster object could be something we will never use - and the reconcile would have to use the actualDc value in this case. That's not a pretty sight (we probably need a validation to prevent this in K8ssandraCluster webhook).

pkg/secret/replicated.go
burmanm: ReplicatedSecret is per k8ssandra-operator deployment, but I'll add that to the superUserSecret.

controllers/k8ssandracluster_controller.go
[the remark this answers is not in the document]
jsanda: Thanks for pointing out the cass-operator behavior. It looks like cass-operator prevents you from pointing to a different secret but you could still potentially modify the existing secret.

Two questions,

* Can we wind up in state from which we cannot recover?
* Will existing behavior cause any problems for k8ssandra 1.x migration scenarios?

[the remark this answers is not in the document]
burmanm: We could, if somehow different CassandraDatacenters use different superuserSecretName. That could happen if user modifies the K8ssandraCluster's SuperuserSecretName and then adds a new Datacenter.

It shouldn't hurt migrations, we already have a superuser there. We need to just copy that setting to our K8ssandraCluster.

main.go
```diff
@@ -167,6 +172,8 @@ func main() {
 			setupLog.Error(err, "unable to create controller", "controller", "SecretSync")
 			os.Exit(1)
 		}
+		// Part of initial configuration check, but only done on the control-plane
+		secret.VerifyReplicatedSecret(ctx, uncachedClient, options.LeaderElectionID, watchNamespace, contextNames)
```
jsanda, line 176: Shouldn't we handle the returned error here?

burmanm: I'm moving this to per-cluster-replicatedSecret.

controllers/controllers_test.go
```diff
@@ -62,7 +62,6 @@ func TestControllers(t *testing.T) {
 	t.Run("Stargate", func(t *testing.T) {
 		testStargate(ctx, t)
 	})
-
 	t.Run("SecretController", func(t *testing.T) {
```
jsanda, line 65: The K8ssandraCluster controller integration tests need to be updated. We need updates in the e2e tests as well. Maybe it is sufficient to check that the superuser secret in each cluster matches.

controllers/k8ssandracluster_controller_test.go
```diff
@@ -17,7 +20,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"testing"
 )
 
 func testK8ssandraCluster(ctx context.Context, t *testing.T) {
```
jsanda, line 25: We need to verify that the ReplicatedSecret is created.

We also need coverage in the e2e tests. It might be sufficient to check that the secrets are the same across clusters.

jsanda (changes requested):

pkg/secret/replicated.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		result[i] = charset[n.Int64()]
+	}
+
+	return result, nil
+}
+
+// DefaultSuperuserSecretName follows the convention from k8ssandra Helm charts
+func DefaultSuperuserSecretName(clusterName string) string {
+	cleanedClusterName := strings.ReplaceAll(strings.ReplaceAll(clusterName, "_", ""), "-", "")
```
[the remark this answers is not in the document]
burmanm: This is what our helm charts did.

controllers/k8ssandracluster_controller.go
```diff
@@ -190,7 +193,7 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
 			if errors.IsNotFound(err) {
 				logger.Info("Creating superuserSecret", "SecretName", desiredDc.Spec.SuperuserSecretName)
 				// This could be the first Datacenter to be created. Verify superUserSecret is created also
-				if err = secret.VerifySuperuserSecret(ctx, r.Client, desiredDc.Spec.SuperuserSecretName, kc.Spec.Cassandra.Cluster, kc.GetNamespace()); err != nil {
+				if err = secret.ReconcileSuperuserSecret(ctx, r.Client, desiredDc.Spec.SuperuserSecretName, kc.Spec.Cassandra.Cluster, kc.GetNamespace()); err != nil {
```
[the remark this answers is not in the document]
burmanm: Yup, it's in the wrong place.

pkg/secret/replicated.go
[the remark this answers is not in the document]
jsanda: I think it is replacing whitespace and underscores with dashes. Here is the helper template:

```
{{- include "k8ssandra.clusterName" . | replace " " "-" | replace "_" "-" }}-superuser
```

jsanda (approved):
Awesome work!

PR #116, opened 2021-09-07
Add ability to watch multiple namespaces [K8SSAND-843]

### LINKED ISSUE #95: K8SSAND-843 ⁃ #95 - k8ssandra-operator should support cluster-scoped deployments
Currently k8ssandra-operator is namespace-scoped. It only watches objects in specified namespaces which defaults to the namespace in which the operator is deployed. Interestingly the operator can be configured to watch different namespaces in remote clusters as compared to the ones it watches in the control plane cluster.

We should be able to configure the operator to be cluster-scoped such that it will watch object across all namespaces. Furthermore we should be able to create a multi-DC C* cluster (within a single k8s cluster) where each DC is in a different namespace.

There are a number of things that need (or may need) to be done for this including:

* Implement changes in operator code
  *  Some initial work has been started in #116. I did an initial review and pointed out some other changes that are needed.
  * K8ssandraClusterReconciler configures a watch on Stargate and CassandraDatacenter objects with a mapping function that checks for the cluster.  We may need to update the value of the label to also include the namespace of the K8ssandra cluster (or alternatively add a label for the namespace of the K8ssandraCluster).
*   Update kustomize manifests 
*  Update Helm chart
* Add `envtest` integration tests for
  * Multi-DC cluster in a single cluster across multiple namespaces
  * Multi-DC cluster in multiple cluster across multiple namespaces
* Add e2e tests for
  * Multi-DC cluster in a single cluster across multiple namespaces
  * Multi-DC cluster in multiple cluster across multiple namespaces 
* Update install guide 

It might make sense to break some of this up into multiple tickets.

Lastly note that the work describe here should also take care of #19.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-843) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-843
┆Priority: Medium

burmanm:
**What this PR does**:
Adds the ability to watch multiple namespaces when k8ssandra-operator is deployed in cluster scoped mode.

Fixes #95 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
@@ -33,6 +34,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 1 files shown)


jsanda (changes requested):
Thanks for this. We need some tests with a cluster-scoped deployment, definitely an e2e test.

config/cluster/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
...
+    name: k8ssandra-operator
+  patch: |-
+    - op: replace
+      path: /roleRef/kind
+      value: ClusterRole
+    - op: replace
+      path: /kind
+      value: ClusterRoleBinding
+resources:
+- ../../config/default
```
jsanda, line 38: You cannot use `../../config/default` because it will deploy the namespace-scoped cass-operator. I pushed changes to my branch. See https://github.com/jsanda/k8ssandra-operator/commit/bf6d4fd338313cf8cfa3c8653fdf8042ad5a1237 to address this. I introduced another kustomization directory because I assume we want k8ssandra-operator deployed in a namespace of the same name and the same for cass-operator.

jsanda (dismissed):

main.go
```diff
@@ -91,6 +93,17 @@ func main() {
 		Namespace:              watchNamespace,
 	}
 
+	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
+	if strings.Contains(watchNamespace, ",") {
```
jsanda, line 97: If `watchNamespace` specifies multiple namespaces then we need to change the `cluster.New` calls to look something like:

```go
c, err := cluster.New(cfg, func(o *cluster.Options) {
	o.Scheme = scheme
	o.Namespace = ""
	o.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(watchNamespace, ","))
})
```
We also need to change the query that fetches ClientConfigs. It needs to specify multiple/separate namespace options.

After those changes, we need to deploy the operator with a ClusterRole and ClusterRoleBinding instead of the Role and RoleBinding.

Lastly, we need updated test coverage.

jsanda (changes requested):

PR #150, opened 2021-09-23
fix errors in create-clientconfig.sh for getting current contexts

jsanda:
While doing some testing today I came across a couple errors in `create-clientconfig.sh` for getting the current context for the src and dest clusters. This PR fixes those errors.

### DIFF
--- a/scripts/create-clientconfig.sh
+++ b/scripts/create-clientconfig.sh
@@ -14,7 +14,7 @@
 # TODO Accept multiple values for the src-context option and generate a kubeconfig with
 #      entries for each
 
-set -e
+set -ex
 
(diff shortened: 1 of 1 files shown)


scripts/create-clientconfig.sh
```diff
@@ -63,7 +63,6 @@ while true; do
     --serviceaccount ) service_account="$2"; shift 2 ;;
     --output-dir ) output_dir="$2"; shift 2 ;;
     --in-cluster-kubeconfig ) in_cluster_kubeconfig="$2"; shift 2 ;;
-    --dev ) dev_mode=true; shift 1 ;;
```
adutra, line 66: So dev mode never really existed?

adutra (approved):

scripts/create-clientconfig.sh
jsanda: When I was working on the install steps for the docs, I realized that there were some problems with the script if you try to use it with our kind cluster setup. When I wrote it I was really only thinking about real clusters. I wanted to come up with a quick, minimally invasive fix. Initially I thought about a dev mode but abandoned the idea. I missed removing that part previously.

PR #152, opened 2021-09-24
configure num_tokens [K8SSAND-923]

### LINKED ISSUE #151: Set defaults for num_tokens and make it configurable
Even though we use the stock defaults for `num_tokens` we need to explicitly set it because the config builder uses a default of `num_tokens: 1`. It should be configurable as well in case users want to use a non-default value.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-923) by [Unito](https://www.unito.io)
┆Fix Versions: k8ssandra-operator-1.0.0-alpha.1
┆Issue Number: K8SSAND-923
┆Priority: Medium

jsanda:
**What this PR does**:
Configures `num_tokens` and `allocate_tokens_for_local_replication_factor`

Fixes #151 

**Checklist**
- [ ] Changes manually tested


pkg/cassandra/config_test.go
```diff
@@ -269,22 +296,67 @@ func TestCreateJsonConfig(t *testing.T) {
...
+					AllocateTokensForLocalReplicationFactor: intPtr(5),
+				},
+			},
+			want: `{
+              "cassandra-yaml": {
+                "allocate_tokens_for_local_replication_factor": 5,
+				"num_tokens": 16
               }
             }`,
 		},
```
adutra, line 362: Not entirely related to the changes introduced here, but I noticed that the test is not creating subtests for each test case, and it's not using testify either. I think you could update that part of the code to:

```go
for _, tc := range tests {
	t.Run(tc.name, func(t *testing.T) {
		var err error
		tc.got, err = CreateJsonConfig(tc.config, tc.cassandraVersion)
		require.NoError(t, err, "failed to create json dcConfig")
		expected, err := gabs.ParseJSON([]byte(tc.want))
		require.NoError(t, err, "failed to parse expected value")
		actual, err := gabs.ParseJSON(tc.got)
		require.NoError(t, err, "failed to parse actual value")
		assert.Equal(t, expected, actual)
	})
}
```

adutra (approved):

pkg/cassandra/config_test.go
jsanda: Nice :)

PR #154, opened 2021-09-25
Fix Stargate BadRequest auth response due to invalid credentials [K8SSAND-926]

adutra:
**What this PR does**:

WIP attempt to fix Stargate auth failures on GHA

None

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/test/e2e/stargate_apis_test.go
+++ b/test/e2e/stargate_apis_test.go
@@ -55,7 +55,6 @@ func testStargateNativeApi(t *testing.T, ctx context.Context, k8sContextIdx int,
 	createKeyspaceAndTableNative(t, connection, tableName, keyspaceName, replication)
 	insertRowsNative(t, connection, 10, tableName, keyspaceName)
 	checkRowCountNative(t, connection, 10, tableName, keyspaceName)
-	dropKeyspaceNative(t, connection, keyspaceName)
 }
 
 func testSchemaApi(t *testing.T, restClient *resty.Client, k8sContextIdx int, token string, replication map[string]int) {
@@ -64,7 +63,6 @@ func testSchemaApi(t *testing.T, restClient *resty.Client, k8sContextIdx int, to
 	createKeyspaceAndTableRest(t, restClient, k8sContextIdx, token, tableName, keyspaceName, replication)
 	insertRowsRest(t, restClient, k8sContextIdx, token, 10, tableName, keyspaceName)
 	checkRowCountRest(t, restClient, k8sContextIdx, token, 10, tableName, keyspaceName)
-	dropKeyspaceRest(t, restClient, k8sContextIdx, token, keyspaceName)
 }
 
 func testDocumentApi(t *testing.T, restClient *resty.Client, k8sContextIdx int, token string, replication map[string]int) {
@@ -73,7 +71,6 @@ func testDocumentApi(t *testing.T, restClient *resty.Client, k8sContextIdx int, 
 	createDocumentNamespace(t, restClient, k8sContextIdx, token, documentNamespace, replication)
 	writeDocument(t, restClient, k8sContextIdx, token, documentNamespace, documentId)
 	readDocument(t, restClient, k8sContextIdx, token, documentNamespace, documentId)
-	dropDocumentNamespace(t, restClient, k8sContextIdx, token, documentNamespace)
 }
 
 func createKeyspaceAndTableRest(t *testing.T, restClient *resty.Client, k8sContextIdx int, token, tableName, keyspaceName string, replication map[string]int) {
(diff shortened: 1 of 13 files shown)


test/e2e/stargate_apis_test.go
```diff
@@ -209,18 +225,20 @@ func readDocument(t *testing.T, restClient *resty.Client, k8sContextIdx int, tok
...
+	require.IsType(t, map[string]interface{}{}, genericJson["data"], "Expected field data to be map[string]interface{}")
 	assert.Equal(t, awesomeMovieDirector, genericJson["data"].(map[string]interface{})["Director"], "Expected JSON payload to contain a field data.Director")
 	assert.Equal(t, awesomeMovieName, genericJson["data"].(map[string]interface{})["Name"], "Expected JSON payload to contain a field data.Name")
 }
 
 func authenticate(t *testing.T, restClient *resty.Client, k8sContextIdx int, username, password string) string {
 	url := fmt.Sprintf("http://stargate.127.0.0.1.nip.io:3%v080/v1/auth", k8sContextIdx)
-	body := fmt.Sprintf("{\"username\": \"%s\", \"password\": \"%s\"}", username, password)
+	body := map[string]string{"username": username, "password": password}
 	request := restClient.NewRequest().
```
jsanda, line 242: Would it make sense and help to wrap the request in an `Eventually` call?

test/e2e/suite_test.go
```diff
@@ -237,6 +237,13 @@ func afterTest(t *testing.T, namespace string, f *framework.E2eFramework, deploy
 }
 
 func cleanUp(t *testing.T, namespace string, f *framework.E2eFramework, deployTraefik bool) error {
+
+	if t.Failed() {
```
jsanda, line 241: We don't need to dump the logs here since it is done in f.DumpClusterInfo()`. The functions are nice have though in case we want to dump logs at point before the test finishes. It would be nice to have them written to a directory like `DumpClusterInfo` but this is a secondary concern right now.

jsanda:
@adutra the latest failure in `CreateStargateAndDatacenter` appears to be caused by this:

```
INFO  [MigrationStage:1] 2021-09-29 20:25:39,811 ColumnFamilyStore.java:427 - Initializing data_endpoint_auth.token
ERROR [MigrationStage:1] 2021-09-29 20:25:40,838 CassandraDaemon.java:235 - Exception in thread Thread[MigrationStage:1,5,main]
org.apache.cassandra.exceptions.ConfigurationException: Column family ID mismatch (found 69ef6ba0-2163-11ec-9958-d18d6ebc904d; expected 68f8d6a0-2163-11ec-962f-b309ed200eca)
        at org.apache.cassandra.config.CFMetaData.validateCompatibility(CFMetaData.java:984)
        at org.apache.cassandra.config.CFMetaData.apply(CFMetaData.java:938)
        at org.apache.cassandra.config.Schema.updateTable(Schema.java:687)
        at org.apache.cassandra.schema.SchemaKeyspace.updateKeyspace(SchemaKeyspace.java:1478)
        at org.apache.cassandra.schema.SchemaKeyspace.mergeSchema(SchemaKeyspace.java:1434)
        at org.apache.cassandra.schema.SchemaKeyspace.mergeSchema(SchemaKeyspace.java:1403)
        at org.apache.cassandra.schema.SchemaKeyspace.mergeSchemaAndAnnounceVersion(SchemaKeyspace.java:1380)
        at org.apache.cassandra.db.DefinitionsUpdateVerbHandler$1.runMayThrow(DefinitionsUpdateVerbHandler.java:51)
        at org.apache.cassandra.utils.WrappedRunnable.run(WrappedRunnable.java:28)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at org.apache.cassandra.concurrent.NamedThreadFactory.lambda$threadLocalDeallocator$0(NamedThreadFactory.java:84)
        at java.lang.Thread.run(Thread.java:748)
```
This looks like a legitimate failure for a change 😅  It is due to creating the Deployments at the same time. I know we were discussing how to handle this the other day. 

Here is my suggestion for now. It would be nice to have all tests passing for the alpha release. How about we 1) document this behavior in a ticket  and 2) change the test to deploy only a single Stargate node provided that isn't too much effort. wdyt?

jsanda:
@adutra I will go ahead and create the ticket and update the test since you are probably/hopefully out for the day.

jsanda:
How about we rename `CreateStargateAndDatacenter` to something like `CreateMultiStargateAndDatacenter` and then add a new test that deploys a single Stargate along with the CassandraDatacenter? Then we disable the multi-Stargate test for now.

Ultimately I think it would be better to update the test to create the schema before deploying Stargate. Currently this isn't possible as both the Stargate and CassandraDatacenter are created at the same time, and this would lead to a race condition. We would to make some changes to the test framework to better support this. I'd prefer to do that after the alpha release.

test/e2e/stargate_apis_test.go
adutra: The very first time we authenticate is actually done inside `DeployStargateIngresses` and that call is wrapped in an `Eventually` block. If that `Eventually` succeeds there should be no reason for this call here to fail.

adutra: Also, this call here was failing previously because in `DeployStargateIngresses` we weren't doing a full authentication, we were just pinging the auth endpoint.

test/e2e/stargate_apis_test.go
```diff
@@ -33,30 +33,18 @@ func testStargateApis(t *testing.T, ctx context.Context, k8sContextName string,
...
-	println()
-	fmt.Println("=============", "BEGIN STARGATE LOGS", dc, rack, "context", k8sContextName, "namespace", namespace, "=============")
-	fmt.Println(string(output))
-	fmt.Println("=============", "END STARGATE LOGS", dc, rack, "context", k8sContextName, "namespace", namespace, "=============")
-	println()
-}
-
 func testStargateRestApis(t *testing.T, k8sContextIdx int, username string, password string, replication map[string]int) {
 	restClient := resty.New()
 	token := authenticate(t, restClient, k8sContextIdx, username, password)
```
jsanda, line 41: Do we need to authenticate again here? Could we reuse the token acquired in `DeployStargateIngresses`? I am mostly asking out of curiosity and don't want to change anything at the moment unless absolutely necessary.

jsanda (approved):
@adutra thanks for all the time and effort you have put into this!

PR #180, opened 2021-10-01
fix tag name

jsanda:
fix image tag name

### DIFF
--- a/docs/install/README.md
+++ b/docs/install/README.md
@@ -899,7 +899,7 @@ Stargate node.
 
 ```sh
 cat <<EOF | kubectl -n k8ssandra-operator apply -f -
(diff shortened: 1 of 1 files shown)


docs/install/README.md
```diff
@@ -899,7 +899,7 @@ Stargate node.
 
 ```sh
 cat <<EOF | kubectl -n k8ssandra-operator apply -f -
-apiVersion: k8ssandra.io/v1alpha1
+apiVersion: k8ssandra.io/v1alphav1.0.0-alpha.11
```
jdonenine, line 902: I think you've got an extra 1 in there... Unless we really did a lot of work today to get to 11 😉

jdonenine (changes requested):

jdonenine (approved):

PR #191, opened 2021-10-08
Stargate auth table creation may trigger a table ID mismatch (fixes #156)

### LINKED ISSUE #156: K8SSAND-926 ⁃ Stargate auth table creation may trigger a table ID mismatch
When there are more than one Stargate pod to create in a Stargate deployment, sometimes the initialization of the auth keyspace fails: even if the keyspace is created beforehand, two or more Stargate pods may attempt to create the `token` table at the same time, in which case the first one will succeed, and all others with fail with a table ID mismatch error:

Example from recent runs:

Stargate 1 creates the keyspace and the table:

```java
INFO  [main] 2021-09-25 15:04:08,327 AuthnTableBasedService.java:64 - Initializing keyspace data_endpoint_auth and table token for table based auth
INFO  [Native-Transport-Requests-1] 2021-09-25 15:04:08,411 MigrationManager.java:338 - Create new Keyspace: KeyspaceMetadata{name=data_endpoint_auth [...]
INFO  [Native-Transport-Requests-1] 2021-09-25 15:04:10,092 MigrationManager.java:368 - Create new table: org.apache.cassandra.config.CFMetaData@139d53c[cfId=d6569ac0-1e11-11ec-80a1-b7f7776197d6,ksName=data_endpoint_auth,cfName=token [...]
INFO  [MigrationStage:1] 2021-09-25 15:04:14,781 ColumnFamilyStore.java:427 - Initializing data_endpoint_auth.token
```

Stargate 2 notices that the keyspace exists but fails to notice the table exists too, and attempts to create it with a different ID:

```java
INFO  [main] 2021-09-25 15:04:09,611 AuthnTableBasedService.java:64 - Initializing keyspace data_endpoint_auth and table token for table based auth
INFO  [Native-Transport-Requests-1] 2021-09-25 15:04:09,880 MigrationManager.java:368 - Create new table: org.apache.cassandra.config.CFMetaData@27d1256e[cfId=d6364180-1e11-11ec-b65e-adadd905ac52,ksName=data_endpoint_auth,cfName=token [...]
INFO  [MigrationStage:1] 2021-09-25 15:04:12,632 ColumnFamilyStore.java:427 - Initializing data_endpoint_auth.token
ERROR [MigrationStage:1] 2021-09-25 15:04:12,942 CassandraDaemon.java:235 - Exception in thread Thread[MigrationStage:1,5,main]
org.apache.cassandra.exceptions.ConfigurationException: Column family ID mismatch (found d6569ac0-1e11-11ec-80a1-b7f7776197d6; expected d6364180-1e11-11ec-b65e-adadd905ac52)
```

Note that this may happen regardless of whether the Stargate resource has been created by the K8ssandra controller or not. Indeed, the K8ssandra controller handles the keyspace creation, but not the table creation currently.

I think we don't need to care about the case where this happens on a standalone Stargate resource, since the rationale until now is that if the user creates the Stargate resource themselves, it's their responsibility to initialize that keyspace with appropriate replication.

I think the best fix is to have the k8ssandra controller also handle the creation of the `token` table. Unfortunately, this means a new Management API endpoint to create.

Alternatively we could turn Stargate deployments into statefulsets but that's a bit too heavy-handed imo.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-926) by [Unito](https://www.unito.io)
┆Fix Versions: k8ssandra-operator-1.0.0-beta.1
┆Issue Number: K8SSAND-926
┆Priority: Medium

adutra:
**What this PR does**:

This PR:
* adds new methods to `ManagementApiFacade` to handle table creation and retrieval of keyspace replication settings;
* changes the Stargate auth schema reconciliation logic to:
  * only alter the keyspace when the desired replication doesn't match the actual one;
  * also create the `token` table if it doesn't exist.

It depends on:

* https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/143
* https://github.com/k8ssandra/cass-operator/pull/193

Fixes #156

**Checklist**
- [x] Changes manually tested


adutra:
@jsanda @burmanm can we merge this or should we wait for a cass-operator release?

jsanda:
@adutra I am fine with merging before cass-operator is released.

jsanda (changes requested):
@adutra sorry for the delayed review. There is an e2e test failure. I wanted to see if it was a flake but logs aren't available anymore. You also need to update the cass-operator version in the kustomize manifests [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/cluster/kustomization.yaml), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/control-plane/kustomization.yaml), and [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/data-plane/kustomization.yaml).

Then we will need a similar update to the Helm chart. That'll be a separate, follow-up PR though since it is in the k8ssandra repo.

adutra:
> You also need to update the cass-operator version in the kustomize manifests here, here, and here.

Done, thanks! I had to use `ref=master` in the URLs as commit SHAs do not seem to be supported.

> There is an e2e test failure. I wanted to see if it was a flake but logs aren't available anymore. 

I got a clean build 10 mins ago, I will run the e2e tests again to check.

config/deployments/cluster/kustomization.yaml
```diff
@@ -4,7 +4,7 @@ kind: Kustomization
 # Do not set namespace here!
 
 resources:
-  - github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=v1.8.0
+  - github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=master
```
jsanda, line 7: Commit SHAs do work. It would look like this:

```
- github.com/k8ssandra/cass-operator/config/deployments/default?ref=8cb917829b7250278c5202ffcd16378d386599a9
```
I believe you need to specify the full SHA.

adutra: Ah right, full SHAs do work indeed.

adutra:
> @adutra sorry for the delayed review. There is an e2e test failure. I wanted to see if it was a flake but logs aren't available anymore. You also need to update the cass-operator version in the kustomize manifests [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/cluster/kustomization.yaml), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/control-plane/kustomization.yaml), and [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/config/deployments/data-plane/kustomization.yaml).
> 

In fact it seems that we also reference the cass-operator version [here](https://github.com/k8ssandra/k8ssandra-operator/blob/0416308bfb778cfdc3cf34c1bfeace91cf487bea/pkg/test/testenv.go#L36). But it seems that this is restricted to the CRD definition itself, which didn't change lately. I think it's fine to leave it as is. Can you confirm?

jsanda:
> In fact it seems that we also reference the cass-operator version [here](https://github.com/k8ssandra/k8ssandra-operator/blob/0416308bfb778cfdc3cf34c1bfeace91cf487bea/pkg/test/testenv.go#L36). But it seems that this is restricted to the CRD definition itself, which didn't change lately. I think it's fine to leave it as is. Can you confirm?

Yeah, I think we are ok to leave this one as is.

jsanda (approved):
Nice work!

PR #197, opened 2021-10-11
Use cass-config-builder for Stargate deployments

### LINKED ISSUE #189: K8SSAND-1428 ⁃ Stargate config generation does not work correctly for 4.0
**What happened?**
When implementing #130 it became clear that the tests themselves are not working. So implementing running tests against multiple versions will not pass since the tests are broken.

This is because while we do remove 4.0 properties from Cassandra deployment, it's not done correctly for Stargate which has all those 3.11 properties for Stargate 4.0 deployment and that fails.

```
 Caused by: org.apache.cassandra.exceptions.ConfigurationException: Invalid yaml. Please remove properties [start_rpc, thrift_prepared_statements_cache_size_mb] from your cassandra.yaml
	at org.apache.cassandra.config.YamlConfigurationLoader$PropertiesChecker.check(YamlConfigurationLoader.java:275)
	at org.apache.cassandra.config.YamlConfigurationLoader.loadConfig(YamlConfigurationLoader.java:130)
	at io.stargate.db.cassandra.Cassandra40PersistenceActivator.makeConfig(Cassandra40PersistenceActivator.java:66)
```

I expected them to work since this type of removal of unused parameters is what cass-config-builder should do.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1428) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1428
┆priority: Medium

burmanm:
**What this PR does**:
When Stargate deployment sets the CassandraConfigRefMap, the configuration should be ran through cass-config-builder to ensure it is correctly parsed and invalid fields are removed.

Fixes #189 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.0.md
+++ b/CHANGELOG/CHANGELOG-1.0.md
@@ -14,6 +14,8 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## Unreleased
 
+[CHANGE] [#189](https://github.com/k8ssandra/k8ssandra-operator/issues/189) Use cass-config-builder with Stargate deployment
+
 ## v1.0.0-alpha.1 - 2021-09-30
 
 [FEATURE] [#98](https://github.com/k8ssandra/k8ssandra-operator/issues/98) Create Helm chart for the operator
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -44,11 +44,6 @@ func (c config) MarshalJSON() ([]byte, error) {
 	config := make(map[string]interface{}, 0)
 
 	if c.CassandraYaml != nil {
-		if isCassandra4(c.cassandraVersion) {
-			c.StartRpc = nil
-			c.ThriftPreparedStatementCacheSizeMb = nil
-		}
-
 		// Even though we default to Cassandra's stock defaults for num_tokens, we need to
 		// explicitly set it because the config builder defaults to num_tokens: 1
 		if c.NumTokens == nil {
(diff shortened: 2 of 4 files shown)


jsanda:
I haven't looked at the PR yet but I am not sure that I agree that config-builder should be run when CassandraConfigRefMap is set. I tend to consider CassandraConfigRefMap an advanced user feature, kinda like image overrides in cass-operator. I would consider CassandraConfigRefMap and config-builder to be mutually exclusive.

jsanda:
>Well, cass-operator does put it through cass-config-builder if you set config. I don't see how this would be that much different?

It is different. The `config` property does not point to a cassandra.yaml file. It specifies properties that get merged with the config files generated by config-builder. If we are going to use config-builder with Stargate, I would expect a similar `config` property. This brings up a question though.

If we want to replace config-builder with something else, do want to invest the time and effort into Stargate using it?

>We already have tests that do not start because Stargate can deploy broken configs.

Those configs are not broken. They are specific to C* 3.11. That issue can easily be resolved by creating separate 4.0 configs for the tests.

burmanm:
I don't think we should have 4.0 tests separate from 3.11 tests since we're not testing anything specific to their behaviour. We want the K8ssandraClusters to behave the exact same way.

jdonenine:
Going to close this PR, the basic functionality here will get consumed by a later PR expected to come from similar work happening in cass-operator do this without cass-config-builder.

PR #216, opened 2021-11-05
Migrate reaper_controller to k8ssandra-operator (fixes #4)

### LINKED ISSUE #4: K8SSAND-580 ⁃ Migrate reaper_controller to k8ssandra-operator
[reaper-operator](https://github.com/k8ssandra/reaper-operator) consists of the Reaper CRD, the reaper_controller, and the cassandradatacenter_controller. reaper_controller configures and deploys Reaper. We want migrate that code and the Reaper CRD into this repo. It will probably be easiest to recreate the CRD. We just need to make sure it has the same fields. 

cassandradatacenter_controller's main function is registering CassandraDatacenters with Reaper. We can move this functionality into k8ssandracluster_controller.

For this ticket assume that a Reaper instance per CassandraDatacenter is the default.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-580) by [Unito](https://www.unito.io)
┆Epic: Reaper Controller Integration
┆Issue Number: K8SSAND-580
┆Priority: Medium

adutra:
**What this PR does**:

Migrates Reaper controller from reaper-operator project.

Fixes #4

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -250,7 +250,11 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
 		}
 
 		if err = remoteClient.Get(ctx, dcKey, actualDc); err == nil {
-			result, err := r.reconcileStargate(ctx, kc, dcTemplate, actualDc, logger, remoteClient)
+			result, err := r.reconcileStargate(ctx, kc, dcTemplate, actualDc, logger, remoteClient, &stargateAuthSchemaReconciled)
+			if !result.IsZero() || err != nil {
+				return result, err
+			}
+			result, err = r.reconcileReaper(ctx, kc, dcTemplate, actualDc, logger, remoteClient, &reaperBackendSchemaReconciled)
```
jsanda, line 257: Provided that the schema initialization is done up front, there is no reason that Stargate and Reaper couldn't be deployed at the same time. Just an observation for now 😄

pkg/reaper/resource.go
```diff
@@ -0,0 +1,147 @@
...
+		coalesced.ServiceAccountName = clusterTemplate.ServiceAccountName
+	}
+
+	if dcTemplate != nil && len(dcTemplate.Keyspace) != 0 {
+		coalesced.Keyspace = dcTemplate.Keyspace
+	} else if clusterTemplate != nil && len(clusterTemplate.Keyspace) != 0 {
+		coalesced.Keyspace = clusterTemplate.Keyspace
+	}
+
+	if dcTemplate != nil && len(dcTemplate.CassandraUserSecretRef) != 0 {
```
jsanda, line 98: If `CassandraUserSecretRef` is `nil`, then K8ssandraClusterReconciler should create the user similar to how/where it creates the default superuser. It should get copied to remote clusters via the ReplicatedSecret.

If `CassandraUserSecretRef` is not `nil` then the necessary labels (see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/secret/replicated.go#L133)) should be added to it so that it gets replicated.

pkg/reaper/resource.go
```diff
@@ -0,0 +1,147 @@
...
+		coalesced.CassandraUserSecretRef = dcTemplate.CassandraUserSecretRef
+	} else if clusterTemplate != nil && len(clusterTemplate.CassandraUserSecretRef) != 0 {
+		coalesced.CassandraUserSecretRef = clusterTemplate.CassandraUserSecretRef
+	}
+
+	if dcTemplate != nil && len(dcTemplate.JmxUserSecretRef) != 0 {
```
jsanda, line 104: If `JmxUserSecretRef` is `nil`, then K8ssandraClusterReconciler should create the user similar to how/where it creates the default superuser. It should get copied to remote clusters via the ReplicatedSecret.

If `JmxSecretRef` is not `nil` then the necessary labels (see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/secret/replicated.go#L133)) should be added to it so that it gets replicated.

Also note that remote JMX needs to be configured. See [here](https://github.com/k8ssandra/k8ssandra/blob/main/charts/k8ssandra/templates/cassandra/cassdc.yaml#L226) for how it is done in the k8ssandra Helm chart.

pkg/reaper/schema.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"github.com/go-logr/logr"
+	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
+	k8ssandraapi "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	api "github.com/k8ssandra/k8ssandra-operator/apis/reaper/v1alpha1"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+)
+
+func ReconcileBackendSchema(
```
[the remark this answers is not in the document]
adutra: So I thought about that, but I could see a potential problem:

The extracted function is likely to have this signature:

```
func EnsureReplication(keyspaceName, clusterName string, replication map[string]int , managementApi cassandra.ManagementApiFacade, logger logr.Logger) error
```

As you can see it would reference the `cassandra` package. If this function is placed in the `utils` package, then there could be no reference to `utils` from `cassandra` otherwise we would face a circular reference.

Note that currently no such reference exists; but I believe this could change in the future.

If we want to go forward with the common function approach, I would recommend moving the `management.go` file to its own package, then placing the `EnsureReplication` function in that package.

Wdyt?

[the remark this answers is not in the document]
adutra: In fact, I realize that we could turn `EnsureReplication` into a method of the `ManagementApiFacade` interface. I think this is the most elegant solution. I will go with that for now but let me know if you agree.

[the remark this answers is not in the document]
adutra: I also moved the existing, replication-related stuff from `utils` to `cassandra`. It looks better this way imo.

pkg/reaper/resource.go
adutra: OK I took a stab at that, see af30ccb.

> If CassandraUserSecretRef is not nil then the necessary labels (see here) should be added to it so that it gets replicated.

This is not being done currently for `CassandraDatacenter` secrets. E.g. if  `kc.Spec.Cassandra.SuperuserSecretName` is empty, we are going to generate a default secret name. But if a secret with that name already exists, we are not updating it with the "managed-by" labels.

I therefore made the necessary changes to ensure _every_ secret has the labels inside the `ReconcileSecret` function. Hope that's what you had in mind.

adutra:
Sorry folks putting this back to draft for today as I'm working on a small refactoring of reaper_controller.go. Will set back to ready tomorrow!

adutra:
Ready again for review!

jsanda:
I will start reviewing today, probably in my evening. Not sure I will make it through all 68 files through 😄

pkg/secret/replicated.go
```diff
@@ -22,7 +23,7 @@ import (
 )
 
 const (
-	passwordCharacters = "!#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~"
+	passwordCharacters = "-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
adutra, line 26: This is temporary-ish, see https://github.com/thelastpickle/cassandra-reaper/pull/1144 for context.

pkg/reaper/schema.go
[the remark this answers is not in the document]
jsanda: Sorry for the delayed response. Both changes sound good.

pkg/reaper/resource.go
jsanda: Yup, that's what I had in mind. Thanks for addressing it!

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -325,23 +332,27 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
 		}
 	}
 
-	// Reconcile Stargate across all datacenters
+	kcLogger.Info("All dcs reconciled")
+
+	// Reconcile Stargate and Reaper across all datacenters
 	stargateAuthSchemaReconciled := false
+	reaperBackendSchemaReconciled := false
```
jsanda, line 339: What do you think about moving the `reconcileStargateAuthSchema` call and corresponding code in reaper_reconciler.go (lines 142 - 152) here, before the for loop? It would be something like this:

```
if cluster has stargate then
    reconcile stargate auth schema

if cluster has reaper then
    reconcile reaper schema

for each DC
    reconcile stargate
    reconcile reaper
```
While a somewhat minor change, I think it would simply the code a bit. It would eliminate some branching as well as eliminate the `stargateAuthSchemaReconciled` and `reaperBackendSchemaReconciled` vars.

adutra: Agreed, will do as you suggested.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -325,26 +336,47 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+		} else if err := r.reconcileStargateAuthSchema(ctx, kc, actualDcs[0], remoteClient, kcLogger); err != nil {
+			return ctrl.Result{RequeueAfter: r.ReconcilerConfig.LongDelay}, err
 		}
+	}
 
-		if err = remoteClient.Get(ctx, dcKey, actualDc); err == nil {
-			result, err := r.reconcileStargate(ctx, kc, dcTemplate, actualDc, logger, remoteClient, &stargateAuthSchemaReconciled)
-			if !result.IsZero() || err != nil {
-				return result, err
+	if reaperKeyspaces := r.getReaperKeyspaces(kc); len(reaperKeyspaces) > 0 {
```
jsanda, line 351: I don't think it makes sense to allow multiple Reaper keyspaces, does it? The keyspace should probably be a cluster setting only and not a DC-level setting. Alternatively, do a validation check to make sure it is the across all DC templates, but IMO keeping it the DC-level would be confusing and possibly error-prone.

adutra: Indeed, I will create a `ReaperClusterTemplate` and move the keyspaces field there. I think it would also make sense to move the secrets to this new cluster template, thus making it easier to enforce the same secret for all datacenters.

controllers/reaper/reaper_controller.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// +kubebuilder:rbac:groups=reaper.k8ssandra.io,namespace="k8ssandra",resources=reapers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=reaper.k8ssandra.io,namespace="k8ssandra",resources=reapers/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="apps",namespace="k8ssandra",resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="core",namespace="k8ssandra",resources=pods;secrets,verbs=get;list;watch
+// +kubebuilder:rbac:groups="core",namespace="k8ssandra",resources=services,verbs=get;list;watch;create
+
+func (r *ReaperReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithValues("Reaper", req.NamespacedName)
```
jsanda, line 57: The `FromContext` method accepts key/value pairs so these calls can be combined

controllers/reaper/reaper_controller.go
```diff
@@ -0,0 +1,347 @@
...
+			return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+		}
+	}
+	if !utils.CompareAnnotations(actualService, desiredService, k8ssandraapi.ResourceHashAnnotation) {
+		logger.Info("Updating Reaper Service")
+		updatedService := actualService.DeepCopy()
+		desiredService.DeepCopyInto(updatedService)
+		updatedService.SetResourceVersion(actualService.GetResourceVersion())
+		updatedService.Spec.ClusterIP = actualService.Spec.ClusterIP
+		updatedService.Spec.ClusterIPs = actualService.Spec.ClusterIPs
```
jsanda, line 257: This might be problematic. `ClusterIP` (and I think `ClusterIPs a well) is immutable.

adutra: Well that's why I did this. If you don't keep the same values that were assigned upon service creation, the update will fail. I actually think that we need to do the same for other services, e.g. Stargate services.

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -195,10 +195,16 @@ type ReaperSpec struct {
 type ReaperProgress string
 
 const (
-	ReaperProgressPending     = ReaperProgress("Pending")
-	ReaperProgressDeploying   = ReaperProgress("Deploying")
+	// ReaperProgressPending is Reaper's status when it's waiting for the datacenter to become ready.
```
jsanda, line 203: Thanks for adding this!

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -94,6 +94,24 @@ func (in *K8ssandraCluster) HasStargates() bool {
 	return false
 }
 
+// HasReapers returns true if at least one Reaper resource will be created as part of the creation
+// of this K8ssandraCluster object.
+func (in *K8ssandraCluster) HasReapers() bool {
```
jsanda, line 99: Looks like this returns `true` even if no Reaper instances are created. In other words, it returns `true` if there is no Reaper instance declared at the dc-level as long as it is declared at the cluster-level. Is that the expectation?

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -155,6 +136,30 @@ type AutoScheduling struct {
...
+	ReaperDatacenterTemplate `json:",inline"`
+
+	// The keyspace to use to store Reaper's state. Will default to "reaper_db" if unspecified. Will be created if it
+	// does not exist, and if this Reaper resource is managed by K8ssandra.
+	// +kubebuilder:default="reaper_db"
+	// +optional
+	Keyspace string `json:"keyspace,omitempty"`
+
+	// Defines the username and password that Reaper will use to authenticate CQL connections to Cassandra
+	// clusters. These credentials need to be stored on each Cassandra node. If CQL authentication is not required,
```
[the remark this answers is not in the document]
adutra: Actually this sentence is not accurate: the CQL user secret does not need to be "stored" on nodes, however the secret needs to be created beforehand so that the corresponding role can be created by cass-operator. I will fix the comment.

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -155,6 +136,30 @@ type AutoScheduling struct {
...
+	// leave this field empty. The secret must be in the same namespace as Reaper itself and must contain two keys:
+	// "username" and "password".
+	// +optional
+	CassandraUserSecretRef string `json:"cassandraUserSecretRef,omitempty"`
+
+	// Defines the username and password that Reaper will use to authenticate JMX connections to Cassandra
+	// clusters. These credentials need to be stored on each Cassandra node. If JMX authentication is not required,
```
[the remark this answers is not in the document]
adutra: Both actually. The jmx credentials will be stored on Cassandra nodes in a `jmxremote.password` file; on Reaper nodes, it will be stored in `cassandra-reaper.yml`. I will fix the comment here too.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Yes. A non-nil cluster-level Reaper means that all dcs will get a Reaper deployment inherited from the cluster-level specs. This is in line with how the corresponding `Coalesce` function is implemented.

adutra: I added a small unit test to exercise the method's contract.

jsanda: Thanks for the clarification.

jsanda (approved):
Nice work!

PR #220, opened 2021-11-15
Fix cass-operator deployments

adutra:
**What this PR does**:

This commit replaces `config/default` with `config/deployments/default` when generating kustomization for cass-operator.

This fixes the incorrect deployment of the controller-manager pod for cass-operator.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/test/framework/e2e_framework.go
+++ b/test/framework/e2e_framework.go
@@ -127,7 +127,7 @@ func generateCassOperatorKustomization(namespace string) error {
 kind: Kustomization
 
 resources:
-- github.com/k8ssandra/cass-operator/config/default?ref=v1.8.0
+- github.com/k8ssandra/cass-operator/config/deployments/default?ref=211eac97af729b03d8551ebccb88c6e7e376c567
 namespace: {{ .Namespace }}
 `
 	k := Kustomization{Namespace: namespace}


test/framework/e2e_framework.go
```diff
@@ -127,7 +127,7 @@ func generateCassOperatorKustomization(namespace string) error {
 kind: Kustomization
 
 resources:
-- github.com/k8ssandra/cass-operator/config/default?ref=v1.8.0
+- github.com/k8ssandra/cass-operator/config/deployments/default?ref=211eac97af729b03d8551ebccb88c6e7e376c567
```
burmanm, line 130: Why not just ref master?

adutra: @jsanda told me to use a specific ref instead. This ref is the commit that brings more management-api endpoints. Btw it doesn't really matter here, the manifests did not change between these 2 refs.

One good point about using this specific ref is that it becomes easier to search-replace it later on after the next cass-operator release.

burmanm: It also prevents seeing in the tests if cass-operator changes and breaks k8ssandra-operator. After that commit there's already been 3 other commits to master.

jsanda: I am not in favor of referencing cass-operator master. It can lead to an unstable main branch in k8ssandra-operator. We could set up a CI job to test against the latest cass-operator to see if it breaks anything.

burmanm: But this is just for test?

jsanda: I still favor referencing the same version for tests. I fear that we could get into a situation where there is some divergence that could lead subtle, undetected errors.

PR #224, opened 2021-11-30
K8SSAND-995 Mgmt api heap

### LINKED ISSUE #212: K8SSAND-995 ⁃ Make the management-api heap size configurable
**What is missing?**
The heap size for the management-api had been hard coded to either 128M or 256M. Changes were made to make it configurable via an env var. The default was reduced to 16M which helped with the resource challenges in e2e tests. The default setting caused some OutOfMemoryErrors in non-test environments. The default has been reverted back to 128M.

I was doing some manual testing and having problems with my Cassandra pods becoming unstable. Then I remembered about the heap size change. After increasing the memory limits things became stable again in my cluster.

**Why do we need it?**
We need to make the management-api's heap size configurable so that we can reduce the memory requirements of the `cassandra` container in dev/test enviromments. This may be generally useful as well in non-test environments as folks look to tune things.

**Goals**
The expectation is that properties are added to the K8ssandraCluster spec for configuring the heap size. The fact that the heap size is actually configured via env vars on the `cassandra` container should be considered an implementation detail with which the user need not be concerned. 

To be consistent with the approach taken thus far it probably makes sense to make the heap size configurable both at the cluster and dc levels. This means properties should be added in both `CassandraClusterTemplate` and in `CassandraDatacenterTemplate` with the latter taking precedence over the former.

We can implement this without any changes in cass-operator by configuring the heap size via the PodTemplateSpec property. My preference however would be to first implement the above changes in cass-operator and then have k8ssandra-operator simply set the new property in the CassandraDatacenterSpec.

* K8ssandra Operator version:

v1.0.0-alpha.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-995) by [Unito](https://www.unito.io)
┆Fix Versions: k8ssandra-operator-v1.0.0-alpha.2
┆Issue Number: K8SSAND-995
┆Priority: Medium

Miles-Garnsey:
**What this PR does**:

* Make the management API heap size configurable via a field in the `K8ssandraCluster`.
* Add error handling for cases where `StorageConfig` or `CassandraConfig` are not defined in a `DatacenterConfig`. (These conditions previously caused a runtime panic).
* Add unit tests for all of the above.
* Add new test function to get a minimal Cassandra `DatacenterConfig`.

Fixes #212.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -227,6 +227,10 @@ type CassandraDatacenterTemplate struct {
 	// deploying Reaper in this datacenter.
 	// +optional
 	Reaper *reaperapi.ReaperDatacenterTemplate `json:"reaper,omitempty"`
+	// MgmtAPIHeap defines the amount of memory (in java format - i.e. an int followed by a G or M) devoted to the management
+	// api heap.
+	// +optional
+	MgmtAPIHeap string `json:"mgmtAPIHeap,omitempty"`
 }
 
 type EmbeddedObjectMeta struct {
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -171,6 +171,11 @@ type CassandraClusterTemplate struct {
 	// Datacenters a list of the DCs in the cluster.
 	// +optional
 	Datacenters []CassandraDatacenterTemplate `json:"datacenters,omitempty"`
+
+	// MgmtAPIHeap defines the amount of memory devoted to the management
+	// api heap.
+	// +optional
+	MgmtAPIHeap *resource.Quantity `json:"mgmtAPIHeap,omitempty"`
 }
 
 // +kubebuilder:pruning:PreserveUnknownFields
@@ -227,10 +232,10 @@ type CassandraDatacenterTemplate struct {
 	// deploying Reaper in this datacenter.
 	// +optional
 	Reaper *reaperapi.ReaperDatacenterTemplate `json:"reaper,omitempty"`
-	// MgmtAPIHeap defines the amount of memory (in java format - i.e. an int followed by a G or M) devoted to the management
+	// MgmtAPIHeap defines the amount of memory devoted to the management
 	// api heap.
 	// +optional
-	MgmtAPIHeap string `json:"mgmtAPIHeap,omitempty"`
+	MgmtAPIHeap *resource.Quantity `json:"mgmtAPIHeap,omitempty"`
 }
 
 type EmbeddedObjectMeta struct {
--- a/config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
+++ b/config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
@@ -203,10 +203,13 @@ spec:
                           - name
                           type: object
                         mgmtAPIHeap:
-                          description: MgmtAPIHeap defines the amount of memory (in
-                            java format - i.e. an int followed by a G or M) devoted
+                          anyOf:
+                          - type: integer
+                          - type: string
+                          description: MgmtAPIHeap defines the amount of memory devoted
                             to the management api heap.
-                          type: string
+                          pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                          x-kubernetes-int-or-string: true
                         networking:
                           description: Networking enables host networking and configures
                             a NodePort ports.
@@ -5401,6 +5404,14 @@ spec:
                       type: object
                       x-kubernetes-preserve-unknown-fields: true
                     type: array
+                  mgmtAPIHeap:
+                    anyOf:
+                    - type: integer
+                    - type: string
+                    description: MgmtAPIHeap defines the amount of memory devoted
+                      to the management api heap.
+                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                    x-kubernetes-int-or-string: true
                   networking:
                     description: Networking enables host networking and configures
                       a NodePort ports.
--- a/config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
+++ b/config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
@@ -202,6 +202,11 @@ spec:
                           required:
                           - name
                           type: object
+                        mgmtAPIHeap:
+                          description: MgmtAPIHeap defines the amount of memory (in
+                            java format - i.e. an int followed by a G or M) devoted
+                            to the management api heap.
+                          type: string
                         networking:
                           description: Networking enables host networking and configures
                             a NodePort ports.
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -41,7 +41,7 @@ func isCassandra4(version string) bool {
 }
 
 func (c config) MarshalJSON() ([]byte, error) {
-	config := make(map[string]interface{}, 0)
+	config := make(map[string]interface{})
 
 	if c.CassandraYaml != nil {
 		if isCassandra4(c.cassandraVersion) {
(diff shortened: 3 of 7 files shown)


apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -227,6 +227,10 @@ type CassandraDatacenterTemplate struct {
 	// deploying Reaper in this datacenter.
 	// +optional
 	Reaper *reaperapi.ReaperDatacenterTemplate `json:"reaper,omitempty"`
+	// MgmtAPIHeap defines the amount of memory (in java format - i.e. an int followed by a G or M) devoted to the management
+	// api heap.
+	// +optional
+	MgmtAPIHeap string `json:"mgmtAPIHeap,omitempty"`
```
burmanm, line 233: This looks very much out of place at this level. Even JvmOptions would be a better place since it quite directly affects the JVM properties in the Cassandra container.

pkg/cassandra/config.go
```diff
@@ -75,32 +75,36 @@ func (c config) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&config)
 }
 
-func newConfig(apiConfig *api.CassandraConfig, cassandraVersion string) config {
-	config := config{cassandraVersion: cassandraVersion}
+func newConfig(apiConfig *api.CassandraConfig, cassandraVersion string) (config, error) {
```
burmanm, line 78: These types and name refactoring look odd. In this place the config name shadowing has been modified, but in other places in the same file they haven't been.

pkg/cassandra/datacenter.go
```diff
@@ -88,9 +93,41 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
+	if dc.Spec.PodTemplateSpec == nil {
+		dc.Spec.PodTemplateSpec = &corev1.PodTemplateSpec{}
+	}
+	if len(dc.Spec.PodTemplateSpec.Spec.Containers) == 0 {
+		dc.Spec.PodTemplateSpec.Spec.Containers = []corev1.Container{{Name: "cassandra"}}
+	}
+	var cassIndex int
+	for i, container := range dc.Spec.PodTemplateSpec.Spec.Containers {
+		if container.Name == "cassandra" {
+			cassIndex = i
```
burmanm, line 118: Missing break

pkg/cassandra/datacenter.go
```diff
@@ -88,9 +93,41 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
+		if err := SetMgmtAPIHeap(dc, template.MgmtAPIHeap); err != nil {
+			return nil, err
+		}
+	}
+
 	return dc, nil
 }
 
+// SetMgmtAPIHeap sets the management API heap size on a CassandraDatacenter
+func SetMgmtAPIHeap(dc *cassdcapi.CassandraDatacenter, heapSize string) error {
```
burmanm, line 106: In which case does it return err? I don't see any.

burmanm (changes requested):

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: I don't think it should be in JvmOptions because that refers to Cassandra's JVM options and I think we should be clear that this is management API heap related, not Cassandra heap related. 

I could put it [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/1dfe518f9add98a327825f7aee8cbb668aa54e17/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L250) wdyt?

pkg/cassandra/config.go
Miles-Garnsey: I can't have `config` shadowed because I need to use it again [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/1dfe518f9add98a327825f7aee8cbb668aa54e17/pkg/cassandra/config.go#L82).

What are you asking me to change here?

pkg/cassandra/datacenter.go
Miles-Garnsey: Damn, good call Michael! Fixed.

pkg/cassandra/datacenter.go
Miles-Garnsey: It doesn't, you're right. Fixed.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jsanda: I agree that it should not be part
Of JvMOptions since it does not apply to Cassandra itself.

pkg/cassandra/config.go
Miles-Garnsey: NB: I actually can't see another instance where the internal `config` struct is shadowed by a variable named `config`.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: Counterargument to my proposal above [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/1dfe518f9add98a327825f7aee8cbb668aa54e17/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L150). If `SystemLoggerResources` is set in the `CassandraClusterTemplate` is it better to also set the management API resources in the same place/at the same level?

We should also consider that ideally at some point it would be good to separate the Cassandra and management API processes into two containers.

pkg/cassandra/config.go
burmanm: I mean you should **NOT** shadow it. You removed shadowing in one place, but not in others (which you still modified). Consistency.

pkg/cassandra/datacenter.go
```diff
@@ -88,9 +93,41 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
+	// TODO: it would be nice to have a generic `StrategicMergePatch` method which produces merged API objects instead
```
jsanda, line 107: If we were exposing a PodTemplateSpec with k8ssandra-operator and passing it through then I could see using a generic merge approach, but we're not doing that. In fact I would ideally prefer to minimize the use the PodTemplateSpec in the CassandraDatacenter to the extent we can.

pkg/cassandra/config.go
Miles-Garnsey: I'm still not grokking what you're asking for I'm sorry. Are you suggesting that I should replace `config` with `cfg` for local variables everywhere in this file? The only functions I've modified are `newConfig` and `CreateJsonConfig`. The latter already had a variable called `cfg` so if you're nitpicking with me about consistency, the code is no more inconsistent now than it was before.

The other functions (except `MarshalJSON`) don't use the `config` struct, so the shadowing doesn't matter for them. I do not believe that shadowing must be avoided in all cases.

The real problem here is that `config{}` is a bad name. It is up there with anti-patterns like `Xmanager` and `a`. In fact, I don't understand what the purpose of the internal `config{}` is. It seems to be designed purely as a place to attach `MarshalJSON()`, but that function isn't called anywhere.

But this isn't my code, and I'm not going to go at it with a hacksaw. Closing this item.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: So you want the same field to be available in both `CassandraClusterTemplate` and in `CassandraDatacenterTemplate`? I currently have it in `CassandraDatacenterTemplate` only.

pkg/cassandra/config.go
jsanda: >It seems to be designed purely as a place to attach `MarshalJSON()`, but that function isn't called anywhere.

The `config` struct, while it may suffer from a major naming anti-pattern up there with the likes of `Xmanager` and `a` :smile, actually provides some key logic. If you take a look at `config_test.go` you will find tests that exercise and verify this functionality. It's a good starting point to understanding the code involved.


The `config` struct is an adapter to provide custom marshaling. The `Marshaler` interface is defined in the `json` package. The `json.Marshal` function traverses the value recursively. If it is encounters an object that implements `Marshaler` it calls its `MarshalJSON` function.

If you look at the `MarshalJSON` function you will notice that it does several things:

* If the version is Cassandra 4, remove properties that are only relevant to Cassandra 3.x 
* If `NumTokens` is `nil` set a default value based on Cassandra version
* Make sure appropriate `jvm-options` or `jvm-server-options` is used based on Cassandra version

>But this isn't my code, and I'm not going to go at it with a hacksaw

It's not your code. It's not my code. It's our code. And I certainly agree that you shouldn't go at it with a hacksaw, at least not until you have a better understanding of it 😉

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jsanda: Yes, it should be in both places.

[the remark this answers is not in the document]
Miles-Garnsey: Done.

pkg/cassandra/datacenter.go
Miles-Garnsey: Done.

config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
```diff
@@ -202,6 +202,11 @@ spec:
                           required:
                           - name
                           type: object
+                        mgmtAPIHeap:
+                          description: MgmtAPIHeap defines the amount of memory (in
```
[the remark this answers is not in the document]
Miles-Garnsey: Done.

Miles-Garnsey:
Changes made:
* Use resource.Quantity for management API heap size. 
* Add `MgmtAPIHeap` to `CassandraClusterTemplate`.
* Removed a comment John objects to.

Will not fix:
* Changes to config variable naming.
* Moving `MgmtAPIHeap` from CassandraDatacenterTemplate to a lower level.

pkg/cassandra/config.go
```diff
@@ -75,32 +75,36 @@ func (c config) MarshalJSON() ([]byte, error) {
...
+	cfg := config{cassandraVersion: cassandraVersion}
+
+	if apiConfig == nil {
+		return config{}, DCConfigIncomplete{"CassandraConfig"}
+	}
 
 	if apiConfig.CassandraYaml == nil {
-		config.CassandraYaml = &api.CassandraYaml{}
+		cfg.CassandraYaml = &api.CassandraYaml{}
```
jsanda, line 86: Thanks for changing the variable names. It was definitely a bit confusing as was originally written. I'll pass that along to the author 😉

jsanda (approved):
Changes look good to go. Thanks!

pkg/cassandra/config.go
Miles-Garnsey: @burmanm merging is blocked for me because of your requested changes. Can you please approve or talk to @jsanda  about whatever changes are required to get consensus with him.

I have manually tested and this is ready to go once Github gets rid of the 'changes requested' block.

PR #226, opened 2021-12-01
Improve seeds handling

### LINKED ISSUE #210: K8SSAND-994 ⁃ Improve seeds handling
**What is missing?**
There needs to be better management of seeds across the cluster. 

cass-operator creates a seeds headless service for each CassandraDatacenter. It applies the `cassandra.datastax.com/seed-node: "true"` label to pods. Cassandra is configured to use a custom seeds resolver that does a hostname lookup against the seeds service. This takes care of the local DC but does not address remote DCs.

The CassandraDatacenter spec defines an `AdditionalSeedsService` which expects a list of IP addresses or hostnames. `AdditionalSeedsService` can be used to specify seeds in remote DCs. This is what k8ssandra-operator does. Here is an example of what gets passed to the config builder init container to configure `seeds` in cassandra.yaml (where the cluster name is test):

```
"seeds":"test-seed-service,test-dc1-additional-seed-service
```

There are a couple problems with the current implementation. First, the seeds list should be centrally managed. Ideally, there would be a single seeds service shared across DCs. Until we have multi-cluster services with multi-cluster DNS, this will be hard to accomplish. We have centralized management to an extent with k8ssandra-operator managing the additional seeds. This brings me to the next issue.

cass-operator creates and manages the Endpoints object for the additional seeds service. This works perfectly fine. The problem is the current approach is in efficient for a multi-DC cluster. Let's say we have 3 DCs - dc1, dc2, and dc3. As each DC becomes ready k8ssandra-operator will update `AdditionalSeeds`. Now let's say we apply some to dc2 that results in a rolling restart and cass-operator updates seeds (by toggling the value of `cassandra.datastax.com/seed-node:` on dc2 pods). This will trigger a cascade of reconciliations across all DCs as k8ssandra-operator goes through and updates `AdditionalSeeds`.

If k8ssandra-operator manages the Endpoints objects for the additional seeds service, then there is no need update `AdditionalSeeds`. This will allow us to avoid the cascading reconciliations since we aren't updating the CassandraDatacenters.

Here is why this works. cass-operator automatically creates the additional seeds service. During its reconciliation loop it refreshes seeds in pods where Cassandra has been started. cass-operator calls the `api/v0/ops/seeds/reload` endpoint of the management-api which results in the seed provider being called. It does the hostname lookup on the k8s seeds and additional seeds services to get the up to date seeds list. 

* K8ssandra Operator version:

v1.0.0-alpha.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-994) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-994
┆Priority: Medium

jsanda:
**What this PR does**:
* Create and manage the additional seeds service Endpoints objects in k8ssandra-operator
* Do not set `AdditionalSeeds` property on the CassandraDatacenters
* Add unit tests
* Update integration tests

Fixes #210

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/seeds.go
```diff
@@ -0,0 +1,79 @@
...
+	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// findSeeds queries for pods labeled as seeds. It does this for each DC, across all
+// clusters.
+func (r *K8ssandraClusterReconciler) findSeeds(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) ([]corev1.Pod, error) {
```
burmanm, line 16: Shouldn't this exclude pods from the same datacenter as the target updated seed? As in, return map[datacenter][]corev1.Pod for example.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -266,6 +271,54 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+		//
+		// TODO Refactor the if block into a separate method/function.
+		// Note: Some other refactoring needs to be done first so that this can be done cleanly.
+		if len(seeds) > 0 {
+			desiredEndpoints := newEndpoints(desiredDc, seeds)
+			actualEndpoints := &corev1.Endpoints{}
+			endpointsKey := client.ObjectKey{Namespace: desiredEndpoints.Namespace, Name: desiredEndpoints.Name}
+
+			if err = remoteClient.Get(ctx, endpointsKey, actualEndpoints); err == nil {
+				if actualHash, found := actualDc.Annotations[api.ResourceHashAnnotation]; !(found && actualHash == desiredDcHash) {
```
adutra, line 288: 1. Shouldn't we be comparing the annotations on `actualEndpoints`?
2. Can be simplified to:

```
if !utils.CompareAnnotations(actualDc, desiredDc, api.ResourceHashAnnotation) {
```

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -266,6 +271,54 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+					logger.Info("Updating Endpoints", "K8SContext", dcTemplate.K8sContext, "CassandraDatacenter", dcKey, "Endpoints", endpointsKey)
+					actualEndpoints := actualEndpoints.DeepCopy()
+					resourceVersion := actualEndpoints.GetResourceVersion()
+					desiredEndpoints.DeepCopyInto(actualEndpoints)
+					actualEndpoints.SetResourceVersion(resourceVersion)
+					if err = remoteClient.Update(ctx, actualEndpoints); err != nil {
+						logger.Error(err, "Failed to update Endpoints",
+							"K8SContext", dcTemplate.K8sContext,
+							"CassandraDatacenter", dcKey,
```
adutra, line 297: 1. `logger` already has a `CassandraDatacenter` entry.
2. `logger` doesn't have a `K8SContext` entry, maybe you should add it on line 257?

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -300,26 +353,6 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
 
 			logger.Info("The datacenter is ready")
 
-			endpoints, err := r.SeedsResolver.ResolveSeedEndpoints(ctx, actualDc, remoteClient)
```
adutra, line 303: Do we still need the `SeedsResolver` field? I don't see it used anywhere.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -829,16 +829,6 @@ func createMultiDcCluster(t *testing.T, ctx context.Context, f *framework.Framew
 	err = f.Get(ctx, dc2Key, dc2)
 	require.True(err != nil && errors.IsNotFound(err), "dc2 should not be created until dc1 is ready")
 
-	seedsResolver.callback = func(dc *cassdcapi.CassandraDatacenter) ([]string, error) {
```
adutra, line 832: Same: do we still need `seedsResolver` and `fakeSeedsResolver`?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -693,7 +693,7 @@ func parseCassandraConfig(config *api.CassandraConfig, serverVersion string, sys
 // k8s clusters. It also verifies that status updates are made to the K8ssandraCluster.
 func createMultiDcCluster(t *testing.T, ctx context.Context, f *framework.Framework, namespace string) {
 	require := require.New(t)
-	assert := assert.New(t)
+	//assert := assert.New(t)
```
adutra, line 696: Could be removed.

controllers/k8ssandra/seeds.go
```diff
@@ -0,0 +1,79 @@
...
+		})
+	}
+
+	ep.Subsets = []corev1.EndpointSubset{
+		{
+			Addresses: addresses,
+		},
+	}
+
+	ep.Annotations[api.ResourceHashAnnotation] = utils.DeepHashString(ep)
```
adutra, line 76: You could use `utils.AddHashAnnotation()` here.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -266,6 +270,64 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+						return ctrl.Result{}, err
+					}
+				}
+			}
+		} else {
+			if errors.IsNotFound(err) {
+				// If we have seeds then we want to go ahead and create the Endpoints. But
+				// if we don't have seeds, then we don't need to do anything for a couple
+				// of reasons. First, no seeds means that cass-operator has not labeled any
+				// pods as seeds which would be the case when the CassandraDatacenter is f
```
adutra, line 314:
```suggestion
				// pods as seeds which would be the case when the CassandraDatacenter is
```

adutra: OK but shouldn't we requeue if endpoints not found and len(seeds) == 0? Is there any value in keeping reconciling the DC without seeds?

adutra (approved):

controllers/k8ssandra/k8ssandracluster_controller.go
jsanda: We have to proceed here; otherwise, we would never create the first CassandraDatacenter.

controllers/k8ssandra/seeds.go
jsanda: I am going mark this resolved based on the conversation with @adejanovski where he said having redundant seeds shouldn't be a problem. @burmanm please feel free to reopen if you want to further discuss.

PR #239, opened 2021-12-07
add cass-operator dev kustomization

### LINKED ISSUE #237: K8SSAND-1089 ⁃ Add cass-operator dev kustomization
**What is missing?**
We need kustomize support for deploying local, dev builds (i.e., `latest`) of cass-operator.

**Why do we need it?**
I was working on #21 and needed to make changes in cass-operator as well. Initially I did things manually which included the following:

* Make cass-operator code change
* Build cass-operator image
* Load image into kind clusters
* Update cass-operator Deployment in each cluster to use `latest` 

The last steps is actually rather cumbersome and error-prone when doing it by hand. It is very likely that there will be other occasions when we want to use local, dev builds of cass-operator. It makes sense to provide some automation.

**Environment**

* K8ssandra Operator version:
1.0.0-alpha.2

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1089) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1089
┆priority: Medium

jsanda:
**What this PR does**:
Adds kustomize for deploying local, dev builds of cass-operator.

Fixes #237

**Checklist**
- [x] Changes manually tested


config/deployments/control-plane-cass-operator-dev/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: k8ssandra-operator
+
+resources:
+  - ../control-plane
+
+images:
+  - name: k8ssandra/cass-operator
```
burmanm, line 10: While this might change the image to latest, it will not actually necessarily work, since the other changes from cass-operator master are not picked up with this approach. This includes RBAC changes, CRD changes etc

jsanda: Good points. It worked for my particular situation but not a robust solution. This needs some more work.

jsanda: @burmanm what do you think about changing the `kustomization.yaml` so that it points to the local cass-operator repo? Obviously people might have it stored in different places relative to their k8ssandra-operator repo, but they could work around that with a symlink.

burmanm: Yep, I guess ../cass-operator would be fine default.

burmanm: (add as many .. as necessary)

Makefile
```diff
@@ -149,63 +156,55 @@ kind-load-image:
...
+	$(KUSTOMIZE) build config/deployments/control-plane$(DEPLOY_TARGET) | kubectl apply --server-side --force-conflicts -f -
 
 single-reload: build manifests kustomize docker-build kind-load-image cert-manager
 	kubectl config use-context kind-k8ssandra-0
-ifeq ($(CLUSTER_SCOPE),true)
-	$(KUSTOMIZE) build config/deployments/control-plane-cluster-scope | kubectl apply --server-side --force-conflicts -f -
-else
-	$(KUSTOMIZE) build config/deployments/control-plane | kubectl apply --server-side --force-conflicts -f -
-endif
+	$(KUSTOMIZE) build config/deployments/control-plane$(DEPLOY_TARGET) | kubectl apply --server-side --force-conflicts -f -
```
burmanm, line 163: These control-plane-a, control-plane-b, control-plane-c etc directories feel odd. Wouldn't it be easier if it was a tree struct? control-plane/a, /b, /c etc

config/deployments/control-plane-cass-operator-dev/kustomization.yaml
jsanda: I have been using this extensively to work with your rebuild branch. It makes things a lot easier.

Makefile
jsanda: >Wouldn't it be easier if it was a tree struct?

How so? If we end up with a proliferation of control-plane-xxx dirs, then yeah from an organization standpoint it would be better. From a functional standpoint, I am not sure how there would be a difference.

jsanda: Discussed offline briefly with @burmanm and I agree that the tree structure is better. I'll update the PR accordingly.

config/deployments/control-plane/cass-operator-dev/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: k8ssandra-operator
+
+resources:
+  - ../../default
+  - ../../../../../cass-operator/config/deployments/default
+
+images:
```
burmanm, line 10: Is this actually needed? The image is set to newTag: latest in the cass-operator in any case. Unless you actually point to a branch in your local repo, but in that case you'll run into issues since latest does not necessarily match other configs..

burmanm: Can be addressed later.

burmanm (approved):

burmanm:
The test failures seem random (different test failing), so I doubt they have anything to do with this change

PR #241, opened 2021-12-08
Cassdc prometheus integration [K8SSAND-1083]

### LINKED ISSUE #234: K8SSAND-1083 ⁃ Add configuration points + logic to deploy service monitors when a CassandraDatacenter is deployed.
When a CassandraDatacenter is deployed, a Prometheus ServiceMonitor should be deployed alongside it (including if the deployment is occurring in a remote DC).

Not in scope:

* Validating whether Prometheus is actually deployed, handling cases where it is not.
* Deploying a ServiceMonitor for stargate when it is deployed.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1083) by [Unito](https://www.unito.io)
┆epic: Prometheus integration
┆friendlyId: K8SSAND-1083
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Adds an integration to install a Prometheus operator ServiceMonitor when a CassandraDatacenter is deployed, to monitor that DC.

**What we are not doing**
* Deploying a ServiceMonitor to provide Stargate telemetry
* Providing protection in the case where a user has requested Prometheus telemetry but Prometheus CRDs are not installed in the cluster.
* Adding hooks into status to report whether telemetry has successfully come up.
* Adding RBAC rules to the k8ssandra helm charts to cater to the ServiceMonitor resources which we are creating.

Fixes #234 

**Checklist**
- [x] Changes manually tested


Miles-Garnsey:
This is ready for preliminary review. I am still working to add more tests for the reconciler and in `telemetry_methods.go` but the reconciliation logic and api are there - manual tests suggest the changes are good to go.

The existing test failures are all intentional, they are failing with a 'not implemented' error to ensure that we remember to add them prior to merge. 

I'm flagging as ready now, because this is quite a large PR (28 files 🙀 ) so I expect it will take some poor soul a while to get to it/through it.

apis/telemetry/v1alpha1/telemetry_methods.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/go-logr/logr"
+	promapi "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+)
+
+// Merge takes an object a and merges another object, b's values into it, overwriting any which conflict.
+func (a *TelemetrySpec) Merge(b *TelemetrySpec) *TelemetrySpec {
```
adutra, line 16: We've been using the word `Coalesce` for this kind of function, and indeed it's brittle because it has to be updated manually when the struct changes.

apis/telemetry/v1alpha1/telemetry_methods.go
```diff
@@ -0,0 +1,99 @@
...
+	// patch logic.
+	out := PrometheusTelemetrySpec{}
+	if a == nil && b == nil {
+		return nil
+	} else if b == nil {
+		return a
+	} else if a == nil {
+		return b
+	}
+	if a.CommonLabels != nil {
```
adutra, line 49: You could use `utils.MergeMap` here.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -73,6 +74,7 @@ type K8ssandraClusterReconciler struct {
 // +kubebuilder:rbac:groups=reaper.k8ssandra.io,namespace="k8ssandra",resources=reapers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,namespace="k8ssandra",resources=pods;secrets,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,namespace="k8ssandra",resources=endpoints,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
```
adutra, line 77: Missing `namespace="k8ssandra",` ?

controllers/k8ssandra/telemetry_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !promInstalled {
+		return ctrl.Result{}, nil
+	}
+	// Determine if we want a cleanup or a resource update.
+	switch {
+	case mergedSpec == nil:
+		logger.Info("Telemetry not present for CassDC, will delete resources", "datacenter", actualDc.Name)
```
adutra, line 54: Use key `"CassandraDatacenter"` instead of `"datacenter"` for consistency.

controllers/k8ssandra/telemetry_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+	promInstalled, err := telemetryapi.IsPromInstalled(remoteClient, logger)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !promInstalled {
+		return ctrl.Result{}, nil
+	}
+	// Determine if we want a cleanup or a resource update.
+	switch {
+	case mergedSpec == nil:
```
adutra, line 53: The first 3 case branches are fairly identical, and similar to the default branch. Is it really useful to distinguish all of them?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -81,9 +82,10 @@ type K8ssandraClusterCondition struct {
 
 // K8ssandraStatus defines the observed of a k8ssandra instance
 type K8ssandraStatus struct {
-	Cassandra *cassdcapi.CassandraDatacenterStatus `json:"cassandra,omitempty"`
-	Stargate  *stargateapi.StargateStatus          `json:"stargate,omitempty"`
-	Reaper    *reaperapi.ReaperStatus              `json:"reaper,omitempty"`
+	Cassandra          *cassdcapi.CassandraDatacenterStatus `json:"cassandra,omitempty"`
+	CassandraTelemetry *telemetryapi.TelemetryStatus        `json:"cassandraTelemetry,omitempty"`
```
adutra, line 86: Caution here: currently if you add a new field to the `K8ssandraStatus` struct, this has potential impacts in all controller methods called `setStatusForX` and `removeXStatus`. You will need to modify all of them. I know this isn't ideal, we might want to eventually merge all these methods into one single method that updates all fields at once.

apis/telemetry/v1alpha1/telemetry_methods.go
```diff
@@ -0,0 +1,99 @@
...
+			out.CommonLabels[k] = v
+		}
+	}
+	if b.Enabled == nil {
+		out.Enabled = a.Enabled
+	}
+	return &out
+}
+
+func (tspec *TelemetrySpec) IsValid(client client.Client, logger logr.Logger) (bool, error) {
```
adutra, line 65: I'm not sure it's a good idea to import `client.Client` into an api package.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+      - __name__
+      - prom_name
+      targetLabel: __name__
+    - action: labeldrop
+      regex: prom_name`
+
+// mustLabels() returns the set of labels essential to managing the Prometheus resources. These should not be overwritten by the user.
+func (cfg CassPrometheusResourcer) mustLabels() map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/managed-by": "k8ssandra-operator",
```
adutra, line 321: We have constants for some of these labels in `apis/k8ssandra/v1alpha1/constants.go`. Also have a look at `pkg/utils/labels.go`.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+		"app.kubernetes.io/part-of":    "k8ssandra",
+		"k8ssandra.io/cluster":         cfg.ClusterName,
+		"k8ssandra.io/datacenter":      cfg.DataCenterName,
+		"k8ssandra.io/monitors":        "cassandra-datacenter",
+	}
+}
+
+// NewServiceMonitor returns a Prometheus operator ServiceMonitor resource.
+func (cfg CassPrometheusResourcer) NewServiceMonitor() (promapi.ServiceMonitor, error) {
```
adutra, line 330: `ServiceMonitor` is a huge struct, maybe pass it by reference instead of by value? Would also make sense in case of an error, as you would return nil instead of the zero value.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+	// validate the object we're being passed.
+	if cfg.CassandraNamespace == "" || cfg.ServiceMonitorName == "" || cfg.DataCenterName == "" || cfg.ClusterName == "" {
+		return promapi.ServiceMonitor{}, TelemetryConfigIncomplete{}
+	}
+	// Overwrite any CommonLabels the user has asked for if they conflict with the labels essential for the functioning of the operator.
+	var mergedLabels = map[string]string{}
```
adutra, line 336: Same: use `utils.MergeMap`.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+
+// CreateResources executes the creation of the desired Prometheus resources on the cluster.
+func (cfg CassPrometheusResourcer) UpdateResources(ctx context.Context, client runtimeclient.Client, owner *k8ssandraapi.K8ssandraCluster) error {
+	desiredSM, err := cfg.NewServiceMonitor()
+	if err != nil {
+		return err
+	}
+	cfg.CassTelemetryResourcer.Logger.Info("checking whether Prometheus ServiceMonitor for Cassandra already exists")
+	// Logic to handle case where SM does not exist.
+	var actualSM promapi.ServiceMonitor
```
adutra, line 399: It would be simpler IMO to use pointers for variables like `desiredSM` and `actualSM`.

apis/telemetry/v1alpha1/telemetry_methods.go
adutra: We could maybe look into https://github.com/kubernetes/gengo to see if that tool could help us generate these functions automatically.

Miles-Garnsey: Indeed. I want to look into how this can be done, ideally with reference to how k8s [does](https://github.com/kubernetes/apimachinery/blob/0cb2c3db59f8f8fcab4dec3a648529a6e4c10010/pkg/util/strategicpatch/patch.go#L94) it. 

If that fails though then your gengo idea is a good one (and I have a few other places I may be able to use that, so thanks for sharing!)

For now, I've used the same approach we use elsewhere which is just to manually construct the `Merge()` function.

apis/telemetry/v1alpha1/telemetry_methods.go
Miles-Garnsey: That wouldn't hurt, done!

Miles-Garnsey: Well, I thought it was done. Importing utils into the APIs seems to lead to circular dependancies, so this isn't possible unfortunately.

controllers/k8ssandra/k8ssandracluster_controller.go
Miles-Garnsey: Since we decided to make this single namespace, we can indeed re-scope the permissions down to the k8ssandra namespace. Done.

controllers/k8ssandra/telemetry_reconciler.go
Miles-Garnsey: I don't think this key is actually required at all given it is added to the logger passed into the reconciler. Removed, thanks for calling it out.

controllers/k8ssandra/telemetry_reconciler.go
Miles-Garnsey: It is, because if we simply wrote `case mergedSpec.Prometheus.Enabled:` but the `Prometheus` part of the path was nil, we'd get a segfault AGAIK. 

I have been agonising over this however, and I think that one simplification we can make is to make `mergedSpec.Prometheus.Enabled` a bool instead of a bool pointer. That way it is not optional and is always defined when `Prometheus` is non-nil. This eliminates one branch from this switch statement.

apis/telemetry/v1alpha1/telemetry_methods.go
Miles-Garnsey: Can you expand on why that might be harmful?

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: Good idea, done.

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: Couldn't hurt to try. Done, although I'll have to put this through another round of testing to ensure I haven't introduced any subtle bugs here.

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: I can do that in this case. Done.

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: Sure, I don't see why not! Done.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: I think I'm just going to have to suck that up regrettably.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+	for k, v := range cfg.CommonLabels {
+		mergedLabels[k] = v
+	}
+	mustLabels := cfg.mustLabels()
+	for k, v := range mustLabels {
+		mergedLabels[k] = v
+	}
+	var endpointHolder promapi.ServiceMonitor
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	_, _, err := decode([]byte(endpointString), nil, &endpointHolder)
```
jsanda, line 341: Since `endpointString` is static, couldn't we just didn't this at package initialization time in an `init` function? Actually, is it necessary at all at runtime? Couldn't this be done in a unit test?

jsanda (changes requested):

controllers/k8ssandra/k8ssandracluster_controller.go
Miles-Garnsey: > Just so you are aware, the operator does support watching multiple namespaces, i.e., cluster-scoped

Thanks for the info. 

> Just so you are aware, the operator does support watching multiple namespaces, i.e., cluster-scoped. The changes to use ClusterRoles and ClusterRoleBindings is done in kustomize. I think the Helm chart needs to be updated to support it.

Kustomize changes already made (via `make manifests`), but what helm chart are we talking about here?

apis/telemetry/v1alpha1/telemetry_methods.go
Miles-Garnsey: OK, I'll move this elsewhere. It seemed nice to be able to write `MyTelemetrySpec.IsValid()` but having to write `TelemetrySpecIsValid(MyTelemetrySpec)` is not the end of the world...

controllers/k8ssandra/telemetry_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
+	ctx context.Context,
+	kc *api.K8ssandraCluster,
+	dcTemplate api.CassandraDatacenterTemplate,
+	actualDc *cassdcapi.CassandraDatacenter,
+	logger logr.Logger,
+	remoteClient client.Client,
+) (ctrl.Result, error) {
+	logger.Info("reconciling telemetry", "cluster", kc.Name, "datacenter", actualDc.Name)
+	mergedSpec := kc.Spec.Cassandra.CassandraTelemetry.Merge(dcTemplate.CassandraTelemetry)
```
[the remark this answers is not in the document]
Miles-Garnsey: > We are only deploying ServiceMonitors, which is basically just configuration. Why do we this in the status at all?

The reason I think status may be important is that any telemetry we integrate with runs the risk that the dependancies (e.g. Prometheus operator) are not installed. In my view, things are very likely to go wrong at this edge of the system, because we have more external dependancy and less control over the things we are integrating with.

The logical mitigations are validating webhooks and clear status/problem reporting to the user.

Those considerations notwithstanding, I am very open to removing all status-related code out of this PR, at the very least. Setting status is actually quite complex (not least due to the problems @adutra was raising where you have to change the way that e.g. Stargate and Reaper set status whenever you add an additional status field).

Let me know which way you want to go.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+		},
+		Spec: promapi.ServiceMonitorSpec{
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app.kubernetes.io/managed-by": "cass-operator",
+					//It appears that this label is always set true by cass-operator, so I don't think filtering on it adds any value and we should look to deprecate.
+					//"cassandra.datastax.com/prom-metrics": "true",
+				},
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{Key: "cassandra.datastax.com/cluster", Operator: "In", Values: []string{cfg.ClusterName}},
```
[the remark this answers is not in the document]
Miles-Garnsey: Good idea, thanks. Done.

controllers/k8ssandra/k8ssandracluster_controller.go
jsanda: Sorry, I am referring to https://github.com/k8ssandra/k8ssandra/tree/main/charts/k8ssandra-operator.

controllers/k8ssandra/telemetry_reconciler.go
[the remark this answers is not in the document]
Miles-Garnsey: > To be consistent, this should be done in the Coalesce function in datacenter.go.

I'm not sure I understand your intention here. Coalesce merge a `*api.CassandraClusterTemplate` with a `*api.CassandraDatacenterTemplate` to produce a `DatacenterConfig`. 

I'm merging two `TelemetrySpec` objects to produce a `TelemetrySpec` object, and that's all my reconciler needs to proceed, it doesn't need to know about the entire DC, and `Coalesce` doesn't seem to do what I need it to?

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+	utils.AddHashAnnotation(&sm, k8ssandraapi.ResourceHashAnnotation)
+	return sm, nil
+}
+
+// GetCassandraPromSMName gets the name for our ServiceMonitors based on
+func GetCassandraPromSMName(cfg CassTelemetryResourcer) string {
+	return strings.Join([]string{cfg.ClusterName, cfg.DataCenterName, "prom-servicemonitor"}, "-")
+}
+
+// CreateResources executes the creation of the desired Prometheus resources on the cluster.
```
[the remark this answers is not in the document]
Miles-Garnsey: Thanks, fixed.

apis/telemetry/v1alpha1/telemetry_types.go
```diff
@@ -0,0 +1,38 @@
...
+}
+type PrometheusTelemetrySpec struct {
+	// Enable the creation of Prometheus serviceMonitors for this resource (Cassandra or Stargate).
+	Enabled *bool `json:"enabled,omitempty"` // A bool flag required here to disambiguate when e.g. the cluster should have telemetry turned on but one DC should have it explicitly turned off.
+	// CommonLabels are applied to all serviceMonitors created.
+	// +optional
+	CommonLabels map[string]string `json:"commonLabels,omitempty"`
+}
+
+type TelemetryStatus struct {
```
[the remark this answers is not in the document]
Miles-Garnsey: From my comment above 

> The reason I think status may be important is that any telemetry we integrate with runs the risk that the dependancies (e.g. Prometheus operator) are not installed. In my view, things are very likely to go wrong at this edge of the system, because we have more external dependancy and less control over the things we are integrating with.
> 
> The logical mitigations are validating webhooks and clear status/problem reporting to the user.
> 
> Those considerations notwithstanding, I am very open to removing all status-related code out of this PR, at the very least. Setting status is actually quite complex (not least due to the problems @adutra was raising where you have to change the way that e.g. Stargate and Reaper set status whenever you add an additional status field).
> 
> Let me know which way you want to go.
> 
>

controllers/k8ssandra/k8ssandracluster_controller.go
Miles-Garnsey: Oh, interesting. I wasn't aware that this even existed. Thanks for bringing it to my attention. I will raise a separate ticket for it.

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: Change made, I don't particularly like what this does to error handling (we basically need to panic if something goes wrong with the static initialisation...) but if that's preferred, it will surely work.

controllers/k8ssandra/telemetry_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+	dcCfg := telemetry.CassTelemetryResourcer{
+		CassandraNamespace: kc.Namespace,
```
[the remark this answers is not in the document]
Miles-Garnsey: That is a very good catch. Thanks!

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,459 @@
...
+	if err != nil {
+		return err
+	}
+	cfg.CassTelemetryResourcer.Logger.Info("checking whether Prometheus ServiceMonitor for Cassandra already exists")
+	// Logic to handle case where SM does not exist.
+	var actualSM promapi.ServiceMonitor
+	if err := client.Get(ctx, types.NamespacedName{Name: desiredSM.Name, Namespace: desiredSM.Namespace}, &actualSM); err != nil {
+		if errors.IsNotFound(err) {
+			cfg.CassTelemetryResourcer.Logger.Info("Prometheus ServiceMonitor for Cassandra not found, creating")
+			if err := controllerutil.SetControllerReference(owner, &desiredSM, client.Scheme()); err != nil {
```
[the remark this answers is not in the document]
Miles-Garnsey: Setting this to the CassDC as you suggested.

Miles-Garnsey:
**Changes made as requested:**

* Use `client.DeleteAllOf()` instead of a list and then delete in a for loop for `CleanupResources` in `cass_prometheus_resourcer.go`.
* Ownership of Prom SM now resides with CassDC not K8ssandraCluster. 
* Static initialisation for SM endpoints in `cass_prometheus_resourcer.go`. 
* Typos, remove label that we aren't using (may need to add back when adding stargate telemetry). 
* Move validation out of `apis/`.
* Use CassDC namespace when configuring `CassTelemetryResourcer` not k8ssandracluster namespace. 
* Remove status APIs and logic (may add back via another PR).
* Use cass-operator labels constants in prometheus_resourcer. 
* Use pointers consistently in UpdateResources. 
* Use util.MergeMap where possible. 
* Refactor labelling to use constants defined in k8ssandra api. 
* Change Telemetry API so that `Prometheus.Enabled` is not optional and is not a pointer. (When `Telemetry.Prometheus` is defined, `Enabled` must be defined.) 
* Clean up logging, no need for datacenter key in reconciler as it is already added to the logger. 
* Reduce permissions for manager SA down to namespace scoped. Remove ClusterRoleBinding. 

**Will not fix (at this stage):**
* Use Coalesce instead of merge.

**Pending:**

* Fleshing out the rest of the tests.

Miles-Garnsey:
Updates:
* All tests now added. 
* Rebased on main.
* Using result.ReconcileResult for reconciliation return types.

Ready for another round of review - and hopefully merging - once the GHA tests have run. @jsanda - anything missing here?

Miles-Garnsey:
It has become infeasible to keep this PR separate from #260. I am closing this PR and merging it into #260 to progress the work there.

PR #242, opened 2021-12-08
Add medusa APIs

### LINKED ISSUE #45: K8SSAND-745 ⁃ Migrate medusa_controller to k8ssandra-operator
MIgrate and recreate medusa related operator elements to the k8ssandra-operator, including the CRDs and controllers.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-745) by [Unito](https://www.unito.io)
┆epic: Medusa Operator Integration
┆friendlyId: K8SSAND-745
┆priority: Medium

adejanovski:
**What this PR does**:
Add medusa-operator capabilities to k8ssandra-operator

Fixes #45

**Checklist**
- [ ] Changes manually tested


apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	Image ContainerImage `json:"medusaContainerImage,omitempty"`
+
+	// SecurityContext applied to the Medusa containers.
+	// +optional
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
+
+	// Defines the username and password that Medusa will use to authenticate CQL connections to Cassandra clusters.
+	// These credentials will be automatically turned into CQL roles by cass-operator when bootstrapping the datacenter,
+	// then passed to the Medusa instances, so that it can authenticate against nodes in the datacenter using CQL.
+	// The secret must be in the same namespace as Reaper itself and must contain two keys: "username" and "password".
```
jsanda, line 130: The comment mentions Reaper. Looks like a copy/paste error from... me 😄

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+// Create all things Medusa related in the cassdc podTemplateSpec
+func AddMedusa(dcConfig *cassandra.DatacenterConfig, medusaSpec *medusaapi.MedusaClusterTemplate, logger logr.Logger) error {
+	// Medusa isn't enabled (anymore) in this cluster
+	if medusaSpec != nil {
+		logger.Info("Medusa is enabled")
+		updateMedusaInitContainer(dcConfig, medusaSpec, logger)
+		updateMedusaMainContainer(dcConfig, medusaSpec, logger)
+		updateMedusaVolumes(dcConfig, medusaSpec, logger)
+	} else {
+		logger.Info("Medusa is not enabled")
```
jsanda, line 57: Don't we want to make sure that the Medusa containers are not in the PodTemplateSpec here?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+	if !found {
+		// volume doesn't exist, we need to add it
+		dcConfig.PodTemplateSpec.Spec.Volumes = append(dcConfig.PodTemplateSpec.Spec.Volumes, *volume)
+	} else {
+		// Overwrite existing volume
+		dcConfig.PodTemplateSpec.Spec.Volumes[volumeIndex] = *volume
+	}
+}
+
+func findMedusaContainer(dcPodTemplateSpec *corev1.PodTemplateSpec, containerName string, initContainer bool) (int, bool) {
```
jsanda, line 210: I find working with the PodTemplateSpec cumbersome. This is a great function to make things a bit easier. What do you think about moving this to a common/share util package and renaming it `FindContainer`? 

I'd also be in favor of having an explicit `FindInitContainer` method as well.

It might even be good to create new type for working with the CassandraDatacenter's PodTemplateSpec. It would take the PodTemplateSpec (or possibly the whole CassandraDatacenter) as a member variable.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+			if container.Name == containerName {
+				return i, true
+			}
+		}
+	}
+
+	return -1, false
+}
+
+func findMedusaVolume(dcPodTemplateSpec *corev1.PodTemplateSpec, volumeName string) (int, bool) {
```
jsanda, line 231: Similar to my comment about the `FindMedusaContainer` function, what do you think about moving and renaming this to something like `FindVolume`?

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -0,0 +1,235 @@
...
+	require := require.New(t)
+
+	// Check containers presence
+	require.True(f.ContainerExists(dc, "medusa-restore", true), "dc1 doesn't have medusa-restore init container")
+	// require.True(f.ContainerExists(dc, "medusa", false), "dc1 doesn't have medusa main container")
+
+	// Check containers Image
+	require.True(f.GetContainer(dc, "medusa-restore", true).Image == fmt.Sprintf("docker.io/%s:latest", medusaImageRepo), "dc1 doesn't have medusa-restore init container")
+
+	// Check volume mounts
```
jsanda, line 223: Consider using the `assert` variants for the all the volume mount and env var checks so that the test can check all of them and fail at the end.

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+type Storage struct {
+	// The storage backend to use for the backups.
+	// Should be either of "local", "google_storage", "azure_blobs" or "s3".
+	StorageProvider string `json:"storageProvider,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: Sounds good. I've removed the comment listing possible values and traded it for an enum with an updated list of values.

apis/medusa/v1alpha1/medusa_types.go
adejanovski: Oops, busted 😅 

I've changed the comment to adapt it to Medusa's context.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: It's dealt with inside the `update...` functions: if they are present they get updated, if not they get created.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: I moved/renamed the functions to the cassandra package in `datacenter.go` as it looks like it fits there.

adejanovski: > It might even be good to create new type for working with the CassandraDatacenter's PodTemplateSpec. It would take the PodTemplateSpec (or possibly the whole CassandraDatacenter) as a member variable.

Didn't go that far yet 😅

controllers/k8ssandra/medusa_reconciler.go
adejanovski: done 👍

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Default is unlimited.
+	// +optional
+	MaxBackupCount int `json:"maxBackupCount,omitempty"`
+
+	// AWS Profile to use for authentication.
+	// +optional
+	ApiProfile string `json:"apiProfile,omitempty"`
+
+	// Max upload bandwidth in MB/s.
+	// Defaults to 50 MB/s.
```
adutra, line 59: Default values and valid ranges can be enforced with kubebuilder annotations, e.g.

```
	// +kubebuilder:default=50
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=100
```

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// If empty, the cluster name will be used.
+	// +optional
+	Prefix string `json:"prefix,omitempty"`
+
+	// Maximum backup age that the purge process should observe.
+	// +optional
+	MaxBackupAge int `json:"maxBackupAge,omitempty"`
+
+	// Maximum number of backups to keep (used by the purge process).
+	// Default is unlimited.
```
adutra, line 50: How do you represent "unlimited"? Is it zero, -1, ...?

More generally, think twice about whether to use pointers or not for a field. If not using pointers, the field will take the type's zero value by default. For numbers, this means that fields will be 0 by default. However if 0 is a valid value, then there is no way to distinguish "value is absent" from "value is zero". In such cases it's necessary to use pointers.

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Should be either of "local", "google_storage", "azure_blobs" or "s3".
+	StorageProvider string `json:"storageProvider,omitempty"`
+
+	// Kubernetes Secret that stores the key file for the storage provider's API.
+	// If using 'local' storage, this value is ignored.
+	// +optional
+	StorageSecretRef string `json:"storageSecretRef,omitempty"`
+
+	// The name of the bucket to use for the backups.
+	BucketName string `json:"bucketName,omitempty"`
```
adutra, line 38: Is it required? If so, you can add:

```
	// +kubebuilder:validation:Required
```

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Defaults to 1.
+	// +optional
+	ConcurrentTransfers int `json:"concurrentTransfers,omitempty"`
+
+	// File size over which cloud specific cli tools are used for transfer.
+	// Defaults to 100 MB.
+	// +optional
+	MultiPartUploadThreshold int `json:"multiPartUploadThreshold,omitempty"`
+
+	// Host to connect to for the storage backend.
```
adutra, line 74: Which host is used if none provided? Localhost?

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// +optional
+	Host string `json:"host,omitempty"`
+
+	// Region of the storage bucket.
+	// Defaults to "default".
+	// +optional
+	Region string `json:"region,omitempty"`
+
+	// Port to connect to for the storage backend.
+	Port int `json:"port,omitempty"`
```
adutra, line 84: I suppose the default is 9042?

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Defaults to 10 days.
+	// +optional
+	BackupGracePeriodInDays int `json:"backupGracePeriodInDays,omitempty"`
+}
+
+type ContainerImage struct {
+
+	// +kubebuilder:default="docker.io"
+	// +optional
+	Registry *string `json:"registry,omitempty"`
```
adutra, line 100: A pointer type here is not required imo, since an empty string is not a valid registry name. Same for tags and pull policies.

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// +optional
+	PullSecret corev1.LocalObjectReference `json:"imagePullSecret,omitempty"`
+}
+
+type MedusaClusterTemplate struct {
+	// MedusaContainerImage is the image characteristics to use for Medusa containers. Leave nil
+	// to use a default image.
+	// +optional
+	Image ContainerImage `json:"medusaContainerImage,omitempty"`
```
adutra, line 121:
```suggestion
	ContainerImage *ContainerImage `json:"containerImage,omitempty"`
```

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// SecurityContext applied to the Medusa containers.
+	// +optional
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
+
+	// Defines the username and password that Medusa will use to authenticate CQL connections to Cassandra clusters.
+	// These credentials will be automatically turned into CQL roles by cass-operator when bootstrapping the datacenter,
+	// then passed to the Medusa instances, so that it can authenticate against nodes in the datacenter using CQL.
+	// The secret must be in the same namespace as Reaper itself and must contain two keys: "username" and "password".
+	// +optional
+	CassandraUserSecretRef string `json:"cassandraUserSecretRef,omitempty"`
```
adutra, line 132: Prefer `corev1.LocalObjectReference` when referencing other objects in the same namespace.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+// Get the cassandra user secret name from the medusa spec or generate one based on the cluster name
+func cassandraUserSecretName(medusaSpec *medusaapi.MedusaClusterTemplate, clusterName string) string {
+	if medusaSpec.CassandraUserSecretRef == "" {
+		return fmt.Sprintf("%s-medusa", clusterName)
+	}
+	return medusaSpec.CassandraUserSecretRef
+}
+
+// Create all things Medusa related in the cassdc podTemplateSpec
+func AddMedusa(dcConfig *cassandra.DatacenterConfig, medusaSpec *medusaapi.MedusaClusterTemplate, logger logr.Logger) error {
```
adutra, line 49: Couldn't we move this function to the `medusa` package?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+			logger.Error(err, "Failed to reconcile Medusa CQL user secret")
+			return err
+		}
+	}
+	logger.Info("Medusa user secrets successfully reconciled")
+	return nil
+}
+
+// Get the cassandra user secret name from the medusa spec or generate one based on the cluster name
+func cassandraUserSecretName(medusaSpec *medusaapi.MedusaClusterTemplate, clusterName string) string {
```
adutra, line 41: Nit. The function name is misleading, it's about Medusa but the name says "cassandra"?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+		updateMedusaMainContainer(dcConfig, medusaSpec, logger)
+		updateMedusaVolumes(dcConfig, medusaSpec, logger)
+	} else {
+		logger.Info("Medusa is not enabled")
+	}
+
+	return nil
+}
+
+func updateMedusaInitContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec *medusaapi.MedusaClusterTemplate, logger logr.Logger) {
```
adutra, line 63: Pretty much the rest of the functions in this file could imo all go to the `medusa` package.

controllers/medusa/backup_controller_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (f *fakeMedusaClientFactory) GetRequestedBackups() map[string][]string {
+	requestedBackups := make(map[string][]string)
+	for k, v := range f.clients {
+		requestedBackups[k] = v.RequestedBackups
+	}
+	return requestedBackups
+}
+
+type fakeMedusaClient struct {
```
adutra, line 257: FYI you can also generate a mock with mockery + testify, we already have a few ones in the `mocks` package.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+// Generate a secret for Medusa or use the existing one if provided in the spec
+func (r *K8ssandraClusterReconciler) reconcileMedusaSecrets(
+	ctx context.Context,
+	kc *api.K8ssandraCluster,
+	logger logr.Logger,
+) error {
+	logger.Info("Reconciling Medusa user secrets")
+	if kc.Spec.Medusa != nil {
+		cassandraUserSecretRef := kc.Spec.Medusa.CassandraUserSecretRef
+		if cassandraUserSecretRef == "" {
```
adutra, line 24: You are testing `cassandraUserSecretRef == ""` twice: once here, once inside the `cassandraUserSecretName` function.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+
+func updateMedusaInitContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec *medusaapi.MedusaClusterTemplate, logger logr.Logger) {
+	restoreContainerIndex, found := findMedusaContainer(dcConfig.PodTemplateSpec, "medusa-restore", true)
+	restoreContainer := &corev1.Container{Name: "medusa-restore"}
+	if found {
+		logger.Info("Found medusa-restore init container")
+		// medusa-restore init container already exists, we may need to update it
+		restoreContainer = dcConfig.PodTemplateSpec.Spec.InitContainers[restoreContainerIndex].DeepCopy()
+	}
+	restoreContainer.Image = fmt.Sprintf("%s/%s:%s", *medusaSpec.Image.Registry, medusaSpec.Image.Repository, *medusaSpec.Image.Tag)
```
adutra, line 71: 1. See my PR #243 : I think this `Sprintf` call is broken: the generated string contains a registry and a repository, but no image name, e.g. `docker.io/thelastpickle:latest`.
2. How are you handling default values? E.g. if the registry and the repository are both nils you'd get a nil pointer dereference here. And if they are set to their zero values (the empty string), you'd get something like `"/:"`. You probably need a function to generate the image string.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,239 @@
...
+	restoreContainer.ImagePullPolicy = *medusaSpec.Image.PullPolicy
```
adutra, line 72: In #243 I'm leaning towards the idea of defaulting the pull policy to "Always" if the tag is "latest", and to "IfNotPresent" otherwise. This is what Kubernetes does. For that you would need a function here to compute the appropriate pull policy.

go.mod
```diff
@@ -10,16 +10,18 @@ require (
...
 	github.com/rs/zerolog v1.20.0
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
+	google.golang.org/grpc v1.38.0
+	google.golang.org/protobuf v1.26.0
 	gopkg.in/resty.v1 v1.12.0
 	k8s.io/api v0.22.2
 	k8s.io/apimachinery v0.22.2
-	k8s.io/client-go v0.22.2
+	k8s.io/client-go v12.0.0+incompatible
```
adutra, line 24: This change looks suspicious.

apis/medusa/v1alpha1/medusa_types.go
adejanovski: It's zero, I've set a default using the kubebuilder annotation.

apis/medusa/v1alpha1/medusa_types.go
adejanovski: I've added them now 👍

apis/medusa/v1alpha1/medusa_types.go
adejanovski: The host is only used for self hosted storage backends. S3, GCS or AKS don't require it, it's built in their clients.

apis/medusa/v1alpha1/medusa_types.go
adejanovski: No, it's not a Cassandra port. It's the port to connect to the storage backend.
It's actually optional, so I'll add the appropriate kubebuilder annotation.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+		configMapKey := client.ObjectKey{
+			Namespace: kc.Namespace,
+			Name:      fmt.Sprintf("%s-medusa", kc.Spec.Cassandra.Cluster),
+		}
+
+		if err := remoteClient.Get(ctx, configMapKey, actualConfigMap); err != nil {
+			if errors.IsNotFound(err) {
+				if err := remoteClient.Create(ctx, desiredConfigMap); err != nil {
+					logger.Error(err, "Failed to create Medusa ConfigMap")
+					return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
jsanda, line 103: You don't need to set `RequeueAfter` when the error is not nil. The request will requeued with an exponential backoff by default that is configurable via controller-runtime.

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Provides all storage backend related properties for backups.
+	StorageProperties Storage `json:"storageProperties,omitempty"`
+}
+
+// MedusaSpec defines the desired state of Medusa.
+type MedusaSpec struct {
+	MedusaClusterTemplate `json:",inline"`
+}
+
+type Medusa struct {
```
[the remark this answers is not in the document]
adejanovski: Gotcha, I'm removing the `Medusa` struct.

I've taken a look at how the cluster/datacenter level thing was done for Reaper and Stargate, and I'm not sure it applies here. Aside from disabling Medusa in select DCs, I don't think we'd want folks to set different storage properties, or medusa images, for different DCs in the same cluster 🤔  It would be dangerous and support no use case that I can think of.
Maybe I could just provide a way to disable Medusa per datacenter?

controllers/k8ssandra/medusa_reconciler.go
adejanovski: 👍

controllers/k8ssandra/medusa_reconciler.go
adejanovski: Using the new stuff you created to deal with images now that I've rebased 👍

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+func (r *K8ssandraClusterReconciler) reconcileMedusaConfigMap(
+	ctx context.Context,
+	remoteClient client.Client,
+	kc *api.K8ssandraCluster,
+	logger logr.Logger,
+) (ctrl.Result, error) {
+	logger.Info("Reconciling Medusa configMap on namespace : " + kc.Namespace)
+	if kc.Spec.Medusa != nil {
+		medusaIni := createMedusaIni(kc)
+		desiredConfigMap := createMedusaConfigMap(kc.Spec.Medusa, kc.Namespace, kc.Spec.Cassandra.Cluster, medusaIni)
```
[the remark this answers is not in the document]
adejanovski: I think we should not. The configMap, just as the cassdc objects, shouldn't be modified directly otherwise it'll drift from the k8ssandracluster object definition.
Any change to the configmap should then come from modifications to the k8c object that would be reconciled.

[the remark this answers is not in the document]
adejanovski: oooh, gotcha. I'll change this.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: This was changed since the last rebase as reconcile result handling was wrapped the ReconcileResult type.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+			resourceVersion := actualConfigMap.GetResourceVersion()
+			desiredConfigMap.DeepCopyInto(actualConfigMap)
+			actualConfigMap.SetResourceVersion(resourceVersion)
+			logger.Info("Updating configMap on namespace " + actualConfigMap.ObjectMeta.Namespace)
+			logger.Info("Should be on namespace " + actualConfigMap.ObjectMeta.Namespace)
+			if err := remoteClient.Update(ctx, actualConfigMap); err != nil {
+				logger.Error(err, "Failed to update Medusa ConfigMap resource")
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
```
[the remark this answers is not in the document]
adejanovski: Maybe we should (and I don't know if that's possible) make it so that Medusa detects changes to this file and reloads the config without requiring another grpc call?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+	return ctrl.Result{}, nil
+}
+
+func createMedusaIni(kc *api.K8ssandraCluster) string {
+	prefix := kc.Spec.Medusa.StorageProperties.Prefix
+	if prefix == "" {
+		prefix = kc.ClusterName
+	}
+
+	medusaIni := fmt.Sprintf(`
```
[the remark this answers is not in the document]
adejanovski: done 👍

controllers/k8ssandra/medusa_reconciler_test.go
adejanovski: done 👍

apis/medusa/v1alpha1/medusa_types.go
adejanovski: I'm using your container image structure and code now. Should be good.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: done 👍

controllers/medusa/backup_controller_test.go
adejanovski: gotcha, didn't do it due to lack of time.

adutra (approved):
Outstanding work 👍

Issue #63, opened 2021-07-30
Operator configuration


burmanm:
**What is missing?**
I'm opening this for discussion, but I'm not a fan of adding env variables for configuration. They're not typed and not easily validated  in the deployment configuration. Nothing guarantees that the support for such env is still in place or that the env was correctly typed.

In the cass-operator I made some work to move on to kubebuilder config and I would personally chose that route rather than adding new env variables or command line parameters.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-779) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-779
┆Priority: Medium

jsanda:
For reference, https://book.kubebuilder.io/component-config-tutorial/tutorial.html. I agree that this is what we should do. I can work on this.

jsanda:
@burmanm I read through the tutorial and I also looked at [example_test.go](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/config/example_test.go). It looks like the configuration has to come from a file and can then be parsed into a custom type. This is great for providing validations but I do not see this being useful for #49. If the config file is included in the image we would have to build a different image. If the config file is external to the image, would we map the custom type to a ConfigMap that we would mount? Using an environment variable seems much simpler.

jsanda:
@jdonenine let me summarize what the expectation is for this issue. Currently some operator configuration, e.g., control plane vs data plane, is configured via an env var. @burmanm has recommended using the kubebuilder approach which is describe at https://book.kubebuilder.io/component-config-tutorial/tutorial.html. I agree that we should do this. He has already done this for cass-operator. Some of this is already generated as part of the project scaffolding. The tutorial describes the changes necessary in the operator code.

In addition to configuring the operator for control plane vs data plane, this is where we could apply some global, default settings. For example, we could add a flag to enable host networking by default. We would want this when running in kind. It would eliminate having to set it for every K8ssandraCluster. We could also add properties for requeue timeouts which we lower in integration tests in order to speed up the tests. As a last example, we could configure registry overrides here.

@burmanm if you haven't started on this and have other stuff on your plate this could be a good ticket for someone else to pick up as a way to get ramped up.

PR #243, opened 2021-12-08
Make Reaper images configurable

### LINKED ISSUE #24: K8SSAND-631 ⁃ Make the Reaper image configurable
This an extension of #23 but focuses on the Reaper CRD. I am breaking this out into a separate ticket because the work can happen separately and possibly in parallel.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-631) by [Unito](https://www.unito.io)
┆epic: Reaper Controller Integration
┆fixVersions: k8ssandra-operator-v1.0.0-alpha.3
┆friendlyId: K8SSAND-631
┆priority: Medium

adutra:
**What this PR does**:

Makes Reaper images configurable. It also enforces the same struct for both Reaper and Stargate images.

Fixes #24 

**Checklist**
- [x] Changes manually tested


adutra:
@adejanovski this PR might be of interest for your work on the Medusa front. At some point we'll need to homogenize how images are specified across different API groups.

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -81,6 +82,34 @@ type ReaperDatacenterTemplate struct {
 	InitContainerSecurityContext *corev1.SecurityContext `json:"initContainerSecurityContext,omitempty"`
 }
 
+type ContainerImage struct {
```
adutra, line 85: I'm not creating a common type for both Reaper and Stargate because even if the struct is similar, the kubebuilder tags are not. If anyone has a better idea let me know.

pkg/stargate/deployments.go
```diff
@@ -186,42 +202,60 @@ func computeSeedServiceUrl(dc *cassdcapi.CassandraDatacenter) string {
...
+			name = DefaultStargate4ImageName
 		}
 		tag := "latest"
-		if containerImage.Tag != nil {
-			tag = *containerImage.Tag
+		if containerImage.Tag != "" {
+			tag = containerImage.Tag
 		}
-		return fmt.Sprintf("%v/%v:%v", registry, containerImage.Repository, tag)
+		return fmt.Sprintf("%v/%v/%v:%v", registry, repository, name, tag)
```
adutra, line 242: I believe this was wrong since the beginning, it was concatenating the registry and the repository without an image name.

apis/reaper/v1alpha1/reaper_types.go
jsanda: Did you look at how images are handled now in cass-operator? Defaults are supplied via a ConfigMap. One nice thing about this is that it doesn't require any code changes if the user wants to change those defaults. We would want to do #63 for the ConfigMap approach (if we were to be consistent with the way it's done in cass-operator).

adutra: @jsanda I did, and frankly it looks overly complex. What exactly would you like to see implemented in k8ssandra-operator? I see a few problems trying to port cass-operator's image handling code here:

- Most types are too specific to cass-operator, e.g. [this one](https://github.com/k8ssandra/cass-operator/blob/65e47b5ac16adfccfbb335abf9723263554fc3ee/apis/config/v1beta1/imageconfig_types.go#L46-L56).
- I don't see ConfigMaps being used, however I do see a field named `ImageConfigFile` but I am struggling to understand how it works since there is no test at all for it.
- Indeed we would need to do #63 before going further in this direction. Should I close this PR for now and come back to it later?

jsanda: The ConfigMap gets generated. See [here](https://github.com/k8ssandra/cass-operator/blob/master/config/manager/kustomization.yaml#L7). We can revisit using a ImageConfig/ConfigMap when we tackle #63. I will do another review of this today so we can get it merged.

jsanda: I wonder if the `+kubebuilder:default` annotation could help avoid duplicating `ContainerImage`. It says that it accepts an object as one of the types for a value. Might be worth exploring.

pkg/reaper/deployment.go
```diff
@@ -193,6 +201,55 @@ func computeProbe(probeTemplate *corev1.Probe) *corev1.Probe {
 	return probe
 }
 
+func computeImage(containerImage *api.ContainerImage) string {
```
jsanda, line 204: Even if we wind up having separate `ContainerImage` types for Reaper and for Stargate there is a lot of duplication with the `computeImage`, `computeImagePolicy`, and `computeImagePullSecrets` functions. I believe the latter two are identical. Could we introduce some methods and interfaces to eliminate some or most of the duplication? Each of the `ContainerImage` structs could implement some interface(s), and the functions could be refactored to take the interface types as arguments.

adutra: > Could we introduce some methods and interfaces to eliminate some or most of the duplication?

I gave this a try in fa9cf7d, let me know what you think.

pkg/images/images.go
```diff
@@ -0,0 +1,152 @@
...
+	return &image{
+		ImageId:       id,
+		pullPolicy:    pullPolicy,
+		pullSecretRef: pullSecretRef,
+	}
+}
+
+// ImageString returns the string that uniquely identifies the given ImageId. It panics if id is nil. It does not check
+// if the id components are valid.
+func ImageString(i ImageId) string {
```
adutra, line 85: `Image` could inherit from `fmt.Stringer` but that would force all structs satisfying `Image` to also implement `String()` – since there are no default methods in Go.

jsanda (approved):
Thanks for `Image` and `ImageId` changes!

apis/reaper/v1alpha1/reaper_types.go
adutra: As agreed offline, I'm going to merge now and investigate this idea in another PR.

PR #233, opened 2021-12-05
start using result.ReconcileResult

jsanda:
Temporarily copy ReconcileResult interface into project until it is publicly
available in cass-operator.

**What this PR does**:
Introduces the usage of ReconcileResult in K8ssandraClusterReconciler. Using ReconcileResult allows us to more easily refactor the Reconciler into smaller methods.

Other Reconcilers, ReaperReconciler, StargateReconciler, and SecretSyncController, can be updated to use ReconcileResult as necessary in subsequent PRs. 

**Which issue(s) this PR fixes**:
Fixes #232

**Checklist**
- [ ] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CHANGELOG.md updated (not required for documentation PRs)
- [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

Issue #21, opened 2021-06-23
K8SSAND-615 ⁃ Add support for adding a datacenter to an existing cluster


jsanda:
## Background
The steps for adding a DC to an existing cluster differ from those for setting up a new, multi-DC cluster. It involves updating replication strategies of keyspaces and streaming data from an existing DC to the new DC. See the following for more background on the steps involved for adding a DC to an existing cluster:

* https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/operations/opsAddDCToCluster.html
* https://thelastpickle.com/blog/2019/02/26/data-center-switch.html

## Purpose
The purpose of this ticket is to provide support for adding a DC to an existing cluster where the DC will live in a different k8s cluster. 

We will also support adding a DC within an existing k8s cluster, but #95 needs to be done first. Adding a DC within an existing k8s cluster will therefore be addressed in a separate ticket.

## Details
Assume _DC_  refers to a CassandraDatacenter unless noted otherwise.

* Need to distinguish between adding a DC to a cluster that is initializing vs one that is fully initialized
  * If a DC is added to the K8ssandraCluster before all DCs are initialized then the process is the same as setting up a new, multi-DC cluster
  * If a DC is added to the K8ssandraCluster after all DCs are initialized then we follow the process for adding a DC to an existing cluster
  * We can determine if the cluster is initialized by
    * Add a `CassandraInitialized` status condition to K8ssandraCluster
    * Set `CassandraInitialized` to `true` after all DCs have their `Initialized condition set to `true` 
* All user-created keyspaces as well as system_traces, system_distributed, and system_auth should be configured to use NetworkTopologyStrategy before adding a new DC
  * Should we add a validation check for user-created keyspaces in an effort to prevent misconfigured clusters?
* Deploy the new DC
  * Configure additional seeds as we normally do so that the new DC has seeds from other, existing DCs
* Wait for the DC to reach the Ready state
* Update replication strategy for all user-created keyspaces as well as system_traced, system_distributed, and system_auth to include replicas for the new DC
  * Need to provide a way for user to specify replicas for each keyspace and each DC
* Run rebuild on each node in new DC
  * Provide a way for user to specify src DC for streaming
  * Rebuild can run on more than one node at a time. This will be left as an optimization.

#### Should we add a validation check for user-created keyspaces in an effort to prevent misconfigured clusters?
[...]

PR #244, opened 2021-12-08
Check and update system replication

### LINKED ISSUE #240: K8SSAND-1090 ⁃ Scaling the cluster causes a rolling restart
**What happened?**
When you scale a Cassandra cluster, either by adding nodes to a DC, removing nodes from a DC, or by adding a DC, it will trigger an unnecessary rolling restart for existing DCs. 

Replication of system keyspaces is configured by setting the following Java system properties in the `server-config-init` container:

* `cassandra.system_distributed_replication_dc_names`
* `cassandra.system_distributed_replication_per_dc`

They are specified in an env var that looks something like this:

```yaml
    - name: CONFIG_FILE_DATA
      value: '{"cluster-info":{"name":"test","seeds":"test-seed-service,test-dc1-additional-seed-service"},"datacenter-info":{"graph-
enabled":0,"name":"dc1","solr-enabled":0,"spark-enabled":0},"jvm-options":{"additional-jvm-opts":["-
Dcassandra.system_distributed_replication_dc_names=dc1","-Dcassandra.system_distributed_replication_per_dc=1"]}}'
```

This would be for a single node cluster. If we scale up our cluster to 2 nodes, k8ssandra-operator sets `cassandra.system_distributed_replication_per_dc=2`. This will cause a StatefulSet update, i.e., rolling restart, since it updates the underlying PodTemplateSpec.

What's worse is that if the scaling is within an existing DC, the replication doesn't get updated. The Java system properties are basically idempotent. Only their initial values are applied.

The system keyspaces' replication should be configured as follows:

`replication = min(num_nodes, 3)`

Changing the replication should not trigger a rolling restart.

**Environment**

* K8ssandra Operator version:
1.0.0-alpha.2
* Manifests:

```yaml
kind: K8ssandraCluster
metadata:
  name: test
spec:
  cassandra:
    cluster: test
    serverVersion: "4.0.0"
    storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
    config:
      cassandraYaml:
        auto_snapshot: false
        memtable_flush_writers: 1
        commitlog_segment_size_in_mb: 2
        concurrent_compactors: 1
        compaction_throughput_mb_per_sec: 0
        sstable_preemptive_open_interval_in_mb: 0
        key_cache_size_in_mb: 0
        thrift_prepared_statements_cache_size_mb: 1
        prepared_statements_cache_size_mb: 1
        start_rpc: false
        slow_query_log_timeout_in_ms: 0
        counter_cache_size_in_mb: 0
        concurrent_reads: 2
        concurrent_writes: 2
        concurrent_counter_writes: 2
      jvmOptions:
        heapSize: 512Mi
    networking:
      hostNetwork: true
    resources:
      limits:
        memory: 1024Mi
    datacenters:
      - metadata:
          name: dc1
        size: 1
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1090) by [Unito](https://www.unito.io)
┆epic: Addition of new DataCenters
┆fixVersions: k8ssandra-operator-v1.0.0-alpha.3
┆friendlyId: K8SSAND-1090
┆priority: Medium

jsanda:
**What this PR does**:
* Store initial system replication in annotation on K8ssandraCluster
    * Done to help avoid rolling restart after a scaling event
* Check and update replication (if necessary) after each CassandraDatacenter becomes ready
    * Done in support of #21   

Fixes #240

**Note:** I branches off of #233 for this PR, so it should be merged first, and then I will rebase as necessary.

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1179,6 +1179,41 @@ func verifySuperUserSecretCreated(ctx context.Context, t *testing.T, f *framewor
...
+		actualReplication := &cassandra.SystemReplication{}
+		if err := json.Unmarshal([]byte(val), actualReplication); err != nil {
+			t.Logf("Failed to unmarshal system replication annotation: %v", err)
+			return false
+		}
+
+		return reflect.DeepEqual(expectedReplication, *actualReplication)
+	}, timeout, interval, "Failed to verify that the system replication annotation was set correctly")
+
+	kc = &api.K8ssandraCluster{}
```
adutra, line 1209: Why retrieve the CRD again here?

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -197,239 +114,61 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
-			return ctrl.Result{RequeueAfter: r.ReconcilerConfig.LongDelay}, err
-		}
-	}
+	kcLogger.Info("Finished reconciling the k8ssandracluster")
 
-	// Reconcile Stargate and Reaper across all datacenters
+	return result.Done().Output()
+}
+
+func (r *K8ssandraClusterReconciler) reconcileStargateAndReaper(ctx context.Context, kc *api.K8ssandraCluster, dcs []*cassdcapi.CassandraDatacenter, logger logr.Logger) result.ReconcileResult {
```
adutra, line 156: Maybe rename the method to something like `afterDatacentersFullyReconciled` since we are likely going to add more stuff here (Medusa, for instance).

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -602,21 +324,358 @@ func (r *K8ssandraClusterReconciler) setStatusForStargate(kc *api.K8ssandraClust
...
-func (r *K8ssandraClusterReconciler) reconcileStargateAuthSchema(
-	ctx context.Context,
-	kc *api.K8ssandraCluster,
-	dc *cassdcapi.CassandraDatacenter,
-	remoteClient client.Client,
-	logger logr.Logger,
-) error {
-	managementApi, err := r.ManagementApi.NewManagementApiFacade(ctx, dc, remoteClient, logger)
-	if err == nil {
+func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) result.ReconcileResult {
```
adutra, line 327: How about moving this method + `checkFinalizer` to a separate file, to keep this one short and concise? 

(We would need to do the same for Stargate-related methods and methods related to secrets.)

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -602,21 +324,358 @@ func (r *K8ssandraClusterReconciler) setStatusForStargate(kc *api.K8ssandraClust
...
+
+func (r *K8ssandraClusterReconciler) reconcileReplicatedSecret(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) result.ReconcileResult {
+	if err := secret.ReconcileReplicatedSecret(ctx, r.Client, r.Scheme, kc, logger); err != nil {
+		logger.Error(err, "Failed to reconcile ReplicatedSecret")
+		return result.Error(err)
+	}
+	return result.Continue()
+}
+
+func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) (result.ReconcileResult, []*cassdcapi.CassandraDatacenter) {
```
adutra, line 453: And how about moving this one too, maybe to a `datacenters.go` file?

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -602,21 +324,358 @@ func (r *K8ssandraClusterReconciler) setStatusForStargate(kc *api.K8ssandraClust
...
+
+	return result.Continue(), actualDcs
+}
+
+// checkSystemReplication checks for the SystemReplicationAnnotation on kc. If found, the
+// JSON value is unmarshalled and returned. If not found, the SystemReplication is computed
+// and is stored in the SystemReplicationAnnotation on kc. The value is JSON-encoded.
+// Lastly, kc is patched so that the changes are persisted,
+func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) (*cassandra.SystemReplication, error) {
+	if val, found := kc.Annotations[api.SystemReplicationAnnotation]; found {
```
adutra, line 591: Periodic reminder to use `utils.GetAnnotation` :-)

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -602,21 +324,358 @@ func (r *K8ssandraClusterReconciler) setStatusForStargate(kc *api.K8ssandraClust
...
+	}
+
+	return result.Continue(), actualDcs
+}
+
+// checkSystemReplication checks for the SystemReplicationAnnotation on kc. If found, the
+// JSON value is unmarshalled and returned. If not found, the SystemReplication is computed
+// and is stored in the SystemReplicationAnnotation on kc. The value is JSON-encoded.
+// Lastly, kc is patched so that the changes are persisted,
+func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) (*cassandra.SystemReplication, error) {
```
adutra, line 590: And again, I'd move this method and the next one to a `replication.go` file, wdyt?

adutra (changes requested):
@jdonenine and @jsanda I did a review round as requested, but it seems this PR needs some rebasing.

controllers/k8ssandra/k8ssandracluster_controller.go
jsanda: I'm good with renaming it, but Medusa stuff won't go there. The main parts for Medusa are deployed in the Cassandra pod.

jsanda: The more I think about it I am not sure about the change. If I make that name change then `reconcileStargateAuthSchema` and `reconcileReaperSchema` should also be called in that method. Maybe that's fine. I suppose it would make it clear about where to put new method/function calls for later changes. @adutra wdyt?

adutra: Indeed, schema reconciliation would have to happen inside this method. Since Medusa won't be involved here, I'd say let's leave the name as it is for now.

adutra (approved):

adutra (approved):

PR #257, opened 2021-12-15
Extract common Image struct

adutra:
**What this PR does**:

This PR increases the amount of shared code between Stargate and Reaper regarding image handling. It also moves some constants to the `utils` package to avoid circular references.

None.

**Checklist**
- [ ] Changes manually tested


adutra:
Note that I'm not removing `utils.IsNil`, although it is not used anymore.

controllers/k8ssandra/datacenters.go
```diff
@@ -86,7 +86,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 				return result.Error(err), actualDcs
 			}
 
-			if !utils.CompareAnnotations(actualDc, desiredDc, api.ResourceHashAnnotation) {
+			if !utils.CompareHashes(actualDc, desiredDc) {
```
jsanda, line 89: Since we have `AddHashAnnotation` what do you think about `CompareHashAnnotations` for a bit more consistency and clarity?

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -32,14 +33,16 @@ const (
...
-	ContainerImage *ContainerImage `json:"containerImage,omitempty"`
+	// +kubebuilder:default={repository:"thelastpickle",name:"cassandra-reaper",tag:"3.1.0"}
+	ContainerImage *images.Image `json:"containerImage,omitempty"`
 
 	// The image to use for tje Reaper pod init container (that performs schema migrations).
-	// The default is "thelastpickle/cassandra-reaper:3.0.0".
+	// The default is "thelastpickle/cassandra-reaper:3.1.0".
 	// +optional
-	InitContainerImage *ContainerImage `json:"initContainerImage,omitempty"`
+	// +kubebuilder:default={repository:"thelastpickle",name:"cassandra-reaper",tag:"3.1.0"}
```
jsanda, line 44: 👍

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -170,8 +170,8 @@ func (r *K8ssandraClusterReconciler) SetupWithManager(mgr ctrl.Manager, clusters
 	clusterLabelFilter := func(mapObj client.Object) []reconcile.Request {
 		requests := make([]reconcile.Request, 0)
 
-		kcName := utils.GetLabel(mapObj, api.K8ssandraClusterNameLabel)
-		kcNamespace := utils.GetLabel(mapObj, api.K8ssandraClusterNamespaceLabel)
+		kcName := utils.GetLabel(mapObj, utils.K8ssandraClusterNameLabel)
```
jsanda, line 173: I realize that this is subjective, but I am not fan of moving the labels and annotations constants into the `utils` package. It seems fairly common for them to be defined with the api types. cass-operator is of course an example of this. https://github.com/kubernetes-sigs/cluster-api is a really good example because it is a large project with lots of type, and they are documented really well in the source code. The api types provide documentation as do label and annotation constant. It seem less obvious to me (again my opinion) to look in a utils package common, well-known labels and annotations. If we want to move them out of the apis directory, I would prefer a dedicated package, maybe `constants`.

adutra: The problem is that if `utils` references `api`, then `api` cannot reference `utils`, not even transitively.

Currently we have: `api` -> `images` -> `utils`.

Looking at cluster-api, I see that they avoid the circular reference by placing their labels and annotation utilities in separate packages. I will give this idea a try.

pkg/secret/replicated.go
```diff
@@ -85,9 +86,9 @@ func ReconcileSecret(ctx context.Context, c client.Client, secretName string, kc
 	}
 
 	// It exists or was created: ensure it has proper annotations
-	if !utils.IsManagedBy(currentSec, kcKey) {
-		utils.SetManagedBy(currentSec, kcKey)
-		utils.AddAnnotation(currentSec, OrphanResourceAnnotation, "true")
+	if !labels.IsManagedBy(currentSec, kcKey) {
+		labels.SetManagedBy(currentSec, kcKey)
+		annotations.AddAnnotation(currentSec, OrphanResourceAnnotation, "true")
```
adutra, line 91: @jsanda should we move this annotation to an api package as well?

jsanda (approved):

PR #260, opened 2021-12-15
Cassandra DC, Stargate prometheus integration [K8SSAND-1083,K8SSAND-1085]

Miles-Garnsey:
**What this PR does**:

Ensures that , when the Stargate resource's `Spec.Telemetry.Prometheus.Enabled` field is set `true`, a Prometheus ServiceMonitor is deployed alongside the Stargate resource.

Adds an integration to install a Prometheus operator ServiceMonitor when a CassandraDatacenter is deployed, to monitor that DC.

#236, #234

**What we are not doing**

Providing protection in the case where a user has requested Prometheus telemetry but Prometheus CRDs are not installed in the cluster.
Adding hooks into status to report whether telemetry has successfully come up.
Adding RBAC rules to the k8ssandra helm charts to cater to the ServiceMonitor resources which we are creating.

**Notes**

There is something quirky in the coalesce functions for Stargate resources, [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/3460350953823efbb763f985aa1edab3af2ebd07/apis/stargate/v1alpha1/stargate_types.go#L166). I have elected not to implement any sort of merging between DC templates residing in the cluster definition and DCs defined via the datacenters array, pending any changes to the coalesce function for Stargate specs.

(@adutra if you have any comments on the above I'd be open to them, I think the comments there are yours?)

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
Notes on this PR:

* There's still an outstanding piece of work to add tests in `pkg/telemetry` but they can probably be largely patterned after those used for the Cassandra telemetry.
* I'm requesting review early because #241 has been open for some time and needs to be merged before this PR can be.
* Because this code is branched off `Miles-Garnsey:cassdc-prometheus-integration` we'll need to rebase this branch from master before merging, only once `Miles-Garnsey:cassdc-prometheus-integration` is merged.

Miles-Garnsey:
Unit tests all added and passing. Starting manual testing and will also rebase.

Miles-Garnsey:
I have had to merge #241 into this PR. I've updated the description to reflect that it now addresses two tickets. That makes it huge, but there was too much shared code between the two PRs, and as #241 has progressed, it has become impossible to keep them separate.

Miles-Garnsey:
I've now manually tested and both Stargate and Cassandra's ServiceMonitors are created correctly. 

@jsanda has suggested not bothering to check if Prometheus is installed before adding the scheme to the client (as it doesn't matter if the scheme is added on client side, even if the prom. operator is not installed). I will make that change tomorrow.

Miles-Garnsey:
Updates:
* Rebased (a lot of changes to account for there...)
* Updated `main.go` so that prom API will be added to scheme irrespective of whether prom installed. (Per John's suggestion).
* Manually tested, all good for both cassDC and Stargate telemetry.

I'm now seeing a failure in one of the envtests, trying to track down where that's coming from. It seems that Stargate is not coming up in the second DC in multi-cluster tests.

```
k8ssandracluster_controller_test.go:1067: 
            	Error Trace:	k8ssandracluster_controller_test.go:1067
            	            				testenv.go:218
            	Error:      	Condition never satisfied
            	Test:       	TestK8ssandraCluster/CreateMultiDcClusterWithStargate
```

Miles-Garnsey:
I got the tests passing locally, suggesting that this is ready to go. Turns out they started failing when I switched to using `result.Done()` as a return type in one of the rebases. It should have been `result.Continue()`.

@jsanda if you could take a look at this prior to merging any other big PRs I'd be most grateful - I've already rebased this three times, the most recent one introduced a lot of complications :/

pkg/telemetry/stargate_prometheus_resourcer.go
```diff
@@ -0,0 +1,252 @@
...
+		os.Exit(1)
+	}
+}
+
+// mustLabels() returns the set of labels essential to managing the Prometheus resources. These should not be overwritten by the user.
+func (cfg StargatePrometheusResourcer) mustLabels() map[string]string {
+	return map[string]string{
+		k8ssandraapi.ManagedByLabel:  k8ssandraapi.NameLabelValue,
+		k8ssandraapi.PartOfLabel:     k8ssandraapi.PartOfLabelValue,
+		k8ssandraapi.DatacenterLabel: cfg.StargateName,
```
jsanda, line 145: Not sure `StargateName` makes sense as the value for `DatacenterLabel`.

pkg/telemetry/stargate_prometheus_resourcer.go
```diff
@@ -0,0 +1,252 @@
...
+			},
+			Endpoints: stargateEndpointHolder.Spec.Endpoints,
+		},
+	}
+	annotations.AddHashAnnotation(&sm)
+	return &sm, nil
+}
+
+// UpdateResources executes the creation of the desired Prometheus resources on the cluster.
+func (cfg StargatePrometheusResourcer) UpdateResources(ctx context.Context, client runtimeclient.Client, owner *stargateapi.Stargate) error {
```
jsanda, line 189: This function is virtually the same as `CassPrometheusResourcer.UpdateResources`. With a bit of refactoring a lot of duplicate code can be eliminated.

pkg/telemetry/stargate_prometheus_resourcer.go
```diff
@@ -0,0 +1,252 @@
...
+			return nil
+		}
+	}
+	return nil
+}
+
+// CleanupResources executes the cleanup of any resources on the cluster, once they are no longer required.
+// CleanupResources does not rely on user-defined CommonLabels to select which resources to delete. It uses only
+// k8ssandra-operator defined labels.
+func (cfg StargatePrometheusResourcer) CleanupResources(ctx context.Context, client runtimeclient.Client) error {
```
jsanda, line 241: This function is virtually the same as `CassPrometheusResourcer.CleanupResources`. With a bit of refactoring a lot of duplicate code can be eliminated.

pkg/telemetry/cass_prometheus_resourcer.go
```diff
@@ -0,0 +1,453 @@
...
+// NewServiceMonitor returns a Prometheus operator ServiceMonitor resource.
+func (cfg CassPrometheusResourcer) NewServiceMonitor() (*promapi.ServiceMonitor, error) {
+	// validate the object we're being passed.
+	if cfg.CassandraNamespace == "" || cfg.ServiceMonitorName == "" || cfg.DataCenterName == "" || cfg.ClusterName == "" {
+		return nil, TelemetryConfigIncomplete{}
+	}
+	// Overwrite any CommonLabels the user has asked for if they conflict with the labels essential for the functioning of the operator.
+	mergedLabels := utils.MergeMap(cfg.CommonLabels, cfg.mustLabels())
+	sm := promapi.ServiceMonitor{
+		TypeMeta: metav1.TypeMeta{
```
jsanda, line 354: You don't need to specify the TypeMeta.

pkg/telemetry/cass_telemetry_resourcer.go
```diff
@@ -0,0 +1,65 @@
...
+	if err := promResourcer.CleanupResources(ctx, client); err != nil {
+		cfg.Logger.Error(err, "error cleaning up telemetry resources")
+		return err
+	}
+	return nil
+}
+
+// GetCassandraPromSMName gets the name for our ServiceMonitors based on cluster and DC name.
+func GetCassandraPromSMName(cfg CassTelemetryResourcer) string {
+	return strings.Join([]string{cfg.ClusterName, cfg.DataCenterName, "cass-servicemonitor"}, "-")
```
jsanda, line 64: Wouldn't it be better to just use `fmt.Sprintf` or concatenation via `+` vs creating a slice?

controllers/k8ssandra/cassandra_telemetry_reconciler_test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// new_DummyK8ssandraClusterReconciler gives us back a `K8ssandraClusterReconciler` with just the fields that we need to test `reconcileCassandraDCTelemetry()`.
+func newDummyK8ssandraClusterReconciler() K8ssandraClusterReconciler {
+	return K8ssandraClusterReconciler{ReconcilerConfig: &config.ReconcilerConfig{DefaultDelay: interval}}
+}
+
+// Test_CassPrometheusResourcer_UpdateResources_TracksNamespaces tests that the servicemonitor is created in the namespace of the CassandraDC,
+// not the namespace of the k8ssandraCluster.
+func Test_reconcileCassandraDCTelemetry_TracksNamespaces(t *testing.T) {
```
jsanda, line 29: We are relying on `envtest` for controller integration tests. We need a new or updated integration test for the K8ssandraCluster controller (see `k8ssandracluster_controller_test.go`) to verify the ServiceMonitor functionality. 

It would be good to have a happy path test which will of course require having the ServiceMonitor CRD installed. We configure CRDs in `testenv.go`. Assuming you already have unit test coverage for creating the ServiceMonitor, it would be probably be sufficient to check that the object was actually created.

It would be nice to also verify that the ServiceMonitor is deleted when telemetry is disabled.

controllers/stargate/stargate_telemetry_reconciler_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+// new_DummyK8ssandraClusterReconciler gives us back a `K8ssandraClusterReconciler` with just the fields that we need to test `reconcileCassandraDCTelemetry()`.
+func newDummyK8ssandraClusterReconciler() StargateReconciler {
+	return StargateReconciler{ReconcilerConfig: &config.ReconcilerConfig{DefaultDelay: interval}}
+}
+
+// Test_reconcileStargateTelemetry_TracksNamespaces tests that the servicemonitor is created
+func Test_reconcileStargateTelemetry_succeeds(t *testing.T) {
```
jsanda, line 26: We are relying on `envtest` for controller integration tests. We need a new or updated integration test for the Stargate controller (see `stargate_controller_test.go`) to verify the ServiceMonitor functionality. 

It would be good to have a happy path test which will of course require having the ServiceMonitor CRD installed. We configure CRDs in `testenv.go`. Assuming you already have unit test coverage for creating the ServiceMonitor, it would be probably be sufficient to check that the object was actually created.

It would be nice to also verify that the ServiceMonitor is deleted when telemetry is disabled.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
+	ctx context.Context,
+	kc *api.K8ssandraCluster,
+	dcTemplate api.CassandraDatacenterTemplate,
+	actualDc *cassdcapi.CassandraDatacenter,
+	logger logr.Logger,
+	remoteClient client.Client,
+) result.ReconcileResult {
+	logger.Info("reconciling telemetry", "cluster", kc.Name, "datacenter", actualDc.Name)
```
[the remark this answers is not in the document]
Miles-Garnsey: This is true, missed this earlier, thanks.

pkg/telemetry/stargate_prometheus_resourcer.go
Miles-Garnsey: Good catch. Not sure what I was thinking there.

pkg/telemetry/cass_prometheus_resourcer.go
Miles-Garnsey: Good to know, thanks.

controllers/k8ssandra/cassandra_telemetry_reconciler_test.go
Miles-Garnsey: Done, thanks for helping me work through some of that too.

controllers/stargate/stargate_telemetry_reconciler_test.go
Miles-Garnsey: Also done.

pkg/telemetry/cass_telemetry_resourcer.go
Miles-Garnsey: Stylistically, I find this easier to read, but if you're concerned about allocating memory for an additional slice I can change it. Let me know.

Miles-Garnsey:
All requested changes made I think now, specifically:

* Envtests added for both stargate telemetry reconciliation and Cassandra DC reconciliation. 
* All `pkg/telemetry` functions refactored. This eliminates duplication between CassandraDatacenter  telemetry and Stargate telemetry update/cleanup.
* Eliminated a layer of indirection in `pkg/telemetry` which is not currently needed (although may come back at a later date).

Tests passing locally, once they are passing on GHA is this ready for merge @jsanda?

pkg/telemetry/stargate_prometheus_resourcer.go
Miles-Garnsey: Done.

pkg/telemetry/stargate_prometheus_resourcer.go
Miles-Garnsey: Done.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+	switch {
+	case mergedSpec == nil:
+		logger.Info("Telemetry not present for CassDC, will delete resources", "mergedSpec", mergedSpec)
+		if err := cfg.CleanupResources(ctx, remoteClient); err != nil {
+			return result.Error(err)
+		}
+		return result.Continue()
+	case mergedSpec.Prometheus == nil:
+		logger.Info("Telemetry not present for CassDC, will delete resources", "mergedSpec", mergedSpec)
+		if err := cfg.CleanupResources(ctx, remoteClient); err != nil {
```
jsanda, line 61: `CleanupResources` uses `DeleteAllOf`,  but will there be scenarios in which we have more than one ServiceMonitor for a CassandraDatacenter? If not then we look for the single ServiceMonitor (if it exists) to be deleted. We can perform the lookup easily enough since the ServiceMonitor name is basically pre-computed. Note that if we could not easily compute the ServiceMontior name, then we should be able to query for the ServiceMonitor using a label selector.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -195,6 +196,12 @@ type CassandraClusterTemplate struct {
 	// +optional
 	Datacenters []CassandraDatacenterTemplate `json:"datacenters,omitempty"`
 
+	// CassandraTelemetry defines the desired state for telemetry resources in this K8ssandraCluster.
+	// If telemetry configurations are defined, telemetry resources will be deployed to integrate with
+	// a user-provided monitoring solution (at present, only support for Prometheus is available).
+	// +optional
+	CassandraTelemetry *telemetryapi.TelemetrySpec `json:"cassandraTelemetry,omitempty"`
```
jsanda, line 203: Can we change the json field name to just `telemetry`? You already have this for Stargate.

pkg/telemetry/prom_stargate_servicemonitor_test.go
```diff
@@ -0,0 +1,155 @@
...
+      - regex: persistence_cassandra_(\d_\d+)_org_apache_cassandra_metrics_ClientRequest_ContentionHistogram_(Read|Write|CASRead|CASWrite|RangeSlice|ViewWrite)_count
+        replacement: stargate_client_request_contention_histogran_total
+        sourceLabels:
+        - __name__
+        targetLabel: __name__
+`
+
+var stargateServiceMonitorTemplate = &promapi.ServiceMonitor{}
+
+func init() {
```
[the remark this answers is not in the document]
Miles-Garnsey: This isn't technically a problem to do, however I agree with [this](https://stackoverflow.com/a/45278760) Stack Overflow answer which suggests that sometimes multiple init functions can make the code easier to follow. 

In this case, if you look at the functionality of the init functions, they are there purely to parse the huge ServiceMonitor template string into a Go struct statically (you previously asked for static initialisation of this object). While we could certainly combine the files into one, this leads to 1x300 line string and 1*100 line string in a single file, which isn't very ergonomic just in terms of scrolling around the IDE.

Can you expand on what benefits combining these two files would yield?

pkg/telemetry/prometheus_resourcer.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			return nil
+		} else {
+			cfg.Logger.Error(err, "could not get ServiceMonitor resource")
+			return err
+		}
+	}
+	// Logic to handle case where SM exists, but is in the wrong state.
+	actualSM = actualSM.DeepCopy()
+	if !annotations.CompareAnnotations(actualSM, desiredSM, k8ssandraapi.ResourceHashAnnotation) {
```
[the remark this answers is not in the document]
Miles-Garnsey: I don't see the purpose of the additional layer of indirection here, it seems to make things more complicated. But done anyway.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+		return result.Continue()
+	}
+}
+
+// mustLabels() returns the set of labels essential to managing the Prometheus resources. These should not be overwritten by the user.
+func mustLabels(klusterName string, dcName string) map[string]string {
+	return map[string]string{
+		k8ssandraapi.ManagedByLabel:            k8ssandraapi.NameLabelValue,
+		k8ssandraapi.PartOfLabel:               k8ssandraapi.PartOfLabelValue,
+		k8ssandraapi.K8ssandraClusterNameLabel: klusterName,
```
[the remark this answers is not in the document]
Miles-Garnsey: I can see your point here, done.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
Miles-Garnsey: You [asked for ](https://github.com/k8ssandra/k8ssandra-operator/pull/241#discussion_r766950899)this change earlier? 

In either event, I prefer to use `DeleteAllOf` just in case we eventually allow the user to set their own name on the `ServiceMonitor`, it seems more idiomatic to work with labels than names.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: Yes, because `Telemetry` for Stargate is wrapped in a `StargateTemplate`, which makes the target object clear. Conversely, `CassandraTelemetry` is a member of `CassandraClusterTemplate.Datacenters` of type `CassandraDatacenterTemplate` which also contains members (for e.g.) `Stargate`, `Reaper` and various other non-Cassandra structs.

It is therefore unclear what the telemetry target is if we simply add `Telemetry` as a field here. 

In fact, I think this API design could use some thought, but that isn't a matter for this PR.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+	if !validConfig {
+		return result.Error(errors.New("telemetry spec was invalid for this cluster - is Prometheus installed if you have requested it"))
+	}
+	// If Prometheus not installed bail here.
+	if !promInstalled {
+		return result.Continue()
+	}
+	// Determine if we want a cleanup or a resource update.
+	switch {
+	case mergedSpec == nil:
```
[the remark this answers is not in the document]
Miles-Garnsey: I do not believe that these can be combined. You propose to combine:
```
case mergedSpec == nil: ...
case mergedSpec.Prometheus == nil: ...
default:
```

Which means that the condition `case mergedSpec.Prometheus.Enabled:` is hit first. In the case that either `mergedSpec` or `mergedSpec.Prometheus` is nil, this will lead to a nil pointer deference.

You need to go step by step to ensure that each of the members exist, that's why the switch statement is there.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
+	ctx context.Context,
+	kc *api.K8ssandraCluster,
+	dcTemplate api.CassandraDatacenterTemplate,
+	actualDc *cassdcapi.CassandraDatacenter,
+	logger logr.Logger,
+	remoteClient client.Client,
+) result.ReconcileResult {
+	logger.Info("reconciling telemetry")
+	mergedSpec := kc.Spec.Cassandra.CassandraTelemetry.Merge(dcTemplate.CassandraTelemetry)
```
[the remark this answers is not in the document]
Miles-Garnsey: Bugger! That's a good catch! I am going to make this change immediately actually, because the behaviour for [`TelemetrySpec.Merge()`](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/de687f0219da39a26d537e01273f0ef55f9d5121/apis/telemetry/v1alpha1/telemetry_methods.go#L10) is currently inconsistent with the behaviour for [`PrometheusTelemetrySpec.Merge()`](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/de687f0219da39a26d537e01273f0ef55f9d5121/apis/telemetry/v1alpha1/telemetry_methods.go#L31).

I'll also add logic to bail early if telemetry is `nil`.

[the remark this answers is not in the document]
Miles-Garnsey: Actually, I'm removing that early bail when `mergedSpec == nil`. That would prevent deletion of resources if relemetry is removed from the spec. I don't think the optimisation you're proposing is possible.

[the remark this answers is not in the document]
Miles-Garnsey: The `Merge()` behaviour changes remain however, that comment was still very valid.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
jsanda: On my first review, I only noticed the method deleting objects could potentially be simplified by using `DeleteAllOf` On my subsequent review(s) when I had a better understanding of the code it occurred to me that `DeleteAllOf` may not even necessary if we are only dealing with a single ServiceMonitor. Apologies for any confusion that I caused.

> I prefer to use DeleteAllOf just in case we eventually allow the user to set their own name on the ServiceMonitor

We will know the ServiceMonitor name whether it is computed or it was specified by the user, right? If specified, it would be included in the spec. More importantly, using `DeleteAllOf` increases security surface area as we have to add a permission that really isn't necessary. 

>it seems more idiomatic to work with labels than names

We delete objects by key (i.e., types.NamespacedName / client.ObjectKey) pretty much everywhere except for a couple places in tests. I think it's the same for cass-operator.

pkg/test/testenv.go
```diff
@@ -32,9 +33,10 @@ import (
 )
 
 const (
-	clustersToCreate    = 3
-	clusterProtoName    = "cluster-%d"
-	cassOperatorVersion = "v1.9.0"
+	clustersToCreate          = 3
+	clusterProtoName          = "cluster-%d"
+	cassOperatorVersion       = "v1.8.0"
```
jsanda, line 38: This should be v1.9.0. It determines the versions of the CassandraDatacenter CRD to install.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jsanda: This along with some other naming issues are certainly subjective, and I agree a matter for this PR. I will start some separate discussions though before 1.0.0.

pkg/test/testenv.go
Miles-Garnsey: I think that must have been a regression introduced during rebasing somewhere. My bad.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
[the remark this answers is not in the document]
Miles-Garnsey: I don't have a problem implementing it this way if you prefer. It does not affect the behaviour of the reconciliation in either event.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
Miles-Garnsey: I don't think it makes a difference to any hypothetical attack surface given that an attacker can list and individually delete resources as easily as running DeleteAllOf. It is odd to me that these are even distinct permissions in the RBAC. 

Regardless, if you'd prefer it implemented as a single deletion then I'll do so - it will make no difference to the behaviour of the function.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jdonenine: So the naming aspect is indeed something @jsanda just mentioned wanting to generally review the other day, that's one of the areas we wanted to get some feedback from some external to the team users on what makes sense.  So perhaps we should like you both seem to say save that conversation for a different moment.

From the looks of what we have here it seems we're already mixing patterns in this structure

```
type CassandraClusterTemplate struct {
	// Cluster is the name of the cluster. This corresponds to cluster_name in
	// cassandra.yaml.
	// +kubebuilder:validation:MinLength=2
	Cluster string `json:"cluster,omitempty"`

	// SuperuserSecretName allows to override the default super user secret
	SuperuserSecretName string `json:"superuserSecret,omitempty"`

	// ServerImage is the image for the cassandra container. Note that this should be a
	// management-api image. If left empty the operator will choose a default image based
	// on ServerVersion.
	// +optional
	ServerImage string `json:"serverImage,omitempty"`

	// ServerVersion is the Cassandra version.
	// +kubebuilder:validation:Pattern=(3\.11\.\d+)|(4\.0\.\d+)
	ServerVersion string `json:"serverVersion,omitempty"`

	// Resources is the cpu and memory resources for the cassandra container.
	// +optional
	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`

	// SystemLoggerResources is the cpu and memory resources for the server-system-logger
	// container.
	// +optional
	SystemLoggerResources *corev1.ResourceRequirements `json:"systemLoggerResources,omitempty"`

	// CassandraConfig is configuration settings that are applied to cassandra.yaml and
	// jvm-options for 3.11.x or jvm-server-options for 4.x.
	// +optional
	CassandraConfig *CassandraConfig `json:"config,omitempty"`

	// StorageConfig is the persistent storage requirements for each Cassandra pod. This
	// includes everything under /var/lib/cassandra, namely the commit log and data
	// directories.
	// +optional
	StorageConfig *cassdcapi.StorageConfig `json:"storageConfig,omitempty"`

	// Networking enables host networking and configures a NodePort ports.
	// +optional
	Networking *cassdcapi.NetworkingConfig `json:"networking,omitempty"`

	// Racks is a list of named racks. Note that racks are used to create node affinity. //
	// +optional
	Racks []cassdcapi.Rack `json:"racks,omitempty"`

	// Datacenters a list of the DCs in the cluster.
	// +optional
	Datacenters []CassandraDatacenterTemplate `json:"datacenters,omitempty"`

	// CassandraTelemetry defines the desired state for telemetry resources in this K8ssandraCluster.
	// If telemetry configurations are defined, telemetry resources will be deployed to integrate with
	// a user-provided monitoring solution (at present, only support for Prometheus is available).
	// +optional
	CassandraTelemetry *telemetryapi.TelemetrySpec `json:"cassandraTelemetry,omitempty"`

	// MgmtAPIHeap defines the amount of memory devoted to the management
	// api heap.
	// +optional
	MgmtAPIHeap *resource.Quantity `json:"mgmtAPIHeap,omitempty"`
}
```

Since everything is under the `CassandraClusterTemplate` it seems clear to me that everything inside of it is relevant to `Cassandra`, so I'd certainly be inclined to say that calling it `CassandraTelemetry` is redundant and not necessary.  I'd fall on the side of simplicity there.

But, we also have `CassandraConfig`, which doubles up on the naming as well -- I'd personally support a pattern where we don't repeat the naming when the parent already presents it.  So I'd suggest that we simplify this, and should also consider doing the same for `CassandraConfig`.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
Miles-Garnsey: When I started to implement this some complications came up around the assumptions inherent in the `client.Get(ctx, dTarget)... client.Delete(ctx, dTarget)` sequence. It assumes that the deletion target definitely exists, and that assumption is not concordant with our use case here (cleanup is called whenever the telemetry is not explicitly enabled, even if no telemetry resources actually exist).

Not moving forward with this because the current usage of `DeleteAllOf` seems more appropriate for our use case.

jsanda (approved):
Thanks!

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
Miles-Garnsey: @jdonenine there's two different instances of CassandraTelemetry [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/80e661d3db8b3ff20b3938b49f9bcedfb423c7a1/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L275) and [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/stargate-prometheus-integration/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L203). 

Now, the former one on line 275 is in a struct (`CassandraDatacenterTemplate`) which mixes in Reaper and Stargate. The one you've referenced on line 203 is indeed in a pure Cassandra struct (and, very confusingly, resides in `CassandraClusterTemplate`).

So we have a situation where a `CassandraDatacenterTemplate` mixes Reaper and Stargate structs in with Cassandra structs, but `CassandraClusterTemplate` is purely Cassandra related. I certainly wouldn't call the former one `Telemetry` and the latter one `CassandraTelemetry` because they both do exactly the same thing.

The API structure could use a few tweaks in my opinion.  The Stargate API is a lot cleaner and would be a good reference (single type for both cluster + DC config, more aggressive use of inline json/embedded structs). At the least, it would be worth segregating the Cassandra config into it's own api package to clarify the structure.

_(You can see why I didn't want to get into this here...)_

PR #261, opened 2021-12-16
Enable auth by default [K8SSAND-935]

### LINKED ISSUE #170: K8SSAND-935 ⁃ Enable auth by default
**What is missing?**

In K8ssandra 1.x authentication is enabled by default, it should also be the default for K8ssandraClusters deployed by the operator.

**Why do we need it?**

To support best practices.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-935) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-935
┆priority: Medium

adutra:
**What this PR does**:
Adds support for enabling authentication in the cluster.

Fixes #170 

**Checklist**
- [x] Changes manually tested


apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -277,38 +287,68 @@ type EmbeddedObjectMeta struct {
 // TODO Implement Stringer interface. It will helpful for debugging and testing.
 type CassandraConfig struct {
 	// +optional
-	CassandraYaml *CassandraYaml `json:"cassandraYaml,omitempty"`
+	CassandraYaml CassandraYaml `json:"cassandraYaml,omitempty"`
```
adutra, line 290: @jsanda just to make sure you are on board here: I realized that both `CassandraYaml` and `JvmOptions` always end up non-nil due to a lot of default entries being applied during reconciliation, so I figured it would be much easier to remove the pointers altogether.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -277,38 +287,68 @@ type EmbeddedObjectMeta struct {
...
+	SuperuserSecretRef corev1.LocalObjectReference `json:"superuserSecretRef,omitempty"`
 
+	// The cache validity period in milliseconds to apply to all authentication caches in Cassandra (roles, permissions
+	// and credentials). If defined, this value is automatically applied to the following cassandra.yaml settings:
+	// - roles_validity_in_ms
+	// - permissions_validity_in_ms
+	// - credentials_validity_in_ms
+	// Setting this value to 0 disables caching entirely.
 	// +optional
 	CacheValidityPeriodMillis *int64 `json:"cacheValidityPeriodMillis,omitempty"`
```
adutra, line 310: @jsanda I kept these 2 cache settings around since you had them initially, but honestly they force us to pass the `CassandraAuth` struct around which is a bit cumbersome. Do you see a lot of value in these settings?

controllers/k8ssandra/datacenters.go
```diff
@@ -48,14 +48,20 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 		// its fields are pointers, and without the copy we could end of with shared
 		// references that would lead to unexpected and incorrect values.
 		dcConfig := cassandra.Coalesce(kc.Spec.Cassandra.DeepCopy(), dcTemplate.DeepCopy())
-		cassandra.ApplySystemReplication(dcConfig, *systemReplication)
+
+		cassandra.ApplyAuthSettings(dcConfig, kc.Spec.Auth)
+
+		if kc.Spec.Auth {
+			cassandra.ApplySystemReplication(dcConfig, *systemReplication)
```
adutra, line 55: @jsanda in K8ssandra 1.x these settings are only applied if auth is enabled. Is that accurate?

controllers/k8ssandra/reaper.go
```diff
@@ -19,56 +19,26 @@ package k8ssandra
...
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
-	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 )
 
-// TODO should we move this to secrets.go?
```
adutra, line 39: Done :-)

controllers/k8ssandra/secrets.go
```diff
@@ -4,28 +4,66 @@ import (
...
+			logger = logger.WithValues("ReaperCassandraUserSecretRef", cassandraUserSecretRef)
+			if err := secret.ReconcileSecret(ctx, r.Client, cassandraUserSecretRef.Name, kcKey); err != nil {
+				logger.Error(err, "Failed to reconcile Reaper CQL user secret")
+				return result.Error(err)
+			}
+		} else {
+			logger.Info("Auth is disabled: skipping CQL user secret reconciliation")
+		}
+
+		// The JMX secret must be reconciled even when auth is disabled
```
adutra, line 52: This was my guess but I am not sure here. Should we instead allow JMX remote connections without auth if auth is disabled globally?

pkg/cassandra/util.go
```diff
@@ -32,7 +32,7 @@ func DatacenterStopping(dc *cassdcapi.CassandraDatacenter) bool {
 }
 
 func ComputeSystemReplication(kluster *api.K8ssandraCluster) SystemReplication {
-	rf := 3.0
+	rf := 5.0
```
adutra, line 35: @jsanda in K8ssandra 1.x we have 5, in 2.x we had 3 so far. I updated to 5 FYI.

pkg/reaper/datacenter.go
```diff
@@ -52,20 +53,51 @@ func addInitContainer(reaperTemplate *reaperapi.ReaperClusterTemplate, dcConfig
...
 			MountPath: "/config",
 		}},
-	})
+	}
+	if authEnabled {
+		// Note: dcConfig.Auth.SuperuserSecretRef should never be empty here
+		jmxInitContainer.Env = append(
+			jmxInitContainer.Env,
+			corev1.EnvVar{
+				Name: "SUPERUSER_JMX_USERNAME",
```
adutra, line 72: This reflects K8ssandra 1.x but I don't really understand why we need to add those env vars. Reaper itself is not going to use them to open JMX connections, so they are only useful imo if a human attempts to open a JMX connection to the node using the superuser credentials. Am I missing something?

pkg/reaper/datacenter.go
```diff
@@ -7,43 +7,44 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func AddReaperSettingsToDcConfig(reaperTemplate *reaperapi.ReaperClusterTemplate, dcConfig *cassandra.DatacenterConfig) {
-	addUser(reaperTemplate, dcConfig)
+func AddReaperSettingsToDcConfig(reaperTemplate *reaperapi.ReaperClusterTemplate, dcConfig *cassandra.DatacenterConfig, authEnabled bool) {
+	if authEnabled {
+		addUser(reaperTemplate, dcConfig)
```
adutra, line 12: I wonder: if the user specified a CQL user secret in the Reaper CRD, but auth is disabled globally, should we go ahead and create that user?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Especially keeping in mind that we already have the cluster vs dc template overrides that could be used to specify default settings.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1298,14 +1284,6 @@ func parseResource(quantity string) *resource.Quantity {
 	return &parsed
 }
 
-type fakeSeedsResolver struct {
```
adutra, line 1301: Unused.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1318,43 +1296,3 @@ func (f fakeManagementApiFactory) NewManagementApiFacade(context.Context, *cassd
 	})).Return(nil)
 	return m, nil
 }
-
-// verifySecretsMatch checks that the same secret is copied to other clusters
-func verifySecretsMatch(t *testing.T, ctx context.Context, localClient client.Client, remoteClusters []string, secrets map[string]struct{}, namespace string) bool {
```
adutra, line 1323: Unused.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -30,6 +30,14 @@ import (
 
 // K8ssandraClusterSpec defines the desired state of K8ssandraCluster
 type K8ssandraClusterSpec struct {
+
+	// Whether to enable authentication in this cluster. The default is true; it is highly recommended to always leave
+	// authentication turned on. When enabled, authentication will be enforced not only on Cassandra nodes, but also on
+	// Reaper, Medusa and Stargate nodes, if any.
+	// +optional
+	// +kubebuilder:default=true
+	Auth bool `json:"auth"`
```
adutra, line 39: TIL: you can't set `omitempty` here because that would cause the Go client to not include `auth:false` in the JSON payload, even if you explicitly set `Auth: false` in the struct. I guess this is because `false` is the zero value for `bool`.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jsanda: That works for me provided it doesn't requiring the user to define something for those fields.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Actually it's even trickier. I think it's impossible to have a `bool` field with default value `true`. I'll switch to `*bool`.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
jsanda: I'd get rid of them.

adutra:
Here is the current state of this PR wrt to the action items I listed [here](https://github.com/k8ssandra/k8ssandra-operator/issues/170#issuecomment-995653889):

* Global boolean switch to enable / disable auth: implemented
* Cassandra changes: implemented
* Reaper changes: implemented
* Stargate changes: not implemented
* Medusa changes: not implemented, waiting for #242.

Tests:
* Unit tests: implemented;
* Controller tests: one test with auth disabled implemented;
* E2e tests: no new test implemented yet, but existing tests are now all using Cassandra auth.

I'd suggest that we tackle Stargate and Medusa later on, let's focus on getting Cassandra and Reaper with auth enabled working + tests in this PR.

If someone wants to continue the work while I'm away, please go ahead!

controllers/k8ssandra/datacenters.go
jsanda: That is correct; however, I think it would be fine to call `cassandra.ApplySystemReplication` even when auth is disabled. The amount of extra data from the system_auth keyspace should be quite small. It would be one less thing to worry about if auth is initially disabled and then later enabled.

controllers/k8ssandra/secrets.go
jsanda: If Reaper is enabled, we need to enable remote JMX access. If we enable remote JMX access, I think it should always have JMX auth enabled as well regardless of whether Cassandra auth is enabled.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Everything inside CassandraYaml and JvmOptions is a pointer, so I think we are good.

pkg/reaper/datacenter.go
adutra: One thing to keep in mind: if the secret exists, the Reaper controller will add an extra `REAPER_CASS_AUTH_ENABLED=true` env var to the Reaper pods. This is likely to make Reaper startup fail if auth is disabled in Cassandra.

pkg/cassandra/util.go
jsanda: Since that time I have heard multiple folks say RF = 3 should be sufficient for the system_auth which I made the change in k8ssandra-operator. Maybe it's worth revisiting.

pkg/reaper/datacenter.go
jsanda: Good question. I don't have a strong opinion one way or the other at this point. This brings up another question though. If the user has auth enabled and then subsequently disables it, should delete any secrets and users that we create?

pkg/reaper/datacenter.go
jsanda: We decided it would be best to reuse the superuser credentials for JMX. If/when we address https://github.com/k8ssandra/k8ssandra/issues/323, this would go away.

pkg/reaper/datacenter.go
adutra: Indeed, I haven't thought about that. I guess we should delete them if they were created by us, otherwise leave them be (as orphaned resources) if they were user-created. I'll add this item to my TODO list, not sure I'll have time to tackle it today.

pkg/cassandra/util.go
adutra: Thanks for the context! I'll set the default back to 3 in this PR, we can revisit that later imo.

pkg/reaper/datacenter_test.go
```diff
@@ -28,13 +212,11 @@ func TestAddReaperSettingsToDcConfig(t *testing.T) {
...
 				Cluster: "cluster1",
-				Users:   []cassdcapi.CassandraUser{{SecretName: "cluster1-reaper", Superuser: true}},
 				PodTemplateSpec: &corev1.PodTemplateSpec{
 					Spec: corev1.PodSpec{
 						InitContainers: []corev1.Container{{
-							Name:            "jmx-credentials",
-							Image:           "docker.io/busybox:1.33.1",
-							ImagePullPolicy: corev1.PullIfNotPresent,
+							Name:  "jmx-credentials",
+							Image: "docker.io/busybox:1.33.1",
```
jsanda, line 219: We'll need to make this image configurable in order to support using different or private registries.

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	k8s.io/apimachinery v0.22.2
 	k8s.io/client-go v0.22.2
 	k8s.io/kubernetes v1.22.2
+	k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a
```
adutra, line 24: This is because I'm using the `pointer` package.

pkg/reaper/datacenter_test.go
adutra: I added a field to control this image, but I realize that this field cannot be honored when using standalone Reaper resources. This is because the field contents are only read and applied by k8ssandra-controller, not by reaper-controller. Is that OK for you?

pkg/reaper/datacenter.go
adutra: Thank you for the pointer. Indeed when that ticket is implemented, none of these env vars would be required anymore.

However even when using the "old-style" JMX authentication through a `jmxremote.password` file, I still don't see why we need `SUPERUSER_JMX_USERNAME` and `SUPERUSER_JMX_PASSWORD`:

Indeed, Reaper itself is going to use `REAPER_JMX_USERNAME` and `REAPER_JMX_PASSWORD` to authenticate its JMX connections. These vars are always created and written to `jmxremote.password`, even when auth is disabled.

As far as I can tell, no component is making any use of the 2 extra vars that get created only when auth is enabled: `SUPERUSER_JMX_USERNAME` and `SUPERUSER_JMX_PASSWORD`. So I'm wondering if these 2 are necessary at all.

I will do a local test to verify this.

jsanda: You are correct about their not being used. Now I remember how we got here :) Initially, JMX auth was only turned on when Reaper was enabled. Then it was decided that it should be enabled regardless, but I believe a different set of credentials were generated. Users assumed that the superuser credentials should be used for `nodetool`. This caused some confusion. It was then decided to reuse the superuser credentials for JMX to make using `nodetool` less confusing.

adutra: Ah, thanks for the explanation, now it makes a lot more sense. 

I now have a doubt: any call to nodetool in an e2e test should fail, and yet, the tests are passing... I need to investigate a bit further.

adutra: OK confirmed:

1. The tests are passing because the current code has a flaw: auth is only being enabled on Cassandra pods when Reaper is also enabled. I need to fix that.
2. If I call `nodetool status` in a test where Reaper is enabled, I do get:

```
error: Authentication failed! Credentials required
-- StackTrace --
java.lang.SecurityException: Authentication failed! Credentials required
	at java.management/com.sun.jmx.remote.security.JMXPluggableAuthenticator.authenticationFailure(Unknown Source)
	at java.management/com.sun.jmx.remote.security.JMXPluggableAuthenticator.authenticate(Unknown Source)
[...]
command terminated with exit code 2
```

This needs to be fixed in test code by passing the appropriate credentials to `kubectl exec`. I will fix that too.

adutra:
Putting this back in draft state as I need to finalize a few more things, sorry.

controllers/k8ssandra/secrets.go
adutra: This turned out to be way more complex than I imagined. 

The reason is that cassandra-env.sh recognizes the `LOCAL_JMX` env var, but local/remote toggling gets mixed with authentication toggling. Indeed to enable remote JMX access we need: `LOCAL_JMX=no`; but this also enables JMX auth :-(

In summary, using the `LOCAL_JMX` var:
- It is not possible to enable remote access WITHOUT auth;
- It is not possible to restrict access to localhost WITH auth.

The last bullet point is the most annoying IMO.

The only workaround that I could find is to use the `JVM_EXTRA_OPTS` env var. It would be probably nicer if we could expose `JVM_EXTRA_OPTS` in `CassandraDatacenter` CRD.

I tried to explain the whole thing here:

https://github.com/k8ssandra/k8ssandra-operator/blob/f221f13604589355cabe14c50e2d375aedc81470/pkg/cassandra/auth.go#L18-L30

pkg/reaper/datacenter.go
adutra: I now have successfully decoupled the local vs remote JMX logic from the authentication logic.

In summary:

* If auth is enabled globally, all JMX clients must authenticate, including clients connecting to localhost.
* If auth is disabled globally, all JMX connections are unauthenticated, including remote ones. 
* If Reaper is not present, only local JMX connections are allowed.
* If Reaper is present, local and remote JMX connections are allowed. 

@jsanda does that look reasonable to you?

controllers/k8ssandra/secrets.go
jsanda: Would https://github.com/k8ssandra/k8ssandra/issues/323 help at all?

pkg/reaper/datacenter.go
adutra: > If the user has auth enabled and then subsequently disables it, should delete any secrets and users that we create?

@jsanda I eventually decided to not delete them. Mainly because afaict the replication controller is not able to detect that a `corev1.Secret` object was deleted and that it should also be deleted from other data planes.

So currently, if auth was enabled and is subsequently disabled:

* If Reaper secrets were created manually by the user, they should have the Orphan annotation. Thus they won't ever be deleted, even if the `K8ssandraCluster` CRD is deleted.
* If Reaper secrets were generated by k8ssandra controller, they won't have the Orphan annotation. They won't be deleted when auth is disabled, however they will get deleted if the `K8ssandraCluster` CRD is deleted.

Does that sound OK to you?

I put a TODO here just in case: https://github.com/k8ssandra/k8ssandra-operator/blob/f221f13604589355cabe14c50e2d375aedc81470/controllers/k8ssandra/secrets.go#L60-L61

adutra:
Ready for review again: 

- Global boolean switch to enable / disable auth: implemented
- Cassandra changes: implemented
- Reaper changes: implemented
- Stargate changes: implemented
- Medusa changes: not implemented, waiting for Add medusa APIs #242.

Tests:

- Unit tests: added;
- Controller tests: one test with auth disabled implemented;
- E2e tests: 
  - existing tests are now all using Cassandra auth.
  - new test that toggles auth on and off.

pkg/reaper/datacenter_test.go
adutra: FYI this field has moved to `CassandraClusterTemplate`. 

https://github.com/k8ssandra/k8ssandra-operator/blob/f221f13604589355cabe14c50e2d375aedc81470/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L181-L186

pkg/reaper/resource.go
```diff
@@ -15,8 +15,11 @@ const (
 	DatacenterAvailabilityEach  = "EACH"
 )
 
-func ResourceName(klusterName, dcName string) string {
-	return klusterName + "-" + dcName + "-reaper"
+// DefaultResourceName generates a name for a new Reaper resource that is derived from the Cassandra cluster and DC
+// names.
+func DefaultResourceName(dc *cassdcapi.CassandraDatacenter) string {
+	// FIXME sanitize name
```
adutra, line 21: I'm waiting on #267 for this, but I had to at least align Cassandra, Stargate and Reaper names.

test/e2e/suite_test.go
```diff
@@ -72,9 +69,11 @@ func TestOperator(t *testing.T) {
 		fixture:  "multi-dc",
 	}))
 	t.Run("CreateMultiStargateAndDatacenter", e2eTest(ctx, &e2eTestOpts{
-		testFunc:      createMultiDatacenterCluster,
-		fixture:       "multi-dc",
-		deployTraefik: true,
+		testFunc:                     createStargateAndDatacenter,
```
adutra, line 72: This test was not correctly configured.

pkg/reaper/datacenter.go
adutra: Also note that the Cassandra superuser secret is always created and replicated, regardless of authentication.

I explained why here: https://github.com/k8ssandra/k8ssandra-operator/blob/f221f13604589355cabe14c50e2d375aedc81470/controllers/k8ssandra/secrets.go#L14-L19

controllers/k8ssandra/secrets.go
adutra: Unfortunately no, however the changes in this PR will make it easier to switch to Cassandra internals for JMX authentication later on.

Note that IMO we'll need to use `JVM_EXTRA_OPTS` to do that as well.

adutra:
Also worth noting: Reaper REST API and WEB UI are currently unauthenticated, we may want to look into enforcing authentication there in a follow-up PR.

The WEB UI is not used in code nor in tests. The REST API however is used in code (reaper controller) and in e2e tests.

pkg/reaper/datacenter.go
jsanda: Sounds good

pkg/cassandra/config.go
```diff
@@ -37,129 +36,82 @@ func IsCassandra3(version string) bool {
...
 
-	if c.JvmOptions != nil {
-		if strings.HasPrefix(c.cassandraVersion, "3.11") {
-			config["jvm-options"] = c.JvmOptions
+	// Even though we default to Cassandra's stock defaults for num_tokens, we need to
+	// explicitly set it because the config builder defaults to num_tokens: 1
+	if c.NumTokens == nil {
+		if isCassandra4(c.cassandraVersion) {
+			numTokens := 16
+			c.NumTokens = &numTokens
```
jsanda, line 55: This could be changed to `c.NumTokens = pointer.Int(16)`. I only thought of it since I saw you introduce usage of the `pointer` package elsewhere :)

pkg/cassandra/config.go
```diff
@@ -37,129 +36,82 @@ func IsCassandra3(version string) bool {
...
 		} else {
-			config["jvm-server-options"] = c.JvmOptions
+			numTokens := 256
+			c.NumTokens = &numTokens
```
jsanda, line 58: Similar to above we could use `pointer.Int()` here.

pkg/cassandra/auth.go
```diff
@@ -0,0 +1,113 @@
...
+	// script, which means that any value set there will override what the script itself had previously computed.
+	// FIXME this is a bit hacky and might interfere with the way cass-operator enables DSE workloads
+	if dcConfig.PodTemplateSpec == nil {
+		dcConfig.PodTemplateSpec = &corev1.PodTemplateSpec{}
+	}
+	UpdateCassandraContainer(dcConfig.PodTemplateSpec, func(c *corev1.Container) {
+		jmxAuthenticateOpt := fmt.Sprintf("-Dcom.sun.management.jmxremote.authenticate=%v", authEnabled)
+		c.Env = append(c.Env,
+			corev1.EnvVar{
+				Name:  "JVM_EXTRA_OPTS",
```
jsanda, line 39: Can you write up a ticket for making `JVM_EXTRA_OPTS` configurable. I don't think this is something we will get done for 1.0.0, but it would be a nice enhancement. There is a lot in flight right now, and I don't want it to fall through the cracks.

controllers/k8ssandra/secrets.go
```diff
@@ -4,28 +4,67 @@ import (
...
+				return result.Error(err)
+			}
+			if err := secret.ReconcileSecret(ctx, r.Client, jmxUserSecretRef.Name, kcKey); err != nil {
+				logger.Error(err, "Failed to reconcile Reaper JMX user secret", "ReaperJmxUserSecretRef", jmxUserSecretRef)
+				return result.Error(err)
+			}
+			logger.Info("Reaper user secrets successfully reconciled")
+		} else {
+			// TODO if auth was previously enabled, Reaper secrets may have been created;
+			// now that auth has been disabled, what should we do with the dangling secrets?
```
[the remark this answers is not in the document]
adutra: Indeed, that is the behavior _when the K8ssandraCluster is deleted_. 

But here we would like to delete just the Reaper secrets, _without deleting the whole K8ssandraCluster_, and propagate the deletion to all contexts. I don't think we have support for this scenario, so I think the safest choice is to do nothing :-).

I will update the comment and explain why we don't delete the Reaper secrets here.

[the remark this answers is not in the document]
jsanda: Thanks for clarifying. I also think doing nothing is fine in this case.

pkg/stargate/deployments.go
```diff
@@ -201,6 +200,20 @@ func NewDeployments(stargate *api.Stargate, dc *cassdcapi.CassandraDatacenter) m
...
+		if stargate.Spec.IsAuthEnabled() {
+			// Stargate reacts to the sole presence of this variable, regardless of its contents.
+			// When this variable is absent, Stargate will use AllowAllAuthenticator.
+			// When this variable is present, Stargate will by default use PasswordAuthenticator, unless overridden
+			// by the stargate.authenticator_class_name system property (see below, computeJvmOptions).
+			// Note that any other authenticator than PasswordAuthenticator will cause the REST APIs to be unusable,
+			// however the CQL API will still be usable.
+			deployment.Spec.Template.Spec.Containers[0].Env = append(
+				deployment.Spec.Template.Spec.Containers[0].Env,
+				corev1.EnvVar{Name: "ENABLE_AUTH", Value: "true"},
```
jsanda, line 213: Doesn't this env var enable Stargate's token based auth for its api requests? At least in earlier versions of Stargate when we did the initial integration with k8ssandra, we learned that it always had to be enabled regardless of whether Cassandra auth was enabled.

adutra: I believe this isn't the case anymore. `ENABLE_AUTH` will trigger this:

https://github.com/stargate/stargate/blob/136f88bbf7417706043f045ec2ccb9331f4b0f75/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/Cassandra40PersistenceActivator.java#L109-L116

So basically it sets the authenticator to PasswordAuthenticator. Afaict it has no implications in the token based auth for the REST API.

jsanda: I stand corrected. I looked at Stargate code and confirmed that `ENABLE_AUTH` does indeed configure PasswordAuthenticator just as the comment here describes. 

Is the toke based api auth always enabled?

pkg/cassandra/auth.go
adutra: In fact there is a way to achieve what we want without resorting to `JVM_EXTRA_OPTS`: in config-builder, the cassandra-env.sh template also accepts an `additional-jvm-opts` parameter. I suggest that we use it instead, I will make the necessary changes.

adutra: Done here: https://github.com/k8ssandra/k8ssandra-operator/blob/0d829b9246e62870d9345ada8660153a049bb6f8/pkg/cassandra/config.go#L71-L78.

pkg/stargate/deployments.go
adutra: Yes, you cannot disable it. And it is table-based by default, unless you configure a JWT server.

jsanda (approved):
Thanks!

adutra:
FYI I'm going to merge this now, even if some e2e tests are failing (the failures are unrelated).

PR #262, opened 2021-12-17
Add datacenter to existing cluster [K8SSAND-615][K8SSAND-784]

### LINKED ISSUE #21: K8SSAND-615 ⁃ Add support for adding a datacenter to an existing cluster

jsanda:
**What this PR does**:

Fixes #21

**Checklist**
- [ ] Changes manually tested


Makefile
```diff
@@ -225,8 +229,9 @@ create-kind-multicluster:
 	scripts/setup-kind-multicluster.sh --clusters 2 --kind-worker-nodes 4
```
jdonenine, line 229: We'll want to update this script to use the new variable...

```suggestion
	scripts/setup-kind-multicluster.sh --clusters $(NUM_CLUSTERS) --kind-worker-nodes 4
```

I think should do the trick

jsanda:
I think we need to change when we reconcile the Stargate auth and Reaper schemas. The schemas should be reconciled after each CassandraDatacenter is reconciled.

Here is how things are currently done in K8ssandraReconciler (this is in the `main` branch as well):

```go
	var actualDcs []*cassdcapi.CassandraDatacenter
	if recResult, dcs := r.reconcileDatacenters(ctx, kc, kcLogger); recResult.Completed() {
		return recResult.Output()
	} else {
		actualDcs = dcs
	}

	kcLogger.Info("All dcs reconciled")

	if recResult := r.reconcileStargateAuthSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
		return recResult.Output()
	}

	if recResult := r.reconcileReaperSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
		return recResult.Output()
	}

	if recResult := r.reconcileStargateAndReaper(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
		return recResult.Output()
	}
```
We first reconcile all CassandraDatacenters. We will requeue until all DCs are ready. Then we reconcile the Stargate auth and Reaper schemas. These keyspaces are created if necessary. Lastly, we reconcile Stargate and Reaper across all DCs. In other words, `reconcileStargateAndReaper` iterates through the CassandraDatacenters and adds/updates/removes Stargate and Reaper as necessary for each one.

There are a number of changes to the `reconcileDatacenters` method in my `add-dc` branch. I'll try to highlight the relevant parts and leave out the rest:

```go 
	for idx, dcTemplate := range kc.Spec.Cassandra.Datacenters {
                ...
		if err = remoteClient.Get(ctx, dcKey, actualDc); err == nil {
			...

                         logger.Info("The datacenter is ready")

			actualDcs = append(actualDcs, actualDc)

			if recResult := r.updateReplicationOfSystemKeyspaces(ctx, kc, desiredDc, remoteClient, logger); recResult.Completed() {
				return recResult, actualDcs
			}

			if rebuildNeeded {
				// TODO We need to handle the Stargate auth and Reaper keyspaces here.
				if recResult := r.updateUserKeyspacesReplication(ctx, kc, desiredDc, remoteClient, logger); recResult.Completed() {
					return recResult, actualDcs
				}

				if recResult := reconcileDcRebuild(ctx, actualDc, actualDcs, remoteClient, logger); recResult.Completed() {
					return recResult, actualDcs
				}
			}
		} else {
                        ...
		}
	}
```

After the current DC is ready, `updateReplicationOfSystemKeyspaces` is called. It updates replication for system keyspaces, e.g., system_auth, if necessary. For example, suppose the DC has been scaled up from 1 to 3 C* nodes. The replication factor would change to 3.

If `rebuildNeeded` is `true`, we call `updateUserKeyspacesReplication` to add replicas for the new DC. Lastly, we perform the rebuild by calling `reconcileDcRebuild`.

Note the earlier TODO about handling the Stargate and Reaper keyspaces. If Stargate and Reaper are already installed we need to update the replication for their keyspaces before doing the rebuild so that the keyspaces are included in the streaming. It is easy enough to add the necessary calls where the TODO is.

Now let's say we have an existing K8ssandraCluster that does not have Stargate and Reaper enabled. Suppose we update the K8ssandraCluster by adding a second DC and enabling Stargate and Reaper in each DC. With the code as is, the keyspaces won't get created until after the rebuild of dc2. I think we are ok in this case.

As a last example, consider removing a DC from the cluster. One of the first steps we perform in this process is to stop replicating to the DC. We can't do that if we reconcile the Stargate and Reaper keyspaces after reconciling all DCs. 

Based on my findings I am going to move the calls to reconcile the Stargate and Reaper keyspaces to happen after each DC is ready. 

@adutra any thoughts/concerns?

adutra:
@jsanda sorry for the late reply. The suggested changes make sense and look good to me. I suppose many tests will fail because the order in which they wait for components to show up, but other than that, I don't see any red flags.

config/rbac/role.yaml
```diff
@@ -59,6 +59,18 @@ rules:
...
+- apiGroups:
+  - control.k8ssandra.io
+  resources:
+  - cassandratasks
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
```
burmanm, line 71: Are patch/update needed by k8ssandra-operator?

controllers/k8ssandra/add_dc_test.go
```diff
@@ -0,0 +1,522 @@
...
+					},
+				},
+			},
+		},
+	}
+	kcKey := client.ObjectKey{Namespace: kc.Namespace, Name: kc.Name}
+
+	createSuperuserSecret(ctx, t, f, kcKey, kc.Spec.Cassandra.Cluster)
+
+	createReplicatedSecret(ctx, t, f, kcKey, "cluster-0")
```
burmanm, line 72: When adding a new cluster, we don't actually have any replication rule updates. I don't see here if we actually verify that some secrets are even copied to new clustrs? It's not strictly necessary per-datacenter of course, since datacenter could be replicated to an existing cluster in another namespace, but I wonder if these kind of issues came to mind (at least in documentation)

controllers/k8ssandra/datacenters.go
```diff
@@ -93,6 +103,18 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 
 		if err = remoteClient.Get(ctx, dcKey, actualDc); err == nil {
 			// cassdc already exists, we'll update it
+
+			if _, rebuildNeeded = actualDc.Labels[api.RebuildLabel]; rebuildNeeded {
```
burmanm, line 107: Just above (L85) the rebuildNeeded is parsed. Here we overwrite whatever was parsed there and again set the desiredDc.Labels one. Why?

config/rbac/role.yaml
jsanda: As of right now, they are not needed.

controllers/k8ssandra/datacenters.go
jsanda: I initially determine if a rebuild is needed by looking at the K8ssandraCluster status. If the cluster is already initialized and if dc is not present in the status, then the rebuild label is added. On subsequent reconciliations the datacenter will be added to the status, well before the rebuild happens. The initial evaluation will then result in `rebuildNeeded == false`. This is why it is overwritten based on the presence of the label. After the rebuild finishes, the label is removed.

Do you want me to provide some additional comments explaining this?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -51,31 +48,41 @@ import (
...
+		BeforeTest: func() {
+			managementApiFactory.Reset()
+		},
+	}
 
 	reconcilerConfig := config.InitConfig()
 
-	reconcilerConfig.DefaultDelay = 100 * time.Millisecond
-	reconcilerConfig.LongDelay = 300 * time.Millisecond
+	reconcilerConfig.DefaultDelay = 1 * time.Second
```
burmanm, line 76: Weren't these set to lower values originally to speed up the tests? Is there a reason to make them slower again?

pkg/cassandra/datacenter_test.go
```diff
@@ -357,6 +357,31 @@ func TestNewDatacenter_Fail_NoStorageConfig(t *testing.T) {
 	assert.IsType(t, DCConfigIncomplete{}, err)
 }
 
+func TestDatacentersReplication(t *testing.T) {
+	t.Run("EachDcContainsKeyspaces", testEachDcContainsKeyspaces)
```
burmanm, line 361: What's the point of this func? Why not simply capitalize the testEachDcContainsKeyspaces?

pkg/cassandra/util.go
```diff
@@ -31,14 +33,32 @@ func DatacenterStopping(dc *cassdcapi.CassandraDatacenter) bool {
...
+	var datacenters []api.CassandraDatacenterTemplate
+	if kc.Status.GetConditionStatus(api.CassandraInitialized) == corev1.ConditionTrue {
+		datacenters = make([]api.CassandraDatacenterTemplate, 0)
+		for _, dc := range kc.Spec.Cassandra.Datacenters {
+			if status, found := kc.Status.Datacenters[dc.Meta.Name]; found && status.Cassandra.GetConditionStatus(cassdcapi.DatacenterReady) == corev1.ConditionTrue {
+				datacenters = append(datacenters, dc)
+			}
+		}
+	} else {
+		datacenters = kc.Spec.Cassandra.Datacenters
```
burmanm, line 46: I'm slightly confused here, why do we want only Ready ones if initialized is true, but everything if Initialized is false? Feels unnatural why this path should ever be taken instead of just returning nothing.

burmanm: Or is this path taken when we initialize all the datacenters the first time K8ssandraCluster is created?

pkg/cassandra/datacenter_test.go
jsanda: I had multiple tests at one point. I'll ditch the subtest.

pkg/cassandra/util.go
jsanda: >Or is this path taken when we initialize all the datacenters the first time K8ssandraCluster is created?

short answer - yes

long answer...

We configure replication of system keyspaces by setting a couple system properties:

* `cassandra.system_distributed_replication_dc_names`
* `cassandra.system_distributed_replication_per_dc`

management-api looks for these properties and applies changes to the system_auth keyspace using internal Cassandra APIs, not CQL. If we are creating a cluster with 2 dcs, dc1 and dc2, both will be configured for system_auth.   Note that what I am describing is already the current behavior in main.

In 4.0 Cassandra won't let you specify the replication for a keyspace that does not exist (at least through CQL). I had to introduce changes in this PR for configuring the replication of the Stargate auth and Reaper keyspaces. Those are configured via CQL calls. So going back to the 2 DCs example, when dc1 becomes ready we can create the keyspaces and specify replication for dc1 but not for dc2. Cassandra won't allow this.  This resulted in different handling of system keyspaces vs other keyspaces.

Once I added support for handling non-system keyspaces, I attempted (in vain) to handle system keyspaces the same way to make the code less confusing and more maintainable. There are queries with a `LOCAL_ONE` consistency executed in dc2 that will fail with auth error because the replica(s) in dc2 need a repair.

In summary, until we implement support for running repairs, at least of system_auth, we have to handle system keyspaces the way we currently do. That means 1) for system keyspaces we need to specify replication for all DCs up front and 2) for non-system keyspaces we cannot specify the replication for all DCs up front.

I will rename the method and add some docs to make it more clear that it is only intended for system keyspaces.

controllers/k8ssandra/add_dc_test.go
jsanda: `verifyReplicatedSecretReconciled` is called in each test. It verifies that the `ReplicationDone` condition gets set.

controllers/k8ssandra/replication.go
```diff
@@ -50,24 +74,209 @@ func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context,
...
+func (r *K8ssandraClusterReconciler) updateUserKeyspacesReplication(
+	ctx context.Context,
+	kc *api.K8ssandraCluster,
+	dc *cassdcapi.CassandraDatacenter,
+	mgmtApi cassandra.ManagementApiFacade,
+	logger logr.Logger) result.ReconcileResult {
+
+	jsonReplication := annotations.GetAnnotation(kc, api.DcReplicationAnnotation)
+	if jsonReplication == "" {
+		logger.Info(api.DcReplicationAnnotation + " not set. Replication for user keyspaces will not be updated")
```
burmanm, line 120: Should this be logger.Error() or even a EventRecorder.Eventf() ? I don't think users will notice this kind of notice that everything is fine (result.Continue()) but we didn't actually do anything they requested.

burmanm: Or if this is normal behavior, set the logging to debug or just not log anything?

apis/k8ssandra/v1alpha1/constants.go
```diff
@@ -20,6 +20,12 @@ const (
 	// are still deployed serially).
 	DcReplicationAnnotation = "k8ssandra.io/dc-replication"
 
+	// RebuildSourceDcAnnotation tells the operation the DC from which to stream when
+	// rebuilding a DC. If not set the operator will choose the first DC. The value for
+	// this annotation must specify the name of a CassandraDatacenter whose Ready
+	// condition is true.
+	RebuildSourceDcAnnotation = "k8ssandra.io/rebuild-src-dc"
```
burmanm, line 27: A quick question, we leave this (and the other annotation indicating replicationFactors) to the K8ssandraCluster. Should we remove it after the processing has finished to ensure these rules are reset if someone wants to add another?

jsanda: Good question. I have asked myself the same. Part of the reason I didn't implement behavior to remove them is that I thought users might want to see that state for auditing purposes and what not. Now I also think it could be good to remove them because the user might forget about them. 

In the case of the `k8ssandra.io/dc-replication`, it won't matter. It only applies to the dc being added. So if it left in place and another DC is added and the annotation is ignored it should be the same as not specifying the annotation.

For the `k8ssandra.io/rebuild-src-dc` annotation, I don't think it's a big deal to leave it. Doing so would basically only alter the default behavior.

apis/k8ssandra/v1alpha1/constants.go
```diff
@@ -9,6 +9,25 @@ const (
 	// regardless of whether the replication of the system keyspaces changes.
 	SystemReplicationAnnotation = "k8ssandra.io/system-replication"
```
adutra, line 10: Would it provide a bit more clarity to the code if we renamed this to `InitialSystemReplicationAnnotation`? Indeed afaict this is only used inside `checkSystemReplication` for a newly-created cluster. It doesn't play any role when a new DC is added, correct?

pkg/cassandra/util.go
```diff
@@ -31,14 +33,38 @@ func DatacenterStopping(dc *cassdcapi.CassandraDatacenter) bool {
...
+	}
+	return datacenters
+}
+
+func ComputeSystemReplication(kc *api.K8ssandraCluster) SystemReplication {
```
adutra, line 57: I see that this function is now only called from `checkSystemReplication`, which means that this code is only executed for newly-created clusters, and the purpose here is then to compute the _initial_ system replication. Couldn't we rename this to `ComputeInitialSystemReplication` then?

controllers/k8ssandra/replication.go
```diff
@@ -3,15 +3,39 @@ package k8ssandra
...
+	if len(versions) == 1 {
+		return result.Continue()
+	}
+
+	logger.Info("There is schema disagreement", "versions", len(versions))
+
+	return result.RequeueSoon(r.DefaultDelay)
+}
+
 // checkSystemReplication checks for the SystemReplicationAnnotation on kc. If found, the
```
adutra, line 39: Maybe rename to `checkInitialSystemReplication`?

pkg/cassandra/util.go
```diff
@@ -31,14 +33,38 @@ func DatacenterStopping(dc *cassdcapi.CassandraDatacenter) bool {
...
 	rf := 3.0
-	for _, dc := range kluster.Spec.Cassandra.Datacenters {
+
+	datacenters := GetDatacentersForSystemReplication(kc)
```
adutra, line 60: Couldn't you simplify this to just `	datacenters := kc.Spec.Cassandra.Datacenters` ? Since we determined that this code is only used for computing the initial system replication?

controllers/k8ssandra/datacenters.go
```diff
@@ -37,7 +39,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 	}
 
 	// Reconcile CassandraDatacenter objects only
-	for _, dcTemplate := range kc.Spec.Cassandra.Datacenters {
+	for idx, dcTemplate := range kc.Spec.Cassandra.Datacenters {
```
adutra, line 42: Regarding the handling of system replication:

On line 27 we define or retrieve the initial system replication:

```go
systemReplication, err := r.checkSystemReplication(ctx, kc, logger)
```

Then on line 57 we apply that replication to each `dcConfig`:

```go
cassandra.ApplySystemReplication(dcConfig, *systemReplication)
```

This is OK if the dc being reconciled is an initial one. But if it is a dc that is being added after the cluster is initialized, then this means that we are applying a wrong system replication to its config, and a wrong one that does not include the dc being created. Shouldn't we fix this right here?

Instead, it seems this is going to be "fixed" later on on line 182:

```go
if recResult := r.updateReplicationOfSystemKeyspaces(ctx, kc, mgmtApi, logger); 
```

Indeed the `updateReplicationOfSystemKeyspaces` method behaves like a no-op for initial dcs, but triggers a change in the system keyspaces replication when adding a new dc, which eventually fixes the wrong initial system replication that the dc was created with.

2 questions here:

* Wouldn't it be simpler to just create the new dc with the right replication and let management-api do its magic? 
* If that is not possible and we need to keep things as they are: calling `updateReplicationOfSystemKeyspaces` on an initial dc seems like a waste of time, since we know the replication is OK and so `EnsureKeyspaceReplication` will do nothing. Shouldn't we only call it for dcs being added to an initialized cluster?

adutra (commented):
This is just a partial review that only covers `datacenters.go` and system replication (which is giving me headaches...) Hope my questions make sense.

controllers/k8ssandra/datacenters.go
jsanda: The logic is quite cumbersome. An annotation on the `kc` would probably simplify it a good bit. I'll look into making the change.

scripts/setup-kind-multicluster.sh
```diff
@@ -52,7 +62,7 @@ function create_cluster() {
   num_workers=$3
   node_version=$4
 
-cat <<EOF | kind create cluster --name $cluster_name --image kindest/node:$node_version --config=-
+cat <<EOF | kind create cluster --name $cluster_name -v 5 --image kindest/node:$node_version --config=-
```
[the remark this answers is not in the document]
jsanda: I think I put that in when having issues after a Docker Desktop upgrade. I forgot about it. I'll remove it.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -156,6 +156,18 @@ func (in *K8ssandraCluster) HasReapers() bool {
 	return false
 }
 
+func (in *K8ssandraCluster) GetReadyDatacenters() []CassandraDatacenterTemplate {
```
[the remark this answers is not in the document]
jsanda: Not sure how that would help. `FindDatacenterCondition` simply queries a CassandraDatacenterStatus for a particular condition. `GetReadyDatacenters` is looking for all ready DCs in the K8ssandraCluster.

controllers/k8ssandra/datacenters.go
```diff
@@ -74,6 +76,15 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 		dcKey := types.NamespacedName{Namespace: desiredDc.Namespace, Name: desiredDc.Name}
 		logger := logger.WithValues("CassandraDatacenter", dcKey, "K8SContext", dcTemplate.K8sContext)
 
+		if idx > 0 {
+			desiredDc.Annotations[cassdcapi.SkipUserCreationAnnotation] = "true"
+		}
+
+		rebuildNeeded := datacenterAddedToExistingCluster(kc, desiredDc.Name)
+		if rebuildNeeded {
+			desiredDc.Labels[api.RebuildLabel] = "true"
```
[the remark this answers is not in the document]
jsanda: If it is something I (as in the user or within the operator) might be want to query on, then I often lean towards labels.

controllers/k8ssandra/datacenters.go
```diff
@@ -90,6 +101,22 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
+				// true and included in the resource hash. On subsequent reconciliations
+				// where the CassandraDatacenter exists rebuildNeeded will initially be set
+				// to false. If we get here, the label is present and thus the resource hash
+				// needs to be updated.
+				annotations.AddHashAnnotation(desiredDc)
+			}
+
 			if err = r.setStatusForDatacenter(kc, actualDc); err != nil {
 				logger.Error(err, "Failed to update status for datacenter")
 				return result.Error(err), actualDcs
```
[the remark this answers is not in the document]
jsanda: >A question about line 130 below: you are changing `desiredDc` but you aren't recomputing the hash annotation, is that ok?

I assume you are referring to this:

```go
if actualDc.Spec.SuperuserSecretName != desiredDc.Spec.SuperuserSecretName {
		// If actualDc is created with SuperuserSecretName, it can't be changed anymore. We should reject all changes coming from K8ssandraCluster
		desiredDc.Spec.SuperuserSecretName = actualDc.Spec.SuperuserSecretName
		err = fmt.Errorf("tried to update superuserSecretName in K8ssandraCluster")
		logger.Error(err, "SuperuserSecretName is immutable, reverting to existing value in CassandraDatacenter")
}
```
That's actually in main already. I agree that we should enforce that the field remains immutable.

apis/k8ssandra/v1alpha1/constants.go
jsanda: The name change would indeed provide more clarity. I'll do that. And yes you are correct that it plays no role when adding a DC to an existing cluster.

For what it's worth I am not at all happy about how we manage replication for system keyspaces, and I want to change it post-1.0. I'll try to summarize here briefly. 

We pass JVM system properties in an env var which is configured in the PodTemplateSpec. The management-api looks for those system properties. They are only applied once. Knowing this, I introduced a number of subtle, non-trivial changes to make sure those system properties don't change. Why? Because changing them changes the PodTemplateSpec which triggers a StatefulSet update, i.e., rolling restart. I want to manage replication for system keyspaces like we do for other ones. In order to do that though, we need to add support for running repairs, namely on `system_auth`. Since that has a small about of data, the management-api equivalent of `nodetool repair -full` should be fine.

pkg/cassandra/util.go
jsanda: 👍

pkg/cassandra/util.go
jsanda: No we can't do that. It would result in DCs added to an existing cluster to be included which leads to incorrect value for initial system replication. I

controllers/k8ssandra/datacenters.go
jsanda: Calling `cassandra.ApplySystemReplication(dcConfig, *systemReplication)` at line 57 for new DCs being added to an existing cluster does not result in the wrong system replication. Let's say we have a cluster with a single DC, dc1. System replication will be configured for it with these JVM system properties:

* `cassandra.system_distributed_replication_dc_names=dc1`
* `cassandra.system_distributed_replication_per_dc=3

Now let's say we after the cluster is fully initialized we add dc2. It will get those same system properties, which means we aren't replicating to dc2 yet. That's ok. As you point out, we could configure the system properties for dc2 so we have `cassandra.system_distributed_replication_dc_names=dc1,dc2`. I don't like this because now instead of managing one initial system replication for the whole cluster, we would have to manage it per DC.

We still have to call `updateReplicationOfSystemKeyspaces` on an initial DC as the replication could change. If we start with a single node, we will have RF = 1. If we later scale dc1 to 3 nodes, then system keyspaces will get updated to RF = 3 only when `updateReplicationOfSystemKeyspaces` is called.

This is confusing and messy. I think it will be a lot better if we do what I describe in https://github.com/k8ssandra/k8ssandra-operator/pull/262#discussion_r789200931.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
jsanda: Note that the label is going away as per your suggestion in https://github.com/k8ssandra/k8ssandra-operator/pull/262#discussion_r789134862.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
adutra: Yeah sorry, it's outside the scope of this PR, we can talk about this later.

apis/k8ssandra/v1alpha1/constants.go
adutra: > Because changing them changes the PodTemplateSpec which triggers a StatefulSet update, i.e., rolling restart.

Ah that bit explains a lot of the motivation here, thanks.

pkg/cassandra/util.go
adutra: Hmm but this function is never going to be called if you add a dc to an existing cluster, since the existing cluster resource would already have its `SystemReplicationAnnotation` set to a previous, initial value, right?

controllers/k8ssandra/datacenters.go
adutra: > It will get those same system properties, which means we aren't replicating to dc2 yet. That's ok. 

👍 

> We still have to call updateReplicationOfSystemKeyspaces on an initial DC as the replication could change.

👍

controllers/k8ssandra/datacenters.go
```diff
@@ -192,6 +211,137 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
+				return result.Error(err)
+			}
+			return result.RequeueSoon(15)
+		}
+		logger.Error(err, "Failed to get rebuild task", "Task", taskKey)
+		return result.Error(err)
+	}
+}
+
+func taskFinished(task *cassctlapi.CassandraTask) bool {
```
adutra, line 315: The cast to `int` in unnecessary.

controllers/k8ssandra/datacenters.go
```diff
@@ -192,6 +211,137 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
+		return result.Error(err)
+	}
+
+	desiredTask := newRebuildTask(dc.Name, dc.Namespace, srcDc)
+	taskKey := client.ObjectKey{Namespace: desiredTask.Namespace, Name: desiredTask.Name}
+	task := &cassctlapi.CassandraTask{}
+
+	if err := remoteClient.Get(ctx, taskKey, task); err == nil {
+		if taskFinished(task) {
+			// TODO what should we do if it failed?
```
adutra, line 288: Reschedule the same task?

controllers/k8ssandra/datacenters.go
```diff
@@ -192,6 +211,137 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
+			patch := client.MergeFromWithOptions(kc.DeepCopy())
+			delete(kc.Annotations, api.RebuildDcAnnotation)
+			if err = r.Client.Patch(ctx, kc, patch); err != nil {
+				err = fmt.Errorf("failed to remove %s annotation: %v", api.RebuildDcAnnotation, err)
+				return result.Error(err)
+			}
+			return result.Continue()
+		} else {
+			logger.Info("Waiting for datacenter rebuild to complete", "Task", taskKey)
+			return result.RequeueSoon(15)
```
adutra, line 299: Why not `r.DefaultDelay`?

controllers/k8ssandra/datacenters.go
```diff
@@ -192,6 +211,137 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
+		}
+	} else {
+		if errors.IsNotFound(err) {
+			logger.Info("Creating rebuild task", "Task", taskKey)
+			if err = remoteClient.Create(ctx, desiredTask); err != nil {
+				logger.Error(err, "Failed to create rebuild task", "Task", taskKey)
+				return result.Error(err)
+			}
+			return result.RequeueSoon(15)
```
adutra, line 308: Same here.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -137,13 +138,13 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
 	kcLogger.Info("All dcs reconciled")
 
-	if recResult := r.reconcileStargateAuthSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
-		return recResult.Output()
-	}
-
-	if recResult := r.reconcileReaperSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
-		return recResult.Output()
-	}
+	//if recResult := r.reconcileStargateAuthSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
```
adutra, line 141: This could be removed now it seems.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1441,15 +1489,40 @@ func parseResource(quantity string) *resource.Quantity {
...
-func (f fakeManagementApiFactory) NewManagementApiFacade(context.Context, *cassdcapi.CassandraDatacenter, client.Client, logr.Logger) (cassandra.ManagementApiFacade, error) {
-	m := new(mocks.ManagementApiFacade)
-	m.On("EnsureKeyspaceReplication", mock.Anything, mock.Anything).Return(nil)
-	m.On("ListTables", stargate.AuthKeyspace).Return([]string{"token"}, nil)
-	m.On("CreateTable", mock.MatchedBy(func(def *httphelper.TableDefinition) bool {
-		return def.KeyspaceName == stargate.AuthKeyspace && def.TableName == stargate.AuthTable
-	})).Return(nil)
-	return m, nil
-}
+//type ManagementApiFactoryAdapter func(
```
adutra, line 1492: This could be removed now it seems.

controllers/k8ssandra/replication.go
```diff
@@ -50,24 +78,209 @@ func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context,
...
+		if err != nil {
+			return nil, err
+		}
+		replication[k] = count
+	}
+
+	return replication, nil
+}
+
+func (r *K8ssandraClusterReconciler) versionCheck(ctx context.Context, kc *api.K8ssandraCluster) result.ReconcileResult {
```
adutra, line 272: Why do we need to do this now?

controllers/k8ssandra/datacenters.go
```diff
@@ -192,6 +211,137 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
+func getSourceDatacenterName(targetDc *cassdcapi.CassandraDatacenter, kc *api.K8ssandraCluster) (string, error) {
+	dcNames := make([]string, 0)
+
+	for _, dc := range kc.Spec.Cassandra.Datacenters {
+		if dcStatus, found := kc.Status.Datacenters[dc.Meta.Name]; found {
+			if dcStatus.Cassandra.GetConditionStatus(cassdcapi.DatacenterReady) == corev1.ConditionTrue {
+				dcNames = append(dcNames, dc.Meta.Name)
+			}
+		}
+	}
```
adutra, line 230:
```suggestion
	for _, template := range kc.GetReadyDatacenters() {
		dcNames = append(dcNames, template.Meta.Name)
	}
```

pkg/test/mgmtapi.go
```diff
@@ -0,0 +1,127 @@
...
+
+type FakeManagementApiFacade struct {
+	*mocks.ManagementApiFacade
+}
+
+func NewFakeManagementApiFacade() *FakeManagementApiFacade {
+	return &FakeManagementApiFacade{ManagementApiFacade: new(mocks.ManagementApiFacade)}
+}
+
+func (f *FakeManagementApiFacade) GetLastCall(method ManagementApiMethod, args ...interface{}) int {
```
adutra, line 83: Is this generated with Mockery? If so I'd suggest to:

* Move the generated code to the `mocks` package;
* Modify the `mocks` package README;
* Modify the `mocks` target in the Makefile.

pkg/test/testenv.go
```diff
@@ -268,7 +275,7 @@ func prepareCRDs() error {
 		return err
 	}
 
-	cassOperatorCrd := "github.com/k8ssandra/cass-operator/config/crd?ref=" + cassOperatorVersion
+	cassOperatorCrd := "github.com/burmanm/cass-operator/config/crd?ref=" + cassOperatorVersion
```
adutra, line 278: We need to remove this later, right?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -52,18 +47,28 @@ import (
...
 
 var (
-	defaultStorageClass = "default"
-	testEnv             *testutils.MultiClusterTestEnv
-	managementApi       = &fakeManagementApiFactory{}
+	defaultStorageClass  = "default"
+	testEnv              *testutils.MultiClusterTestEnv
+	managementApiFactory = &testutils.FakeManagementApiFactory{}
+
+	systemKeyspaces = []string{"system_auth", "system_distributed", "system_traces"}
```
adutra, line 61: Shouldn't we promote this to a global var in production code? Seems error prone to re-declare this list each time we need it.

controllers/k8ssandra/replication.go
```diff
@@ -50,24 +78,209 @@ func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context,
...
+
+// getReplicationForDeployedDcs gets the replication for only those DCs that have already
+// been deployed. The replication argument may include DCs that have not yet been deployed.
+func getReplicationForDeployedDcs(kc *api.K8ssandraCluster, replication *cassandra.Replication) *cassandra.Replication {
+	dcNames := make([]string, 0)
+	for _, dc := range kc.Spec.Cassandra.Datacenters {
+		if status, found := kc.Status.Datacenters[dc.Meta.Name]; found && status.Cassandra.GetConditionStatus(cassdcapi.DatacenterInitialized) == corev1.ConditionTrue {
+			dcNames = append(dcNames, dc.Meta.Name)
+		}
+	}
```
adutra, line 234:
```suggestion
	for _, template := range kc.GetReadyDatacenters() {
		dcNames = append(dcNames, template.Meta.Name)
	}
```

pkg/cassandra/datacenter.go
```diff
@@ -34,6 +34,51 @@ type SystemReplication struct {
...
+		for _, ks := range keyspaces {
+			if _, found := ksMap[ks]; !found {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func (r *Replication) ForDcs(dcs ...string) *Replication {
```
adutra, line 57: It would be nice to have some unit tests for this method.

pkg/cassandra/management.go
```diff
@@ -290,3 +295,20 @@ func (r *defaultManagementApiFacade) EnsureKeyspaceReplication(keyspaceName stri
...
+
+func (r *defaultManagementApiFacade) GetSchemaVersions() (map[string][]string, error) {
+	pods, err := r.fetchDatacenterPods()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, pod := range pods {
+		if schemaVersions, err := r.nodeMgmtClient.CallSchemaVersionsEndpoint(&pod); err != nil {
+			r.logger.V(4).Error(err, "failed to list schema versions", "Pod", pod.Name)
```
adutra, line 307: Why do we need `r.logger.V(4)` here?

adutra (changes requested):
Second round of remarks :-)

controllers/k8ssandra/k8ssandracluster_controller_test.go
adutra: I see similar slices in:
* controllers/k8ssandra/replication.go:90
* controllers/k8ssandra/replication.go:203

pkg/test/mgmtapi.go
```diff
@@ -0,0 +1,127 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ManagementApiFactoryAdapter func(
+	ctx context.Context,
+	datacenter *cassdcapi.CassandraDatacenter,
+	client client.Client,
+	logger logr.Logger) (cassandra.ManagementApiFacade, error)
+
+var defaultAdapater ManagementApiFactoryAdapter = func(
```
adutra, line 21: Nit: typo in the name, should be `defaultAdapter`.

controllers/k8ssandra/datacenters.go
```diff
@@ -143,10 +135,20 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 
 			actualDcs = append(actualDcs, actualDc)
 
-			if recResult := r.updateReplicationOfSystemKeyspaces(ctx, kc, desiredDc, remoteClient, logger); recResult.Completed() {
+			if recResult := r.checkSchemas(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
 				return recResult, actualDcs
 			}
 
+			if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) {
+				//if recResult := r.updateUserKeyspacesReplication(ctx, kc, desiredDc, mgmtApi, logger); recResult.Completed() {
```
adutra, line 143: This seems to have moved to `checkSchemas`.

pkg/test/testenv.go
jsanda: I missed that one. I need to update that to point to upstream master.

controllers/k8ssandra/k8ssandracluster_controller_test.go
jsanda: Actually, a global var makes me nervous that it would be error prone since the slice isn't immutable. How about instead a util function that always returns a new copy?

controllers/k8ssandra/replication.go
jsanda: Once we add support for stopping a DC (#296) it would be wrong to check only the Ready condition. We will need to check the Initialized condition. How about I add a `GetInitializedDatacenters` method?

controllers/k8ssandra/k8ssandracluster_controller_test.go
adutra: That's a very interesting remark. 

We considered this in go-cassandra-native-protocol a while ago. This project exports a lot of global vars, and we were concerned by the fact that these variables were mutable. After some research we came to the conclusion that we shouldn't care about this. If a consumer of your package alters any exported var, it's their fault. The runtime packages themselves export a bunch of mutable variables. If you want to break almost every application out there, simply do this:

```go
import "time"
func init() {
	time.UTC = nil
}
```

That said, I don't mind using the function approach either, if you prefer to err on the safe side.

controllers/k8ssandra/replication.go
adutra: Ah right. Yes, that works for me.

pkg/test/mgmtapi.go
adutra: Hmm after reading this again, I think you wrote this code on top of the generated mock. Please disregard this comment.

controllers/k8ssandra/k8ssandracluster_controller_test.go
jsanda: I'm good with global var :)

pkg/cassandra/management.go
jsanda: We don't; however, I think we should be setting verbosity most of the time. See #282.

pkg/cassandra/util.go
jsanda: That is correct. The function is called from `K8ssandraClusterReconciler.checkInitialSystemReplication()`. The first thing it does is check for the annotation. If the annotation exists and can be unmarshalled, the function returns so `ComputeInitialSystemReplication()` and `GetDatacentersForSystemReplication()` won't get called.

controllers/k8ssandra/replication.go
```diff
@@ -50,24 +78,209 @@ func (r *K8ssandraClusterReconciler) checkSystemReplication(ctx context.Context,
...
+	return result.Continue()
+}
+
+func getUserKeyspaces(mgmtApi cassandra.ManagementApiFacade, kc *api.K8ssandraCluster) ([]string, error) {
+	keyspaces, err := mgmtApi.ListKeyspaces("")
+	if err != nil {
+		return nil, err
+	}
+
+	internalKeyspaces := getInternalKeyspaces(kc)
```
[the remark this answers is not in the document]
adutra: Whoops sorry about `system_peers`. 

But my point was: shouldn't we get bold and evict all keyspaces starting with `system`? More keyspaces could be added in the future, and if they are, they will show up as user keyspaces and could make the dc addition fail. 

For example, C* 4.0 also has a `system_virtual_schema` keyspace. This ks shows up when I do a regular `describe keyspaces` command in CQLSH. How is this keyspace going to be handled here, if it is returned by `ListKeyspaces`?

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -137,13 +138,13 @@ func (r *K8ssandraClusterReconciler) reconcile(ctx context.Context, kc *api.K8ss
...
+	//	return recResult.Output()
+	//}
+	//
+	//if recResult := r.reconcileReaperSchema(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
+	//	return recResult.Output()
+	//}
 
 	if recResult := r.reconcileStargateAndReaper(ctx, kc, actualDcs, kcLogger); recResult.Completed() {
```
adutra, line 149: The method name is now a bit misleading: it also reconciles telemetry. Just FYI.

jsanda: I am renaming to `afterCassandraReconciled`

controllers/k8ssandra/replication.go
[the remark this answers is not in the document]
jsanda: Thinking about this a bit more. What if a user-defined keyspace starts with `system`? We would incorrectly filter that out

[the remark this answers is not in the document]
adutra: > What if a user-defined keyspace starts with system?

They would be asking for trouble :-) For instance, the Java driver would also filter such keyspace out of token range computation by default.

But OK to leave it as is. We can certainly revisit later.

adutra (approved):

adutra:
Approving this as the e2e failures seem unrelated. A few minor questions are still inflight but can be revisited later. It would be good to have this merged asap.

PR #273, opened 2022-01-06
Revisit cluster name and derived resource names [K8SSAND-1125]

### LINKED ISSUE #267: K8SSAND-1125 ⁃ Revisit cluster name and derived resource names
We currently have a field `CassandraClusterTemplate.Cluster` that is used to specify the Cassandra cluster name. This field is also used to create derived resource names, such as secrets, services and deployments.

But this field keeps being constantly mixed up with the `K8ssandraCluster` resource name. For example the superuser secret uses the Cassandra cluster name (because that's what cass-operator does), but Reaper secrets use the `K8ssandraCluster` name. Stargate deployments use the Cassandra cluster name, but Reaper deployments use the `K8ssandraCluster` name.

This has not been detected in tests because our tests always create `K8ssandraCluster` resources whose names are equal to the Cassandra cluster name. This confusion is present e.g. in the `RetrieveSuperuserSecret` method, where the superuser secret (created with the Cassandra cluster name) is looked up by the `K8ssandraCluster` name.

Lastly, sanitization of Cassandra cluster names is not being done optimally:

* The function `DefaultSuperuserSecretName` does a basic sanitization of the Cassandra cluster names, but the logic is not really compliant with valid DNS subdomain names as defined in RFC 1123.
* Stargate deployments are currently not sanitizing their names at all, yet the names are based on Cassandra cluster names.

Here is what I suggest:

* Remove `CassandraClusterTemplate.Cluster` completely, use `K8ssandraCluster` name consistently instead.
* CRD names must be valid DNS subdomain names, which means that, by using the `K8ssandraCluster` name, we don't need sanitizing anymore.
* If we still want to sanitize names to remove more unwanted chars (e.g. `-` is perfectly valid, but we might want to disallow it for aesthetic reasons), then create a dedicated function for that and use it consistently for creating secret, service and deployment names.
* K8ssandra Operator version: 1.0.0.-alpha3

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1125) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1125
┆priority: Medium

adutra:
**What this PR does**:
This commit removes CassandraClusterTemplate.Cluster
completely, and uses K8ssandraCluster name consistently
instead.

CRD names must be valid DNS subdomain names, which means
that, by using the K8ssandraCluster name, we don't need
to sanitize derived resource names anymore. So this
commit also removes the sanitizing function used to
create the superuser secret.

Fixes #267 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/secret/replicated.go
+++ b/pkg/secret/replicated.go
@@ -7,17 +7,15 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/annotations"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
-	"math/big"
-	"reflect"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"strings"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"math/big"
+	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	replicationapi "github.com/k8ssandra/k8ssandra-operator/apis/replication/v1alpha1"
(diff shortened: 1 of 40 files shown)


adutra:
This is now ready for review. The test failures are unrelated.

pkg/secret/replicated.go
```diff
@@ -46,25 +44,10 @@ func generateRandomString(charset string, length int) ([]byte, error) {
...
-// <cluster_name>-superuser.
+// <cluster_name>-superuser. Note that we expect clusterName to be a valid DNS subdomain name as defined in RFC 1123,
+// otherwise the secret name won't be valid.
 func DefaultSuperuserSecretName(clusterName string) string {
-	return fmt.Sprintf("%s-superuser", SanitizeSecretName(clusterName))
-}
-
-// SanitizeSecretName returns a string suitable for creating secret names derived from the given cluster name.
-// Secret names must be a valid DNS subdomain name as defined in RFC 1123.
-func SanitizeSecretName(name string) string {
```
adutra, line 56: I don't think we need this anymore. The cluster name being the CRD name it has already been validated. If you try to create a k8c with an invalid name, you'd get:

```
The K8ssandraCluster "cluster1_fancy_name" is invalid: metadata.name: Invalid value: "cluster1_fancy_name": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

adejanovski (approved):
Just a minor comment on the changelog, otherwise it's all good 👍 
It'll make our code much safer, thanks for coming up with this!

Maybe you should rebase before merging, because a few things were merged into `main` since you pushed that PR.

PR #286, opened 2022-01-11
Fix cluster scoped kustomizations + update e2e tests + add custom namespace kuttl test [K8SSAND-1137, K8SSAND-1159]

### LINKED ISSUE #278: K8SSAND-1137 ⁃ Test custom namespace deployment
Ensure that when installing a cluster-scoped control plane using kustomize, a k8ssandra cluster can be deployed in any namespace in the cluster and reach a healthy state.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1137) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1137
┆priority: Medium

### LINKED ISSUE #285: K8SSAND-1159 ⁃ Cluster scoped deployments currently fail
There are some apparent issues when attempting to deploy a basic cluster-scoped k8ssandra operator deployment, that we need to tackle ahead of getting a test working. 

When running `kustomize build config/deployments/control-plane/cluster-scope` we see the following errors:

```java
Error from server (NotFound): namespaces "cass-operator" not found

Error from server (Invalid): ClusterRoleBinding.rbac.authorization.k8s.io "cass-operator-manager-rolebinding" is invalid: subjects[0].namespace: Required value
```

The problems appear to occur due to the dependancy on `config/cass-operator-cluster-scope.`

I think we can eliminate some of these issues by using the cluster scoping kustomizations in `cass-operator` rather than trying to cluster scope the `cass-operator` deployment from the `k8ssandra-operator` kustomizations.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1159) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1159
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Cluster scoped deployments currently fail due to a combination of factors:

* Flaky kustomizations which aim to deploy a cluster-scoped `cass-operator` in the `k8ssandra-operator` repository. There are working cluster-scoped deployments in the `cass-operator` repo which we adopt in this PR.
* Namespace creation for the `k8ssandra-operator` is set up as a component - this causes failures when the namespace is used as a source in replacements for `ClusterRoleBindings` which are created when cluster-scoping `k8ssandra-operator`. We are moving namespace creation into `config/default` (i.e. as a kustomization, not a component) because - given components applied last in the transformation pipeline - it otherwise can't be a source for the replacements in `ClusterRoleBinding`'s subjects.
* We are moving to always use a cluster-scoped cass-operator, because there is no reason to namespace scope it and this causes additional complexity.
* The use of a `namespace: ...` transformation throughout the kustomizations and e2e tests is undesirable, because it errors when deploying cass-operator into a `cass-operator` NS due to namespace transformation conflicts. A more targeted approach is adopted, whereby we patch the namespace created in `config/default` and then use it as a replacement for `metadata.namespace` for all resources created in the `k8ssandra-operator` ns only.

With these issues fixed, the E2E tests needed to be updated to reflect the new structure. In particular:

* We remove the use of the namespace transformer in all templates. 
* We remove the use of `kustomize edit namespace` type calls in all e2e tests.
* We remove the use of the `--namespace` flag in all `kubectl apply` operations scripted from the e2e tests.
* We centralise instead on the patch -> replace process described above to create e2e test namespaces and deploy resources to them.

Fixes #285, #278

**Checklist**
- [x] Changes manually tested


Miles-Garnsey:
**Tests currently failing**
* `Run e2e tests / CreateSingleMedusa` _(Also fails on `main`)_
* `Run multi-cluster e2e tests / CreateMultiMedusa` _(Also fails on `main`)_
* `Run multi-cluster e2e tests / CheckStargateApisWithMultiDcCluster` _(Also fails on `main`)_

From additional testing on `main` I can confirm that `CreateSingleMedusa` fails there too, and so does `CheckStargateApisWithMultiDcCluster`. 

I am marking this ready for review as a result.

Miles-Garnsey:
> I have briefly started looking at the changes. Awesome that you have a solution for the namespace problem.

It has been a delicate operation but it does work.

> We definitely do not want to always use a cluster-scoped cass-operator deployment. We need to support deploying both namespace-scoped and cluster-scoped.

We do support deploying a cluster scoped cass-operator, but the user needs to write their own kustomizations to deploy cass-operator first and then k8ssandra operator (having removed the cass-operator dependancy in the k8ssandra-operator kustomizations).

If you'd like a specific test or kustomization written for this use case please let me know how you'd like it to look and where you'd like it to appear.

Miles-Garnsey:
> This isn't acceptable approach. Most cass-operator users from my reports are using namespace scoped deployments, not cluster scoped. In cass-operator, all versions <= 1.7.1 are only shipped with namespace scoped deployment model, the user had to manually modify them to get cluster-scoped version.

> Changing that approach upside-down would probably make some eyes roll.

@burmanm John and I have discussed this and I've agreed to reinstate the namespace scoping of cass-operator for the tests.

I usually advise folks against namespace scoping an operator unless they have a very well thought through rationale to justify doing so. But I would rather get this PR merged - so that at least users CAN cluster-scope k8ssandra operator - than waste time debating whether namespace or cluster scope should be the default.

@jsanda the changes you've requested have been made. As we've discussed offline, the `ClusterScoped/MultiDcMultiCluster` fails for a variety of reasons, not least `Error: Artifact name is not valid: k8s-logs-ClusterScoped/MultiDcMultiCluster. Contains the following character:  Forward slash /`.

I have also had to upgrade the kustomize version in use and remove the use of `go-get-tool` to build kustomize from source (in favour of downloading the binary artefact) because `go-get-tool` refuses to work.

Please let me know if you are now happy to approve this, I think that the tests should now pass (barring `ClusterScoped/MultiDcMultiCluster` and `MultiDcAuthOnOff` which I think are known-broken).

Miles-Garnsey:
Oop, looks like `MultiDcAuthOnOff` does work now. Magic. So the only test failing is `ClusterScoped/MultiDcMultiCluster`.

I've added in another kuttl test to confirm that when k8ssandra-operator and cass-operator are cluster-scoped (and installed in namespaces k8ssandra-operator and cass-operator) a K8ssandraCluster deployment to a third NS will result in the expected StatefulSet being deployed.

scripts/setup-kind-multicluster.sh
```diff
@@ -153,7 +153,7 @@ function deploy_cass_operator() {
 
   for ((i=0; i<$num_clusters; i++))
   do
-    kustomize build config/cass-operator | kubectl --context kind-k8ssandra-$i apply -f -
+    kustomize build config/cass-operator/cluster-scoped | kubectl --context kind-k8ssandra-$i apply -f -
```
jsanda, line 156: The `deploy_cass_operator` function isn't used since the script doesn't deploy cass-operator. It did in an earlier version though. Can you remove the function?

Miles-Garnsey: Confirming my understanding here: `deploy_cass_operator` is a function which is no longer called anywhere in the codebase and needs to be removed?

Miles-Garnsey: I was just updating it so that we had consistency, but if it is no longer used it makes sense to get rid of it.

Miles-Garnsey: John and I have spoken offline and this function should be deleted.

Miles-Garnsey: I've implemented this change. Was there anything else required to progress this @jsanda?

Miles-Garnsey:
Oh right. Now we have merge conflicts. I'll look into that.

jsanda:
@burmanm can you please pick up the review on this? Miles has been done with it. I just haven't been able to get back to it.

Miles-Garnsey:
@jsanda I'm not sure if this will be done today now. 

That rebase was a nightmare, but most tests are now working. However for some reason I now see failures on `CreateMultiReaper` to the effect of:

```
Messages:   	secret multi-dc-reaper-t7dwpb/reaper-cql-secret is not managed by test
```

This is the first I've seen of that error, so it must be due to a problem I've introduced while rebasing.

There are a few other areas where some other tests are failing. I'm scratching my head about some of them as they should not have been affected by any of the work I've done. I will continue to investigate.

Makefile
```diff
@@ -265,7 +265,12 @@ controller-gen: ## Download controller-gen locally if necessary.
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.0.5)
+	mkdir -p $$(pwd)/bin
```
burmanm, line 268: What's the point of these changes? Instead of single function that downloads any of the tools, we have to create separate process for each?

config/cass-operator/cluster-scoped/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=6a78a8237173d9322e6c0cab94c615b1f043a906
+
+images:
+  - name: k8ssandra/cass-operator
+    newTag: 6a78a823
```
burmanm, line 9: Do we really have a need to redefine the deployed version?

test/framework/e2e_framework.go
```diff
@@ -134,8 +134,10 @@ func generateCassOperatorKustomization(namespace string) error {
 kind: Kustomization
 
 resources:
-- github.com/k8ssandra/cass-operator/config/default?ref=v1.9.0
-namespace: {{ .Namespace }}
+- github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=6a78a8237173d9322e6c0cab94c615b1f043a906
```
burmanm, line 137: This isn't something to fix in this PR, but shouldn't we make these versions configured to just one place? Like the Makefile and when you run the make <any-test-scenario>, it would then modify these Kustomization files? I know kustomize doesn't allow remote-resource refs to be set on the higher level, but the way we now define and configure these is quite bad.

burmanm: And I guess this applies to the whole `generateCassOperatorKustomization` function. E2E tests shouldn't have a difference in how they deploy these components compared to what we provide to our users. While we hope to maintain that they don't differ or that these patches are not necessary - our e2e tests can't actually verify that's the case. In worst case, the namespace selectors here fix a bug that was introduced in the real kustomization rules (it wouldn't be the first time kustomization rules behave unlike we expected them to)

test/framework/e2e_framework.go
```diff
@@ -274,10 +330,10 @@ func (f *E2eFramework) kustomizeAndApply(dir, namespace string, contexts ...stri
 		return err
 	}
 
-	if err := kustomize.SetNamespace(kdir, namespace); err != nil {
-		f.logger.Error(err, "failed to set namespace for kustomization directory", "dir", kdir)
-		return err
-	}
+	// if err := kustomize.SetNamespace(kdir, namespace); err != nil {
```
burmanm, line 333: Dead code.

test/kuttl/test-user-defined-ns/00-cert-manager.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands: 
+- command: kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.0/cert-manager.yaml
```
burmanm, line 4: This version differs from the one we deploy otherwise (1.5.3). Also, 1.6.0 introduced an upgrade regression which 1.6.1 fixes.

burmanm: Note: there's no check that verifies that cert-manager was started and correctly deployed. This causes errors in faster machines and works on slower machines. The same issue plagues cass-operator a bit, on my Linux workstation the execution of Kube commands is so fast that the integration-tests fail since the webhook wasn't enough quick to initialise, but the tests work fine in GHA and my Mac since they're slower. You could run into the same issue here (unless kettle is so slow that it can't).

test/framework/e2e_framework.go
jsanda: >This isn't something to fix in this PR, but shouldn't we make these versions configured to just one place?

Agree with both points. Definitely an area for improvement and for a different PR.

`generateCassOperatorKustomization` is no longer used. I will only open a separate PR to remove it. `generateK8ssandraOperatorKustomization` is used in the same way that users would. It simply adds another kustomization layer for setting the namespace.

jsanda:
>E2E tests shouldn't have a difference in how they deploy these components compared to what we provide to our users. While we hope to maintain that they don't differ or that these patches are not necessary - our e2e tests can't actually verify that's the case.

When I first read this I hadn't looked at all of the latest changes. This led to some confusion on my part because I thought we are already doing this. Then I saw the changes in `e2e_framework.go`  that use kustomize patches to change the namespace . I agree with the above comment, and I also agree that in general that patches aren't generally how users will configure the namespace. They will use a namespace transform. We actually have an example for that in the install guide.

My confusion led to a follow up conversation with @burmanm. He suggested that it would be better to deploy both k8ssandra-operator and cass-operator in the same namespace. After some discussion I do agree for a couple reasons. First, it's easier for the user if we can deploy everything, or as much as possible (cert-manager is currently an exception), to the same namespace. cass-operator is an implementation detail. There are k8ssandra users who are unaware that there is this cass-operator component that is driving things. When I think about from that perspective, installing both operators in the same namespace makes a lot of sense. Secondly, for a namespace-scoped install we already install both of them into the same namespace. Installing both operators into the same namespace for a cluster-scoped deployment keeps things more consistent which is another plus.

With all that in mind I would like to do the following:

* Change the kustomize code in `e2e_framework.go` back to use the namespace transform as it did previously
* Revert back to keep the kustomize components
    * As the k8s [docs](https://kubectl.docs.kubernetes.io/guides/config_management/components/) mention components are useful for adding reusable pieces of configuration when you have optional features.
* Change the kustomizations so that are not trying to deploy anything into the cass-operator namespace 

@Miles-Garnsey I realize that this will result in some undo/redo/undo work. I apologize for that, but I think it is the best path forward.

Makefile
Miles-Garnsey: The purpose is that go-get-tools fails to build kustomize 4.4.0 so I've changed it so that we download a binary artefact instead. That is all clearly explained in this [comment](https://github.com/k8ssandra/k8ssandra-operator/pull/286#issuecomment-1014115092), please let me know if you have quesitons.

config/cass-operator/cluster-scoped/kustomization.yaml
Miles-Garnsey: @jsanda this was your change that I merged in. 

@burmanm I believe this has something to do with CassandraTask causing test failures due to CRD versions needing to line up with cass-operator versions, or something... I'd rather keep that discussion elsewhere though as it isn't related to this PR.

test/framework/e2e_framework.go
Miles-Garnsey: I don't quite follow this discussion, but if it isn't something to fix in this PR I think I'm safe to mark these comments as resolved?

test/kuttl/test-user-defined-ns/00-cert-manager.yaml
Miles-Garnsey: It doesn't hurt to test multiple cert-manager versions IMO. I've never seen this test flake but if you find it does I'm happy to hear about it. 

> Note: there's no check that verifies that cert-manager was started and correctly deployed

Appreciate the tip here (timing issues are hugely painful and it is great that we're all looking out for them.) In this case `00-assert.yaml` actually does ensure that all cert-manager components are successfully deployed and available prior to proceeding to the second step.

test/framework/e2e_framework.go
Miles-Garnsey: Fair enough, removed.

Miles-Garnsey:
Addressing some comments here:

> E2E tests shouldn't have a difference in how they deploy these components compared to what we provide to our users. While we hope to maintain that they don't differ or that these patches are not necessary - our e2e tests can't actually verify that's the case.

@burmanm I do not understand this comment. The e2e tests are explicitly designed to randomise the namespace (presumably so they can be run in parallel?) so they **have** to differ from how the user deploys things. If anything, I've reduced the discrepancies between test/real deployments here by just adjusting the ns in one place instead of adjusting it in 3 different places as the original code did.

Please clarify.

> I agree with the above comment, and I also agree that in general that patches aren't generally how users will configure the namespace. They will use a namespace transform. We actually have an example for that in the install guide.

@jsanda I've explicitly dealt with this issue in the [original ticket](https://k8ssandra.atlassian.net/browse/K8SSAND-1159?focusedCommentId=12333). 

I think the ticket addresses a lot of the ideas currently under discussion and explains why they are not tenable. But I will address your suggestions again here and link to my previous explanations.

### Namespace transforms

If users attempt to use a NS transform (`namespace:...`) in the kustomize.yaml file as you propose, I believe they will get the same error that our own kustomizations are currently [generating](https://k8ssandra.atlassian.net/browse/K8SSAND-1159?focusedCommentId=12333). I do not believe that this is an effective way to change the namespace where the upstream kustomizations have multiple namespaces, and if our install guide says otherwise, I suspect it is wrong. 

I have also discussed this in the PR description [here](https://github.com/k8ssandra/k8ssandra-operator/pull/286#issue-1098656914).

Please let me know if I'm missing something here, but I do not believe that this proposal will fly.

### Kustomize components

> Revert back to keep the kustomize components

Again, as discussed [here](https://k8ssandra.atlassian.net/browse/K8SSAND-1159?focusedCommentId=12333), this will not work due to the order of transformation applied in the kustomize transformation pipeline.

> As the k8s docs mention components are useful for adding reusable pieces of configuration when you have optional features.

I'm aware of the intended use of components, I believe I first [recommended](https://datastax.slack.com/archives/C02023LQC1E/p1632377796270200?thread_ts=1632371543.261800&cid=C02023LQC1E) (private link, sorry to anyone who doesn't have access) using them. I recommend reading the detailed [KEP](https://github.com/arrikto/kubernetes-enhancements/blob/master/keps/sig-cli/1802-kustomize-components.md) for components in order to understand why using them in the current fashion will break the subject replacement on ClusterRoleBindings as discussed in the original ticket.

> Change the kustomizations so that are not trying to deploy anything into the cass-operator namespace

This may be possible, but will be limiting to users who want to deploy cass-operator and k8ssandra-operator into different namespaces, as it is similar to removing a label. I don't recommend it and would like to address it in a separate PR, this one is becoming a magnet for off-topic discussion and I do not think enlarging it again is productive or helpful.

Makefile
burmanm: Works fine on my machines at least and cass-operator uses that method to download 4.4.1 for every integration test, so it works on GHA also. What in it doesn't work?

burmanm:
The exact problem is that you don't use namespace transformer. That's how we would tell our users to do it, that's how kustomize documentation says to use it. Not some replacements strategy, which works quite differently to what namespace transformer does. 

Here we potentially mask an issue where user can't deploy the k8ssandra-operator since kustomize doesn't work correctly, yet our e2e tests pass fine - because they fix an issue. Especially if you just said "it doesn't work", then that's a crappy story. 

And as said, we shouldn't use more than one namespace if possible. It's a hindrance to users not knowing how many namespaces they need to potentially request access to or why they would need more than one (why would they? I can't come up with any sane reason why it makes sense).

jdonenine:
So let me propose a requirement and we can talk through the solution with that in mind. 

Quite frankly we should have had this conversation sooner, and I'm sorry we didn't. This one snuck around the grooming discussions because I didn't realize it was bigger than a quick bug fix, we'll need to get better at that. 

We want the operators to all be deployed to the same namespace and we do not need to support a scenario now where they can deployed in any other fashion. 

How does that simplify the solution?

We need progress here, so let's looks for that simplification. With that definition what do you all think @jsanda @Miles-Garnsey @burmanm ?

jsanda:
If we deploy the operators to the same namespace, I think it could be as simple as [this](https://github.com/jsanda/k8ssandra-operator/commit/497cea53841b11bdd5349dbd979493576b9b4714).

Makefile
Miles-Garnsey: I was getting an error when go-get-tools tried to build kustomize from source, I resolved it back on the [17th](https://datastax.slack.com/archives/C02023LQC1E/p1642385392477800) and the full [logs](https://github.com/k8ssandra/k8ssandra-operator/runs/4835404880?check_suite_focus=true) from those test runs are now deleted. The two fragments of the error I'd copied into involved  `undefined: os.ReadDir` and `note: module requires Go 1.16`.

Miles-Garnsey:
> If we deploy the operators to the same namespace, I think it could be as simple as [this](https://github.com/jsanda/k8ssandra-operator/commit/497cea53841b11bdd5349dbd979493576b9b4714).

As covered in the ticket and earlier comments, this solution produces the following error:

```
$ kustomize build config/deployments/control-plane/cluster-scope | kubectl apply -f -
Error from server (Invalid): error when creating "STDIN": ClusterRoleBinding.rbac.authorization.k8s.io "manager-rolebinding" is invalid: subjects[0].namespace: Required value
```

Miles-Garnsey:
> We want the operators to all be deployed to the same namespace and we do not need to support a scenario now where they can deployed in any other fashion.

@jdonenine thanks for clarifying that, it certainly would have been good to know at the outset. 

The requirement on these tickets were:

1. Write a test to deploy a k8ssandraCluster into a namespace outside the operator's namespace.
2. (When it became clear this ^^ functionality was broken) Fix the cluster-scoped kustomizations.

Looking at the current deployments we have the following:

```
$kustomize build config/deployments/control-plane/cluster-scope | grep 'namespace: '
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: system
  namespace: k8ssandra-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: cass-operator
  namespace: k8ssandra-operator
  namespace: cass-operator
  namespace: cass-operator
      namespace: cass-operator
```

We can see that currently, the cluster-scoped control-plane deployment deploys cass-operator into the cass-operator namespace. If this was something we wanted to change, that change was not in the original requirements.

I'm happy to look at that topic as a separate ticket, but I am keen to get this merged, since expanding scope at this point (when we don't have consensus on show stopping issues like [this](https://github.com/k8ssandra/k8ssandra-operator/pull/286#issuecomment-1023681699)) seems like it risks wasting a lot of effort.

jdonenine:
When we talked yesterday we were under the impression that in a namespace-scoped deployment everything went into k8ssandra-operator, it was that behavior that we wanted to be consistent with.

I guess it's safe to assume that's correct?  What you mentioned was for a cluster-scoped deployment, which right now we would assume to be a mixture.  But our working assumption was in that case that it could be pretty easily made to work in that fashion.  

You mentioned the error that you get with the quick example @jsanda threw into a previous comment, obviously there's something missing from what he had there, but I don't know that it was meant to be a fully baked solution or not - I don't know that that error means that it can't be as minimal a change as proposed there or not I guess.

Miles-Garnsey:
> When we talked yesterday we were under the impression that in a namespace-scoped deployment everything went into k8ssandra-operator,

This is correct, in the existing (broken) cluster-scoped kustomizations, cass-operator goes into the cass-operator namespace. 

> I guess it's safe to assume that's correct? What you mentioned was for a cluster-scoped deployment, which right now we would assume to be a mixture. But our working assumption was in that case that it could be pretty easily made to work in that fashion.

> You mentioned the error that you get with the quick example @jsanda threw into a previous comment, obviously there's something missing from what he had there, but I don't know that it was meant to be a fully baked solution or not - I don't know that that error means that it can't be as minimal a change as proposed there or not I guess.

I really want to move the conversation on that topic elsewhere because this PR has taken ~3 days of rebasing time already and is in its 4th major revision. I really advise against adding more to the scope right now given there are disagreements on other topics (e.g whether the `namespace:` transformation should even be used).

I don't know what the intention of that code sample was, I can get into detail explaining that when you are using the `namespace: ` transform you need to take special care over the `ClusterRolebinding`'s subject namespace because it sometimes isn't correctly patched. This applies even moreso when you are patching it from a  RoleBinding to a ClusterRoleBinding. 

I don't know if I'd describe any of that as simple, or that it `could be pretty easily made to work in that fashion` when there's so much contention over other issues, to be completely honest.

I really advocate solving the issue as scoped before we start bringing in additional questions like whether we want to move all resources in `k8ssandra-operator` because there are just too many unrelated conversations happening at once on this PR, and I am starting to think it needs to be closed, re-scoped, and started from scratch.

Makefile
burmanm: Are you trying to build with something older than go 1.16?

burmanm: Because both of those errors look like it, os.ReadDir was introduced in 1.16 and that module would require go 1.16 it seems.

Miles-Garnsey: No, this error comes up when you try to upgrade kustomize to 4.4.0 and up. It doesn't happen locally, I only saw it on GHA, but it was consistently occurring.

Try installing kustomize 4.4.0 using go-get-tools on GHA and you'll see what I'm talking about. I'd be curious if you have any thoughts as to what's going on there, because it didn't make sense to me.

Miles-Garnsey: Whatever version of Go I was trying to build with was simply the same one that GHA always uses (which I assume is a 1.16).

burmanm: As I said, this same method installs kustomize 4.4.1 in every GHA run on the cass-operator side and it works fine. And that method works fine in our other tests also. The reason you see that error is because there's wrong go version installed in the kuttl_tests.

Miles-Garnsey: Michael, if I include this step, do you think the go-get-tools install will work? 

```
      - name: Set up Go 1.16
        uses: actions/setup-go@v1
        with:
          go-version: 1.16
      - uses: actions/cache@v2
```

burmanm: See PR #341

test/kuttl/test-user-defined-ns/02-k8ssandra-cluster.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands: 
+- command:  kubectl create ns test-ns
+  ignoreFailure: false 
+- command:  kubectl apply -n test-ns --server-side --force-conflicts -f ../../testdata/samples/k8ssandra-single-kind.yaml
```
burmanm, line 6: Isn't this silly to test for failures if we intentionally apply with "do not care about failures, overwrite my mistakes" ? The --force-conflicts shouldn't probably be here since it masks issues.

Miles-Garnsey: I thought `--force-conflicts` had something to do with server side apply and ensuring that even if some other controller was updating the resource, our mutation would still be applied? 

What issues does this mask? I don't follow your question I'm sorry.

Miles-Garnsey:
I think this rebase is now working. @jsanda if you want to take a look at whether the tests are passing in your morning, we might be lucky enough that this is ready for approval.

Miles-Garnsey:
FYI, near as I can tell, the broken test was because the kustomizations that `generateCassOperatorKustomization ` in `e2e_framework.go` generates used to move cass-operator into the test's namespace, but they now adhere more closely to what happens when the user deploys (which was, and remains, deployment to `cass-operator` ns).  

That did necessitate a small change in the e2e test `BeforeTest()` function though so that it knows not to look in the test namespace for cass-operator in cluster scoped tests.

We'll probably revert all of that back out eventually, as we move to single namespace deployment (tracked [here](https://k8ssandra.atlassian.net/browse/K8SSAND-1208)) but I like passing tests.

jsanda (approved):
I check the logs on the failing test and it is due to open file limit issue in management-api that we have been hitting recently. The PR is good to go. Thanks @Miles-Garnsey!

PR #287, opened 2022-01-11
Include all Cassandra yaml settings from 3.11, 4.0 and trunk in the CRD [K8SSAND-1098]

### LINKED ISSUE #249: K8SSAND-1098 ⁃ Structured cassandra.yaml configuration
Provide a more structured method for the configuration of high-value cassandra.yaml configuration properties.  This should work in conjunction with the more flexible unstructured ConfigMap pass through methodology also to be implemented.

Will need to identify the properties to be covered through this effort.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-1098) by [Unito](https://www.unito.io)
┆epic: Cassandra Configuration
┆friendlyId: K8SSAND-1098
┆priority: Medium

adejanovski:
**What this PR does**:
Declares all Cassandra config settings from 3.11 to trunk in a strongly typed fashion in the CRD.
This will allow safer configuration for Cassandra while supporting all settings that aren't automatically handled by K8ssandra (such as seeds or data directories for example).

Fixes #249

**Checklist**
- [ ] Changes manually tested


burmanm:
The code that generates all of these settings should be in a script that could be run from the Makefile. Not "google sheet here", "do manually this part". That makes it unmaintainable.

jdonenine:
While fully automating everything for rerunning later is definitely a good goal to get to, we can certainly live with the current partiality manually generated state for the 1.0 release of the operator.

burmanm:
In that case we should not do anything with trunk in my opinion, since it basically can't be updated. So only filter versions when we have a known supported version.

adejanovski:
@burmanm, I agree that a fully automated way of generating this would be very valuable. I don't see that happening for 1.0 indeed as there are many edge cases to deal with (type conversions, specific values for tests, etc...). 
While trunk is present in the spreadsheet, it's not managed in the code, we just deal with 3.11 and 4.0 there.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1195,6 +1198,114 @@ func createMultiDcClusterWithStargate(t *testing.T, ctx context.Context, f *fram
 	verifyObjectDoesNotExist(ctx, t, f, sg2Key, &stargateapi.Stargate{})
 }
 
+// changeNumTokensValue creates a Datacenter and then changes the numTokens value
+// Such change is prohibited and should fail
+func changeNumTokensValue(t *testing.T, ctx context.Context, f *framework.Framework, namespace string) {
+	require := require.New(t)
+
+	k8sCtx := "cluster-1"
```
jsanda, line 1206: The integration test framework creates 3 clusters - cluster-0, cluster-1, cluster-2. cluster-0 is implicitly the control plane cluster.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1195,6 +1198,114 @@ func createMultiDcClusterWithStargate(t *testing.T, ctx context.Context, f *fram
...
+								StorageClassName: &defaultStorageClass,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	err := f.Client.Create(ctx, kc)
```
jsanda, line 1240: This creates the K8ssandraCluster in `cluster-0` but also note that the CassandraDatacenter will be created in `cluster-1`.

jsanda: `f.Client` always refers to `cluster-0`, i.e., the control plane cluster.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1195,6 +1198,114 @@ func createMultiDcClusterWithStargate(t *testing.T, ctx context.Context, f *fram
...
+			Type:               cassdcapi.DatacenterScalingUp,
+			Status:             corev1.ConditionFalse,
+			LastTransitionTime: lastTransitionTime,
+		})
+	})
+	require.NoError(err, "failed to patch datacenter status")
+
+	// Update the datacenter with a different num_tokens value and check that it failed
+	initialNumTokens := kc.Spec.Cassandra.CassandraConfig.CassandraYaml.NumTokens
+	kcKey := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: "test"}, K8sContext: k8sCtx}
```
jsanda, line 1285: Because the K8ssandraCluster is always (and has to be) created in the control plane cluster, I just use a `client.ObjectKey` which is an alias for `types.NamespacedName`. We have a small helper function, `utils.GetKey(kc)` that I like to use. 

Since the K8ssandraCluster is always in the control plane cluster I typically use `f.Client` for working with it. Maybe it would be better to rename that to `f.ControlPlaneClient`, although that might be a bit verbose 🤔

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -1195,6 +1198,114 @@ func createMultiDcClusterWithStargate(t *testing.T, ctx context.Context, f *fram
...
+	kcPatch := client.MergeFrom(kc.DeepCopy())
+	kc.Spec.Cassandra.CassandraConfig.CassandraYaml.NumTokens = pointer.Int(256)
+	err = f.Patch(ctx, kc, kcPatch, kcKey)
+	require.NoError(err, "got error patching num_tokens")
```
jsanda, line 1289: Hopefully my previous comments make it more clear why this patch fails 😅

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -327,7 +327,7 @@ type EmbeddedObjectMeta struct {
 // TODO Implement Stringer interface. It will helpful for debugging and testing.
 type CassandraConfig struct {
 	// +optional
-	CassandraYaml CassandraYaml `json:"cassandraYaml,omitempty"`
+	CassandraYaml CassandraYaml `json:"cassandra-yaml,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: I reverted to `cassandraYaml`. I found it nice to be consistent with cass-operator, but it's fine either way.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -336,90 +336,943 @@ type CassandraConfig struct {
 // CassandraYaml defines the contents of the cassandra.yaml file. For more info see:
 // https://cassandra.apache.org/doc/latest/cassandra/configuration/cass_yaml_file.html
 type CassandraYaml struct {
+	// Exists in 3.11, 4.0, trunk
```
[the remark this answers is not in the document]
adejanovski: Sure thing 👍

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: ooooh, I see. Things make sense now 😅

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: Gotcha, I'll use `cluster-0` only for this test, which makes things simpler.

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: it does, thanks!

adejanovski:
@jsanda, I made the requested changes and fixed the unit test.
Let me know what you think.

jsanda (approved):
Thanks for this PR!

I know tests are still running but I'm gonna go ahead and approve. I'll look into any e2e failures since we have some flakes.

PR #289, opened 2022-01-12
Add the creation of the CQL user for Medusa

adejanovski:
**What this PR does**:
The Medusa user was created as a secret but wasn't added to the cassdc spec, thus it wasn't created in system_auth.
This PR adds the necessary bits to get that user created correctly.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/medusa/reconcile.go
+++ b/pkg/medusa/reconcile.go
@@ -6,7 +6,6 @@ import (
 	"text/template"
 
 	"github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
-	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	k8ss "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/medusa/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/medusa/reconcile.go
```diff
@@ -360,3 +361,16 @@ func addOrUpdateAdditionalVolume(dcConfig *cassandra.DatacenterConfig, volume *v
 		dcConfig.StorageConfig.AdditionalVolumes[volumeIndex] = *volume
 	}
 }
+
+// If auth is enabled in this cluster, we need to allow Medusa to access the cluster through CQL. This is done by
+// declaring a Cassandra user whose credentials are pulled from CassandraUserSecretRef.
+func AddCqlUser(medusaTemplate *api.MedusaClusterTemplate, dcConfig *cassandra.DatacenterConfig, clusterName string) {
```
jsanda, line 367: The `reaper` package defines a nearly identical function (see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/reaper/datacenter.go#L33)). The only difference is that the first argument is a `*reaperapi.ReaperClusterTemplate`. Could you refactor these into a common, shared function?

adejanovski: Good idea 👍 

Done!

PR #292, opened 2022-01-12
ClientConfig controller [K8SSAND-944]

### LINKED ISSUE #178: K8SSAND-944 ⁃ Updating remote k8s cluster connections requires a restart of the operator
**What is missing?**
If I add a ClientConfig for a new k8s cluster after the operator has already been started, the operator has to be restarted in order for it to create the remote cluster client connection. ClientConfigs are only processed at start up. This has to do with how controller-runtime behaves. 

At startup we create the controllers and add them to a `Manager`(from controller-runtime) by calling its `Add` method. The `Manager` manages the lifecycle of controllers. Finally we call `Manager.Start()` to start our controllers. The `Manager.Add` function accepts a `Runnable` as an argument. `Runnable` is an interface that defines a `Start` method. Controllers implement the `Runnable` interface. 

We are not creating separate controllers for remote clusters. We are creating `Cluster` objects which essentially provide connectivity to a cluster. A `Cluster` creates a `Cache` from which our Reconcilers read. The cache includes the components necessary for getting and processing event notifications from the API server. A `Cache` also has to be started which is handled by a `Manager`.

The problem is that is that the `Manager` won't start a `Cache` after `Manager.Start` has already been called. Could we just start the `Cache` on our own. Potentially yes, but it is non-trivial. I think we would wind up duplicating a fair bit of `Manager` logic. I created https://github.com/kubernetes-sigs/controller-runtime/issues/1673 so that we hopefully can get this fixed in controller-runtime.

When the necessary changes are made in controller-runtime, there is still a bit of work to do in our operator. We will need to implement a controller for ClientConfig.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-944) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-944
┆priority: Medium

burmanm:
**What this PR does**:
When a ClientConfig is modified or the Secret that ClientConfig refers to, the k8ssandra-operator is shutdown to allow automated restart.

Also, fixes a bug where watchNamespace having a comma would prevent correct loading of ClientConfigs.

Fixes #178 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/config/clientconfig_controller.go
+++ b/controllers/config/clientconfig_controller.go
@@ -53,12 +53,38 @@ type ClientConfigReconciler struct {
 			- Security /  user rights? If a namespace is allowed to add a K8ssandraCluster, but not new Kubernetes clusters to connect to?
 */
 
-/*
-	TODO Add a channel that other controllers listen? So that they can add new Watches?
-	TODO Expose new cluster loading watchers in each controller?
-*/
-
 func (r *ClientConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	clientConfig := configapi.ClientConfig{}
+	if err := r.Client.Get(ctx, req.NamespacedName, &clientConfig); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// ClientConfig was deleted, shutdown to refresh correct list
+	if clientConfig.GetDeletionTimestamp() != nil {
+		r.shutdownFunc()
+		return ctrl.Result{}, nil
+	}
+
+	// ClientConfig without proper annotations, must be a new item, shutdown to refresh correct list
+	if !metav1.HasAnnotation(clientConfig.ObjectMeta, ClientConfigHashAnnotation) ||
+		metav1.HasAnnotation(clientConfig.ObjectMeta, KubeSecretHashAnnotation) {
+		r.shutdownFunc()
+		return ctrl.Result{}, nil
+	}
+
+	cCfgHash, secretHash, err := calculateHashes(ctx, r.Client, clientConfig)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Verify hashes are still original
+	if clientConfig.Annotations[ClientConfigHashAnnotation] != cCfgHash ||
+		clientConfig.Annotations[KubeSecretHashAnnotation] != secretHash {
+		// Hashes do not match, something was modified, shutdown to refresh
+		r.shutdownFunc()
+		return ctrl.Result{}, nil
+	}
+
 	return ctrl.Result{}, nil
 }
 
--- a/controllers/config/clientconfig_controller.go
+++ b/controllers/config/clientconfig_controller.go
@@ -71,7 +71,7 @@ func (r *ClientConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// ClientConfig without proper annotations, must be a new item, shutdown to refresh correct list
 	if !metav1.HasAnnotation(clientConfig.ObjectMeta, ClientConfigHashAnnotation) ||
-		metav1.HasAnnotation(clientConfig.ObjectMeta, KubeSecretHashAnnotation) {
+		!metav1.HasAnnotation(clientConfig.ObjectMeta, KubeSecretHashAnnotation) {
 		logger.Info(fmt.Sprintf("ClientConfig %v is missing hash annotations, shutting down the operator", req))
 		r.shutdownFunc()
 		return ctrl.Result{}, nil
@@ -97,7 +97,9 @@ func (r *ClientConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 // SetupWithManager will only set this controller to listen in control plane cluster
 func (r *ClientConfigReconciler) SetupWithManager(mgr ctrl.Manager, cancelFunc context.CancelFunc) error {
 	r.shutdownFunc = cancelFunc
-	r.secretFilter = make(map[types.NamespacedName]types.NamespacedName)
+	if r.secretFilter == nil {
+		r.secretFilter = make(map[types.NamespacedName]types.NamespacedName)
+	}
 
 	// We should only reconcile objects that match the rules
 	toMatchingClientConfig := func(secret client.Object) []reconcile.Request {
@@ -119,6 +121,8 @@ func (r *ClientConfigReconciler) SetupWithManager(mgr ctrl.Manager, cancelFunc c
 // InitClientConfigs will fetch clientConfigs from the current cluster (control plane cluster) and create all the required Cluster objects for
 // other controllers to use. Not called from SetupWithManager since other controllers need the []cluster.Cluster array
 func (r *ClientConfigReconciler) InitClientConfigs(ctx context.Context, mgr ctrl.Manager, watchNamespace string) ([]cluster.Cluster, error) {
+	logger := log.FromContext(ctx)
+
 	uncachedClient := r.ClientCache.GetLocalNonCacheClient()
 	clientConfigs := make([]configapi.ClientConfig, 0)
 	namespaces := strings.Split(watchNamespace, ",")
@@ -133,6 +137,9 @@ func (r *ClientConfigReconciler) InitClientConfigs(ctx context.Context, mgr ctrl
 	}
 
 	additionalClusters := make([]cluster.Cluster, 0, len(clientConfigs))
+
+	// TODO Secret could point to multiple clientConfigs. Shouldn't matter in our current use-case
+	r.secretFilter = make(map[types.NamespacedName]types.NamespacedName, len(clientConfigs))
 
 	for _, cCfg := range clientConfigs {
 		// Calculate hashes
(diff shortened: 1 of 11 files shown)


jsanda:
This is at the front of my queue for reviews tomorrow.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+	Scheme       *runtime.Scheme
+	ClientCache  *clientcache.ClientCache
+	shutdownFunc context.CancelFunc
+
+	// filterMutex  sync.RWMutex
+	secretFilter map[types.NamespacedName]types.NamespacedName
+}
+
+/*
+	What about situations:
```
jsanda, line 44: These are great questions. Thanks for adding them. I'll add some comments, but the larger discussions definitely need to happen outside this PR.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		- Remove clientconfig when the cluster is still added?
```
jsanda, line 45: Not sure I understand. Can you elaborate?

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+			* Cluster properties get updated?
+				* Do we accept or not?
+			* Could just be a cert-update etc, short-term, long-term ones
+		- New ClientConfig doesn't work?
```
jsanda, line 49: Ultimately I think there should be client-side tools that generate the ClientConfigs. They would validate the configuration.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		- Add cluster points to a clientconfig that doesn't exist?
```
jsanda, line 50: Again, something I think best handled by client-side tools.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		- What if the ClientConfig's kubeConfig is changed in the secret?
```
jsanda, line 51: In the context of this PR I am inclined to say that it should trigger a restart. Maybe a delayed restart where the delay is configurable. I suggest a delay in case there are multiple ClientConfigs to be updated.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		- What if certs expire? What if client connection fails - how do we detect when we need to do something to clientConfig?
```
jsanda, line 52: Excellent questions but outside the scope of what we trying to do here. It would be great to create issues for these. 

Cert expiration and connection failing are distinct and I think should be handled separately. Cert expiration is something we should be able to determine easily, failed client connections not so much. How do we distinguish between transitory failures and a permanently bad connection?

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		- Do we load clientConfigs from all watch namespaces (or empty - cluster wide) or just from where operator is installed?
```
jsanda, line 53: If the operator is configured to watch all namespaces, then I say yes load from all namespaces; otherwise, just the namespace in which the operator is deployed. For 1.0 though if limiting it to only the namespace in which the operator is deployed is easier, I think that is perfectly fine.

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,215 @@
...
+
+		metav1.SetMetaDataAnnotation(&cCfg.ObjectMeta, ClientConfigHashAnnotation, cCfgHash)
+		metav1.SetMetaDataAnnotation(&cCfg.ObjectMeta, KubeSecretHashAnnotation, secretHash)
+
+		if err := uncachedClient.Update(ctx, &cCfg); err != nil {
+			return nil, err
+		}
+
+		// Add the Secret to the cache
+		r.secretFilter[secretName] = cCfgName
```
jsanda, line 162: Why do we need this cache? I understand that different ClientConfigs can point to the same secret. The secret needs to be in the same namespace as the ClientConfig and we're not managing the secrets. In particular if a ClientConfig is deleted, we aren't deleting the secret it references.

jsanda (dismissed):
We need an e2e test.

jsanda:
I accidentally selected the approve option in my review. I mean to select the option for changes requested, namely an e2e test.

jsanda (changes requested):
Need an e2e test

controllers/config/clientconfig_controller.go
burmanm: Actually, k8ssandra-operator fails to start at the moment if it can't connect to the cluster it was configured to connect to. The kubeConfig -> RESTClient part throws an error and the initialization fails in main.go

controllers/config/clientconfig_controller.go
burmanm: It does trigger a restart in this PR's implementation.

controllers/config/clientconfig_controller.go
burmanm: Because we can't necessarily manage the secrets, so we need a way to manage which secrets we monitor. If that secret is modified, we have to link that to the ClientConfig which we would reconcile.

controllers/config/clientconfig_controller.go
jsanda: What do you think about a delayed restart?

burmanm: What would be the benefit of delayed restart?

jsanda: Let's say we create 3 ClientConfig objects. I haven't tested yet but I suspect that would result in 3 successive restarts.

controllers/config/clientconfig_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+	scheme      *runtime.Scheme
+	logger      logr.Logger
+	cancelCalls int
+	// secretFilter map[types.NamespacedName]types.NamespacedName
+	reconciler *ClientConfigReconciler
+	usedMgr    manager.Manager
+	kubeConfig []byte
+)
+
+func TestClientConfigReconciler(t *testing.T) {
```
jsanda, line 42: Can you add or update tests to verify shutdown happens when:

* ClientConfig is deleted
* ClientConfig or secret is missing hash annotations
* Hash annotations have changed

test/e2e/config_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	clientConfigList := &configapi.ClientConfigList{}
+	err = f.Client.List(ctx, clientConfigList, client.InNamespace(namespace))
+	require.NoError(err)
+	require.True(len(clientConfigList.Items) > 0)
+
+	currentRestartCount := operatorPod.Status.ContainerStatuses[0].RestartCount
+
+	clientConfig := &configapi.ClientConfig{
+		ObjectMeta: metav1.ObjectMeta{
```
jsanda, line 32: I like the names 🤣

test/e2e/config_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	configapi "github.com/k8ssandra/k8ssandra-operator/apis/config/v1beta1"
+	"github.com/k8ssandra/k8ssandra-operator/test/framework"
+)
+
+func controllerRestart(t *testing.T, ctx context.Context, namespace string, f *framework.E2eFramework) {
```
jsanda, line 17: It would be nice to have this (or a different) test deploy a K8ssandraCluster that uses the new, valid ClientConfig. That will provide good coverage over `main.go`.

jsanda (changes requested):

controllers/config/clientconfig_controller.go
```diff
@@ -0,0 +1,215 @@
...
+		- What if the ClientConfig's kubeConfig is changed in the secret?
+		- What if certs expire? What if client connection fails - how do we detect when we need to do something to clientConfig?
+		- Do we load clientConfigs from all watch namespaces (or empty - cluster wide) or just from where operator is installed?
+			- Security /  user rights? If a namespace is allowed to add a K8ssandraCluster, but not new Kubernetes clusters to connect to?
+*/
+
+func (r *ClientConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	clientConfig := configapi.ClientConfig{}
```
jsanda, line 60: This isn't handling deletions correctly. If I delete a ClientConfig, the `Get` call returns an IsNotFound error. We need to check the error type and call `shutdownFunc` when it is IsNotFound. 

Since we aren't adding a finalizer, I don't think it is necessary to check the deletion timestamp.

jsanda (changes requested):

test/e2e/config_test.go
burmanm: I think main.go is tested every time any test runs? Wouldn't we fail to connect to other clusters otherwise?

jsanda: Good point

jsanda (approved):
LGTM. Thanks!

PR #301, opened 2022-01-17
Remove client.DeleteAllOf() from servicemonitor cleanups [K8SSAND-1149]

### LINKED ISSUE #279: K8SSAND-1149 ⁃ Do not use client.DeleteAllOf to delete ServiceMonitors
If we know or can compute the key for an object it is best to delete by key. ServiceMonitors however are deleted by calling `client.DeleteAllOf`.

#260 provided the initial ServiceMonitor integration for CassandraDatacenter and Stargate objects. The reconciliation logic handles deletions as follows:

```go
func (cfg PrometheusResourcer) CleanupResources(ctx context.Context, client runtimeclient.Client) error {
	if len(cfg.CommonLabels) < 1 {
		return errors.New("no labels were found to target deletion request")
	}
	if err := cfg.validate(); err != nil {
		return TelemetryConfigIncomplete{}
	}
	var deleteTargets promapi.ServiceMonitor
	err := client.DeleteAllOf(ctx,
		&deleteTargets,
		runtimeclient.InNamespace(cfg.MonitoringTargetNS),
		runtimeclient.MatchingLabels(cfg.CommonLabels),
	)
	if err != nil {
		return err
	}
	return nil
}
```

We had some back and forth discussion about the usage of `DeleteAllOf` while #260 was in flight. In fact, I initially suggested using it and then later suggested switching to `client.Delete()`. If the ServiceMonitor CRD is installed the k8s cluster, then `CleanupResources` could in fact be called frequently.

After thinking about it some more, we should be using `client.Delete()` and not `client.DeleteAllOf`. 

While this code isn't necessarily executed all the time, there are performance considerations that merit the change.  With `DeleteAllOf` I believe the api server has to fetch all ServiceMonitors in the namespace. It then does the label filtering on the client side (with respect to etcd).  Let's contrast this with doing a get followed by a delete.

There are a couple alternatives to using `DeleteAllOf`. We could simply do a `client.Delete()` to delete by key. This is slightly better but we still hit the api server every time `CleanupResources` is called. Let's consider first doing a `client.Get()`. If the ServiceMonitor doesn't exist or has already been deleted, `Get()` will return a not found error in which case there is not need to call `client.Delete()`. This is good because all reads go against the cache which means we would eliminate a lot of unnecessary calls against the api server. But what if the cache is stale? Not a problem. It can be and often is. If/when that happens it will get sorted out on a later reconciliation.

This should be done for 1.0.0.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1149) by [Unito](https://www.unito.io)
┆epic: Prometheus integration
┆friendlyId: K8SSAND-1149
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Uses a `client.Delete()` call instead of a `client.DeleteAllOf()` in the `PrometheusResourcer.CleanupResources()` method.

Fixes #279

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
`CreateMultiStargateAndDatacenter` is intermittently failing it seems. It passes locally but has failed in the previous test run. Restarting the workflow to see if we can get it to pass for hygiene purposes.

Miles-Garnsey:
All tests passing (except `ClusterScoped/MultiDcMultiCluster` which is broken in all branches right now) and ready for review.

jsanda (approved):
LGTM. Thanks!

PR #307, opened 2022-01-18
Fix Reaper secret reconciliation

adutra:
**What this PR does**:
This commit fixes Reaper secret reconciliation when
Reaper is present for some dcs, but not for all of them.

If the Reaper secret doesn't exist when the dc is created,
we get the following error:

    Validation of user secret failed due to an error:
    Secret \"test-reaper\" not found

None

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/secrets.go
+++ b/controllers/k8ssandra/secrets.go
@@ -40,8 +40,10 @@ func (r *K8ssandraClusterReconciler) reconcileReaperSecrets(ctx context.Context,
 		if kc.Spec.IsAuthEnabled() {
 			logger.Info("Reconciling Reaper user secrets")
 			var cassandraUserSecretRef corev1.LocalObjectReference
+			var jmxUserSecretRef corev1.LocalObjectReference
 			if kc.Spec.Reaper != nil {
 				cassandraUserSecretRef = kc.Spec.Reaper.CassandraUserSecretRef
+				jmxUserSecretRef = kc.Spec.Reaper.JmxUserSecretRef
 			}
 			if cassandraUserSecretRef.Name == "" {
 				cassandraUserSecretRef.Name = reaper.DefaultUserSecretName(kc.Name)
--- a/controllers/k8ssandra/secrets.go
+++ b/controllers/k8ssandra/secrets.go
@@ -8,6 +8,7 @@ import (
 	"github.com/k8ssandra/k8ssandra-operator/pkg/result"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/secret"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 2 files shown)


adejanovski (commented):
I suggested a small refactoring, but I struggle to understand why we would not want to have one reaper instance per DC when deploying a multi DC cluster (which motivates this PR as `kc.Spec.Reaper` could not exist while Reaper is still deployed on select DCs).
The Reaper instances work as a network, sharing their work and storage, so having them deployed with different settings seems weird and error prone. Also, not deploying them everywhere means we'll be using datacenterAvailability of LOCAL, which is less safe than EACH.
I think we should probably simplify this and only retain the `ReaperClusterTemplate` while dropping the `ReaperDatacenterTemplate`. 
Sorry if I'm a little late to the party here, but doing this post 1.0.0 means breaking changes, which would defer the change to 2.0.0.

adutra:

I am not against the idea, @jsanda what would be your preference?

That said, after reading the page on [multi-dc setups for Reaper](http://cassandra-reaper.io/docs/usage/multi_dc/), I was under the assumption that deploying Reaper in just one dc with availability ALL or in a few dcs with availability LOCAL were also perfectly valid setups.

adejanovski:
> That said, after reading the page on multi-dc setups for Reaper, I was under the assumption that deploying Reaper in just one dc with availability ALL or in a few dcs with availability LOCAL were also perfectly valid setups.

That's true. Also `LOCAL` is valid but less safe.
But it seems like the dcAvailability is computed by the controller, setting it either to `EACH` or `LOCAL` depending on the number of reaper instances vs the number of datacenters: https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/reaper/resource.go#L67-L81

However we set it up, it looks like `ALL` is never an option. Is there some place in the code where this value gets overridden and may be set to `ALL`?

adutra:
> However we set it up, it looks like ALL is never an option. Is there some place in the code where this value gets overridden and may be set to ALL?

You can manually set it to ALL, see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/506d8dae3a3044ee30f0edc485ddab680787b3aa/apis/reaper/v1alpha1/reaper_types.go#L210).

Also note that this field defaults to LOCAL, which seemed to me as the safest choice if you don't know how many Reaper instances are going to be created in total.

But it is true that the controller never assigns ALL by itself. Maybe we should set it to ALL if there is only one Reaper instance for the entire cluster, wdyt?

adutra:
I would like this PR to get merged soonish though. @adejanovski @jsanda can we keep the discussion going in a follow-up PR?

jsanda:
I am definitely late to the discussion. Based on what @adejanovski described, it sounds like we want to avoid different, per-DC settings for Reaper. Am I correct in thinking that this is orthogonal to this PR? If so let's get a ticket created for the change. I agree that it would be better to do it now.

jsanda (approved):
LGTM

adutra:
I just created https://github.com/k8ssandra/k8ssandra-operator/issues/308.

adejanovski:
I forgot to follow up yesterday and approve the PR. Thanks for doing it @jsanda.

PR #312, opened 2022-01-21
Improve logic to compute Reaper availability mode [K8SSAND-1177]

### LINKED ISSUE #308: K8SSAND-1177 ⁃ Consider removing per-DC Reaper templates
Initially reported by adejanovski here:

https://github.com/k8ssandra/k8ssandra-operator/pull/307#pullrequestreview-856321315

> I struggle to understand why we would not want to have one reaper instance per DC when deploying a multi DC cluster (which motivates this PR as kc.Spec.Reaper could not exist while Reaper is still deployed on select DCs).
> The Reaper instances work as a network, sharing their work and storage, so having them deployed with different settings seems weird and error prone. Also, not deploying them everywhere means we'll be using datacenterAvailability of LOCAL, which is less safe than EACH.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1177) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1177
┆priority: Medium

adutra:
**What this PR does**:

This commit changes the logic to be as follows:

* For single-DC clusters, always use ALL;
* For multi-DC clusters:
  * If all DCs have their Reaper instance, use EACH;
  * Otherwise use ALL.

Fixes #308

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reaper/resource.go
+++ b/pkg/reaper/resource.go
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	DatacenterAvailabilityLocal = "LOCAL"
-	DatacenterAvailabilityEach  = "EACH"
+	DatacenterAvailabilityEach = "EACH"
+	DatacenterAvailabilityAll  = "ALL"
 )
 
 // DefaultResourceName generates a name for a new Reaper resource that is derived from the Cassandra cluster and DC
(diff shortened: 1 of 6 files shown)


pkg/reaper/resource.go
```diff
@@ -13,6 +13,7 @@ import (
 const (
 	DatacenterAvailabilityLocal = "LOCAL"
```
jsanda, line 14: Is `LOCAL` used anywhere other than tests now? If not should it be removed?

jsanda (commented):
It would nice to see an e2e test where the DC availability changes from ALL to EACH. That would have to come in a follow up PR because we need #262 merged first.

I briefly looked at your [prototype](https://github.com/k8ssandra/k8ssandra-operator/compare/reaper-availability...adutra:reaper-per-cluster?expand=1). I need to look at it more but I like it.

adutra:
> It would nice to see an e2e test where the DC availability changes from ALL to EACH. That would have to come in a follow up PR because we need #262 merged first.
> 
> I briefly looked at your [prototype](https://github.com/k8ssandra/k8ssandra-operator/compare/reaper-availability...adutra:reaper-per-cluster?expand=1). I need to look at it more but I like it.

It's now a PR, see #322 . It's based on this one though.

adutra:
Superseded by #322.

PR #316, opened 2022-01-24
Add encryption options [K8SSAND-1165]

### LINKED ISSUE #293: K8SSAND-1165 ⁃ Support encryption options in Cassandra Configuration
The first iteration on supporting Cassandra yaml settings in a structured way didn't deal with encryption.
We need to add support for it, which includes mounting volumes to access the key/trust stores and handle the corresponding secrets.

Some pointers for this can be found in this 1.x PR: https://github.com/k8ssandra/k8ssandra/pull/1180/files

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1165) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1165
┆priority: Medium

adejanovski:
**What this PR does**:
Adds support for internode and client to server encryption. 

Fixes #293 

**Checklist**
- [x] Changes manually tested


pkg/stargate/deployments.go
```diff
@@ -332,28 +338,62 @@ func computeHeapSize(template *api.StargateTemplate) resource.Quantity {
...
+	mounts = append(mounts, encryptionStoreMounts(encryptionVolumes)...)
+	return mounts
+}
+
+func encryptionStoreMounts(encryptionVolumes []corev1.Volume) []corev1.VolumeMount {
+	mounts := []corev1.VolumeMount{}
+	for _, volume := range encryptionVolumes {
+		mounts = append(mounts, corev1.VolumeMount{
+			Name:      volume.Name,
+			MountPath: cassandra.StoreMountFullPath(strings.Split(volume.Name, "-")[0], strings.Split(volume.Name, "-")[1]),
```
adutra, line 393: Splitting strings sounds a bit fragile, I'd rather compute the mount path from the original `EncryptionStores` , just as you did for volumes in `computeEncryptionVolumes` above.

pkg/stargate/deployments.go
```diff
@@ -17,8 +19,10 @@ import (
 )
 
 const (
-	cassandraConfigDir  = "/config"
-	cassandraConfigPath = "/config/cassandra.yaml"
+	cassandraConfigDir     = "/config"
+	cassandraConfigPath    = "/config/cassandra.yaml"
+	CassandraConfigMap     = "generated-cassandra-config"
+	UserCassandraConfigMap = "cassandra-config"
```
adutra, line 25: `UserCassandraConfigMap` seems unused.

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("truststore password secret ref was not set")
+	}
+	return nil
+}
+
+func StoreMountFullPath(storeType string, storeName string) string {
+	return fmt.Sprintf("%s/%s-%s", StoresMountPath, storeType, storeName)
+}
+
+func ClientEncryptionEnabled(template *DatacenterConfig) bool {
```
adutra, line 149: Suggestion: turn this function and the next one into methods with pointer receivers: `ClientEncryptionOptions.IsEnabled()` and `ServerEncryptionEnabled.IsEnabled()`. This way you can handle the nil case inside the methods:

```go
func (in *ServerEncryptionOptions) IsEnabled() bool {
	return in != nil && in.Enabled != nil && *in.Enabled
}
```

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, volume := range volumes {
+		indexKey, foundKey := FindVolume(template.PodTemplateSpec, volume.Name)
+		AddOrUpdateVolume(template, &volume, indexKey, foundKey)
+	}
+
+	// Find the cassandra container and add the volume mounts for both stores
+	addEncryptionMountToCassandra(template, volumes["keystore"], volumes["truststore"], storeType)
+}
+
+func EncryptionVolumes(storeType string, encryptionStores encryption.EncryptionStores) map[string]corev1.Volume {
```
adutra, line 58: I see a `storeType` parameter in many functions but it seems it can only be `client` or `server`. We coudl capture this with a dedicated type:

```go
type StoreType string

const (
	StoreTypeClient = StoreType("client")
	StoreTypeServer = StoreType("server")
)

func EncryptionVolumes(storeType StoreType, encryptionStores encryption.EncryptionStores) map[string]corev1.Volume {
...
```

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,217 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	StoresMountPath = "/mnt"
+)
+
+// Sets up encryption in the datacenter config template.
```
adutra, line 18:
```suggestion
// HandleEncryptionOptions sets up encryption in the datacenter config template.
```

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,217 @@
...
+	if encryptionStores == nil {
+		return fmt.Errorf("EncryptionStores is required to set up encryption")
+	}
+	if encryptionStores.KeystoreSecretRef.Name == "" {
+		return fmt.Errorf("keystore secret ref was not set")
+	}
+	if encryptionStores.KeystorePasswordSecretRef.Name == "" {
+		return fmt.Errorf("keystore password secret ref was not set")
+	}
+	if encryptionStores.TruststoreSecretRef.Name == "" {
```
adutra, line 136: I'm not sure here, I think the truststore is only required if `require_client_auth` is true.

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,217 @@
...
+	ClientKeystorePassword   string
+	ClientTruststorePassword string
+	ServerKeystorePassword   string
+	ServerTruststorePassword string
+}
+
+func ReadEncryptionStoresSecrets(ctx context.Context, klusterKey types.NamespacedName, template *DatacenterConfig, remoteClient client.Client, logger logr.Logger) (EncryptionStoresPasswords, error) {
+	encryptionStoresPasswords := EncryptionStoresPasswords{}
+	if ClientEncryptionEnabled(template) {
+		if template.ClientEncryptionStores == nil || template.ClientEncryptionStores.KeystorePasswordSecretRef.Name == "" {
```
adutra, line 167: Why not call `checkMandatoryEncryptionFields` here?

pkg/cassandra/config.go
```diff
@@ -103,6 +103,40 @@ func newConfig(apiConfig api.CassandraConfig, cassandraVersion string) (config,
...
+			keystorePath := fmt.Sprintf("%s/%s", StoreMountFullPath("client", "keystore"), "keystore")
+			truststorePath := fmt.Sprintf("%s/%s", StoreMountFullPath("client", "truststore"), "truststore")
+			updatedConfig.CassandraYaml.ClientEncryptionOptions.Keystore = pointer.String(keystorePath)
+			updatedConfig.CassandraYaml.ClientEncryptionOptions.Truststore = pointer.String(truststorePath)
+			updatedConfig.CassandraYaml.ClientEncryptionOptions.KeystorePassword = &encryptionStoresSecrets.ClientKeystorePassword
+			updatedConfig.CassandraYaml.ClientEncryptionOptions.TruststorePassword = &encryptionStoresSecrets.ClientTruststorePassword
+		}
+	}
+	if updatedConfig.CassandraYaml.ServerEncryptionOptions != nil {
+		if *updatedConfig.CassandraYaml.ServerEncryptionOptions.Enabled {
```
adutra, line 123: Beware that you could have a nil pointer dereference here if `ServerEncryptionOptions.Enabled` is nil – which is why I advocate for creating a dedicated method `IsEnabled()`.

pkg/cassandra/config.go
```diff
@@ -103,6 +103,40 @@ func newConfig(apiConfig api.CassandraConfig, cassandraVersion string) (config,
...
+			keystorePath := fmt.Sprintf("%s/%s", StoreMountFullPath("server", "keystore"), "keystore")
+			truststorePath := fmt.Sprintf("%s/%s", StoreMountFullPath("server", "truststore"), "truststore")
+			updatedConfig.CassandraYaml.ServerEncryptionOptions.Keystore = pointer.String(keystorePath)
+			updatedConfig.CassandraYaml.ServerEncryptionOptions.Truststore = pointer.String(truststorePath)
+			updatedConfig.CassandraYaml.ServerEncryptionOptions.KeystorePassword = &encryptionStoresSecrets.ServerKeystorePassword
+			updatedConfig.CassandraYaml.ServerEncryptionOptions.TruststorePassword = &encryptionStoresSecrets.ServerTruststorePassword
+		}
+		// The encryption stores shouldn't end up in the cassandra yaml, they are specific to k8ssandra
+		if IsCassandra3(cassandraVersion) {
```
adutra, line 132: I'm not sure but it seems that `Optional` is not available for Cassandra 3 either.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	// +optional
+	CipherSuites *[]string `json:"cipher_suites,omitempty"`
+
+	// default: false
+	// +optional
+	RequireClientAuth *bool `json:"require_client_auth,omitempty"`
+
+	// default: none
+	// +optional
+	InternodeEncryption *string `json:"internode_encryption,omitempty"`
```
adutra, line 1146: This is an enum it seems, from [the docs](https://cassandra.apache.org/doc/latest/cassandra/configuration/cass_yaml_file.html#server_encryption_options):

```
# On outbound connections, determine which type of peers to securely connect to.
        #   The available options are :
        #     none : Do not encrypt outgoing connections
        #     dc   : Encrypt connections to peers in other datacenters but not within datacenters
        #     rack : Encrypt connections to peers in other racks but not within racks
        #     all  : Always use encrypted connections
```

So you could add: 

```
	// +optional
	// +kubebuilder:validation:Enum=none;dc;rack;all
	// +kubebuilder:default=none
```

adutra (changes requested):
First round of review, will continue tomorrow but looking good so far.

pkg/stargate/deployments.go
```diff
@@ -332,28 +338,62 @@ func computeHeapSize(template *api.StargateTemplate) resource.Quantity {
 	return resource.MustParse("256Mi")
 }
 
+// This config map will always be created by the k8ssandra controller.
+// It will augment the user provided config map with encryption settings if enabled.
 func computeVolumes(template *api.StargateTemplate) []corev1.Volume {
```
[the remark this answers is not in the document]
adutra: Alternatively it could be merged with `computeEncryptionVolumes` below, just like `computeVolumeMounts` works.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
jsanda: I would take it step further and define a custom type, e.g., `type EncryptionType string` which could be used for both internode and client encryption. Also, no need for a pointer.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+type ClientEncryptionOptions struct {
+	Enabled bool `json:"enabled"`
+
+	// +optional
+	Optional *bool `json:"optional,omitempty"`
+
+	// Should not be set explicitly in the custom resource.
+	// The operator will generate the right value based on the EncryptionStores field.
+	// +optional
+	Keystore *string `json:"keystore,omitempty"`
```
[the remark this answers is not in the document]
jsanda: What is needed are object references to the keystore and truststore secrets. The passwords should be stored in the secrets as well.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	// +optional
+	RequireClientAuth *bool `json:"require_client_auth,omitempty"`
+
+	// default: none
+	// +optional
+	InternodeEncryption *string `json:"internode_encryption,omitempty"`
+
+	// default: false
+	// +optional
+	RequireEndpointVerification *bool `json:"require_endpoint_verification,omitempty"`
```
jsanda, line 1150: I don't think we need pointers here.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
 	LocalReadSize       *int `json:"local_read_size,omitempty"`
 	RowIndexSize        *int `json:"row_index_size,omitempty"`
 }
+
+type ClientEncryptionOptions struct {
```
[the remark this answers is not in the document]
adejanovski: To be fair, this is how they are named in the Cassandra config yaml: https://github.com/apache/cassandra/blob/trunk/conf/cassandra.yaml#L1261

My assumption is that if we use the same names as in Cassandra, it shouldn't require additional comments or explanations.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	Enabled bool `json:"enabled"`
+
+	// +optional
+	Optional *bool `json:"optional,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: Again, this is explained in the Cassandra configuration file and I don't think we should duplicate those explanations in our code.

controllers/k8ssandra/datacenters.go
```diff
@@ -78,14 +90,38 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
+
+		desiredConfig, err := utils.UnmarshalToMap(desiredDc.Spec.Config)
 		if err != nil {
-			logger.Error(err, "Failed to get remote client")
 			return result.Error(err), actualDcs
 		}
 
-		if recResult := r.reconcileSeedsEndpoints(ctx, desiredDc, seeds, remoteClient, logger); recResult.Completed() {
-			return recResult, actualDcs
+		if dcTemplate.Stargate != nil {
```
adutra, line 102: I am wondering if this reconciliation logic shouldn't be done by the Stargate controller:

* The k8ssandra controller remains responsible for creating the `Stargate` resource and filling it with the required specs, including which secrets should be used for encryption.
* Then the Stargate controller picks up the Stargate resource and "merges" those secrets with whatever is specified in `CassandraConfigMapRef` to create a new merged configmap and use it.

I believe that this way it would be easier for standalone Stargate resources to be able to be reconciled correctly.

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+			configYamlString = string(configYaml)
+		}
+	}
+
+	if userConfigMapContent != "" {
+		separator := "\n"
+		if strings.HasSuffix(userConfigMapContent, "\n") {
+			separator = ""
+		}
+		configYamlString = userConfigMapContent + separator + configYamlString
```
adutra, line 228: What if the user configmap contains encryption entries? I think this could result in an invalid yaml file that will cause Stargate to fail to start.

I think it would be safer to manipulate maps instead:

```go
finalConfigMapContent := make(map[string]interface{}, 0)
for k, v := range userConfigMapContent { // assuming userConfigMapContent is a map not a string
	finalConfigMapContent[k] = v
}
for k, v := range filteredCassandraConfig {
	finalConfigMapContent[k] = v
}
finalConfigYaml, err := yaml.Marshal(finalConfigMapContent)
```

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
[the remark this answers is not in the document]
jsanda: Also ServerEncryptionOptions has all of the properties in ClientEncrpytionOptions and some additional ones. You could have ServerEncryptionOptions inherit from ClientEncryptionOptions, e.g.,

```go
type ServerEncryptionOptions struct {
    ClientEncryptionOptions `json:"inline"`

    // other options...
}

type ClientEncryptionOption
```

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	// +optional
+	Protocol *string `json:"protocol,omitempty"`
+
+	// +optional
+	AcceptedProtocols *[]string `json:"accepted_protocols,omitempty"`
+
+	// +optional
+	Algorithm *string `json:"algorithm,omitempty"`
+
+	// +optional
```
jsanda, line 1088: I was thinking we could use an enum here but it seems the supported file types may vary a bit by JVM.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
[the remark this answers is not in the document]
adejanovski: I provided alternative structures [here](https://github.com/adejanovski/k8ssandra-operator/pull/3/files) to partially decorrelate the CRD and the mapping structure.
Let me know what you think about it.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	CipherSuites *[]string `json:"cipher_suites,omitempty"`
+
+	// default: false
+	// +optional
+	RequireClientAuth *bool `json:"require_client_auth,omitempty"`
+}
+
+type ServerEncryptionOptions struct {
+	Enabled *bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: yup, I agree. I made the changes in the PR that reworks these structures and use a pointer for booleans.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	AcceptedProtocols *[]string `json:"accepted_protocols,omitempty"`
+
+	// +optional
+	Algorithm *string `json:"algorithm,omitempty"`
+
+	// +optional
+	StoreType *string `json:"store_type,omitempty"`
+
+	// +optional
+	CipherSuites *[]string `json:"cipher_suites,omitempty"`
```
jsanda, line 1092: If the available cipher suites is fixed, then we should use an enum validation here and also make this more strongly typed, e.g.,

```go
type CipherSuite string

const (   
    TLS_RSA_WITH_AES_128_CBC_SHA  CipherSuite  =  "TLS_RSA_WITH_AES_128_CBC_SHA"
    TLS_RSA_WITH_AES_256_CBC_SHA CipherSuite = "TLS_RSA_WITH_AES_256_CBC_SHA"
    ...
)

I know that the naming convention for constants in Go is camel case but I would inclined to leave them as is unless you can come up with better/shorter names.
```

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+
+	// default: false
+	// +optional
+	RequireClientAuth *bool `json:"require_client_auth,omitempty"`
```
jsanda, line 1096: Typically I find that pointers are unnecessary for bool and string. We should use value types if possible. Makes the code easier to manage.

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,15 @@
+package encryption
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+type EncryptionStores struct {
```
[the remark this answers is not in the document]
adejanovski: Good call 👍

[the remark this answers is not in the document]
adejanovski: I didn't need to use `DeepCopy()` so far with this structure 🤔 Did I miss something or do we not need it?

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,15 @@
...
+	KeystoreSecretRef corev1.LocalObjectReference `json:"keystore_secret_ref"`
```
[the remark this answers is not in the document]
adejanovski: no, they're all mandatory. The structure itself is optional, but if it's used then all 4 fields must be set.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -232,6 +233,14 @@ type CassandraClusterTemplate struct {
 	// api heap.
 	// +optional
 	MgmtAPIHeap *resource.Quantity `json:"mgmtAPIHeap,omitempty"`
+
+	// Internode encryption stores which are used by Cassandra and Stargate.
+	// +optional
+	ServerEncryptionStores *encryption.EncryptionStores `json:"server_encryption_stores,omitempty"`
```
jsanda, line 239: I know we use snake case for the properties that map directly to properties in cassandra.yaml; otherwise, the convention is camel case. It'd be nice to keep things consistent.

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,16 @@ type StargateSpec struct {
 	// +optional
 	// +kubebuilder:default=true
 	Auth *bool `json:"auth,omitempty"`
+
+	// Internode encryption stores which are used by Cassandra and Stargate.
+	// These should not be set explicitly by the user, they will be copied from the K8ssandraCluster object.
+	// +optional
+	ServerEncryptionStores *encryption.EncryptionStores `json:"server_encryption_stores,omitempty"`
```
jsanda, line 196: Similar to my previous comment, it would be nice to make these camel case to keep naming conventions more consistent.

pkg/stargate/deployments.go
```diff
@@ -332,28 +338,62 @@ func computeHeapSize(template *api.StargateTemplate) resource.Quantity {
 	return resource.MustParse("256Mi")
 }
 
+// This config map will always be created by the k8ssandra controller.
```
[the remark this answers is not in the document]
adejanovski: mmmmh, I didn't think of the case where Stargate could be used outside of K8ssandra.
My changes will prevent using the config map ref as expected because it's now used by the k8ssandra operator to generate the config map that will be mounted (which has a fixed name), using the referenced one.
Let me check how I can handle this better.

pkg/stargate/deployments.go
```diff
@@ -332,28 +338,62 @@ func computeHeapSize(template *api.StargateTemplate) resource.Quantity {
...
+// It will augment the user provided config map with encryption settings if enabled.
 func computeVolumes(template *api.StargateTemplate) []corev1.Volume {
-	if template.CassandraConfigMapRef != nil {
-		return []corev1.Volume{{
-			Name: "cassandra-config",
-			VolumeSource: corev1.VolumeSource{
-				ConfigMap: &corev1.ConfigMapVolumeSource{
-					LocalObjectReference: *template.CassandraConfigMapRef,
+	volumes := []corev1.Volume{}
```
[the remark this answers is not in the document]
adejanovski: gotcha, thanks

pkg/stargate/deployments.go
adejanovski: right. I must have removed its use in one of the iterations on the code.

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+	logger = logger.WithValues("StargateConfigMap", configMapKey)
+	desiredConfigMap := createStargateConfigMap(namespace, configYamlString, kc)
+	// Compute a hash which will allow to compare desired and actual configMaps
+	annotations.AddHashAnnotation(desiredConfigMap)
+	actualConfigMap := &corev1.ConfigMap{}
+
+	if err := remoteClient.Get(ctx, configMapKey, actualConfigMap); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Stargate configMap doesn't exist, creating it")
+			if err := remoteClient.Create(ctx, desiredConfigMap); err != nil {
```
jsanda, line 245: You should set an owner reference on the ConfigMap created by the controller so that deletes cascade.

controllers/k8ssandra/cleanup.go
```diff
@@ -85,6 +88,20 @@ func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.
 			}
 		}
 
+		// Delete the Stargate generated config map
+		sgConfigMap := corev1.ConfigMap{}
```
jsanda, line 92: There should be no need to delete the ConfigMap if you set an owner reference on it. The delete will cascade when the Stargate object is deleted.

jsanda (commented):
Initial review done.

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,16 @@ type StargateSpec struct {
 	// +optional
 	// +kubebuilder:default=true
 	Auth *bool `json:"auth,omitempty"`
+
+	// Internode encryption stores which are used by Cassandra and Stargate.
+	// These should not be set explicitly by the user, they will be copied from the K8ssandraCluster object.
```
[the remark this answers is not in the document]
adejanovski: ok, I'll remove the comment 👍

controllers/k8ssandra/stargate.go
adutra: Also, the current implementation might have a problem in the case where the user does not supply any configmap: I think in this case the YAML string that is generated here will contain only the encryption settings and nothing else, which is not what we want.

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+		configYamlString = userConfigMapContent + separator + configYamlString
+	}
+
+	configMapKey := client.ObjectKey{
+		Namespace: namespace,
+		Name:      stargate.CassandraConfigMap,
+	}
+
+	logger = logger.WithValues("StargateConfigMap", configMapKey)
+	desiredConfigMap := createStargateConfigMap(namespace, configYamlString, kc)
```
jsanda, line 237: We need to add a watch for ConfigMaps. This is missing in main as well. As things are currently in main, I thin we only need to add a watch in the Stargate reconciler. It should probably use a cluster label filter (see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/k8ssandra/k8ssandracluster_controller.go#L182)).

With these changes, it gets a little confusing. Since the K8ssandraCluster controller is reconciling the generated ConfigMap, I think it also needs a ConfigMap watch. 

This makes me wonder if it would be better for all of this ConfigMap work to be done in the Stargate controller.

jsanda (commented):
Next round of review :)

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -212,6 +213,11 @@ type ReaperSpec struct {
 	// +kubebuilder:default="LOCAL"
 	// +kubebuilder:validation:Enum:=LOCAL;ALL;EACH
 	DatacenterAvailability string `json:"datacenterAvailability,omitempty"`
+
+	// Client encryption stores which are used by Cassandra and Reaper.
+	// These should not be set explicitly by the user, they will be copied from the K8ssandraCluster object.
+	// +optional
+	ClientEncryptionStores *encryption.EncryptionStores `json:"client_encryption_stores,omitempty"`
```
adutra, line 220: As stated elsewhere, let's stick with camel case.

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -212,6 +213,11 @@ type ReaperSpec struct {
 	// +kubebuilder:default="LOCAL"
 	// +kubebuilder:validation:Enum:=LOCAL;ALL;EACH
 	DatacenterAvailability string `json:"datacenterAvailability,omitempty"`
+
+	// Client encryption stores which are used by Cassandra and Reaper.
+	// These should not be set explicitly by the user, they will be copied from the K8ssandraCluster object.
```
adutra, line 218: Same remark about this comment: better remove it as it doesn't concern standalone Reapers.

controllers/k8ssandra/stargate.go
adutra: > in this case the YAML string that is generated here will contain only the encryption settings and nothing else, which is not what we want.

I just confirmed with @jsanda that this is not a problem, so please disregard my comment.

pkg/cassandra/config.go
```diff
@@ -47,11 +47,11 @@ func (c config) MarshalJSON() ([]byte, error) {
 
 	// Even though we default to Cassandra's stock defaults for num_tokens, we need to
 	// explicitly set it because the config builder defaults to num_tokens: 1
-	if c.NumTokens == nil {
+	if c.CassandraYaml.NumTokens == nil {
```
adutra, line 50: This change isn't really necessary as embedded struct fields can be accessed directly. But I agree that this makes the code easier to understand, so 👍 .

pkg/reaper/deployment.go
```diff
@@ -127,6 +127,33 @@ func NewDeployment(reaper *api.Reaper, dc *cassdcapi.CassandraDatacenter, authVa
 		}
 	}
 
+	volumeMounts := []corev1.VolumeMount{}
+	volumes := []corev1.Volume{}
+	// if client encryption is turned on, we need to mount the keystore and truststore volumes
+	if reaper.Spec.ClientEncryptionStores != nil && keystorePassword != nil && truststorePassword != nil {
```
adutra, line 133: Reminder: if we reach the conclusion that we can have a keystore without truststore, then you will need to change this block.

test/testdata/fixtures/client-encryption-secret.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: client-encryption-stores
```
adutra, line 4: It's interesting to note that in `EncryptionStores` you distinguished 4 secret refs to define the keystore, its password, the truststore and its password. Yet here all 4 items are combined in one single secret, and hence the k8c definition references this single secret several times:

```
    server_encryption_stores:
      keystore_secret_ref:
        name: server-encryption-stores
      truststore_secret_ref:
        name: server-encryption-stores
      keystore_password_secret_ref:
        name: server-encryption-stores
      truststore_password_secret_ref:
        name: server-encryption-stores
    client_encryption_stores:
      keystore_secret_ref:
        name: client-encryption-stores
      truststore_secret_ref:
        name: client-encryption-stores
      keystore_password_secret_ref:
        name: client-encryption-stores
      truststore_password_secret_ref:
        name: client-encryption-stores
```

This hints at the fact that the granularity level is not optimal. I wonder if just one secret ref for keystore + truststore + passwords is enough? Or maybe two, one for the keystore, the other for the truststore?

adutra (changes requested):
Another round of review related to Reaper mostly.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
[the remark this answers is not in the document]
adejanovski: I'll add a link to the cassandra.yaml file from the Cassandra github repo. It contains all required explanations.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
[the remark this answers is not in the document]
adejanovski: After thinking a bit, I made the changes you suggested @jsanda. It'll be part of my next push on this branch.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	// Should not be set explicitly in the custom resource.
+	// The operator will generate the right value based on the EncryptionStores field.
+	// +optional
+	TruststorePassword *string `json:"truststore_password,omitempty"`
+
+	// +optional
+	Protocol *string `json:"protocol,omitempty"`
+
+	// +optional
+	AcceptedProtocols *[]string `json:"accepted_protocols,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: I've removed as many pointers as possible 👍

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
adejanovski: yes, I'd stay away from putting constraints here.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
adejanovski: I'm not knowledgeable enough in cryptography to know all ciphers that could be used here.
I'd rather leave the choice to users rather than put a fixed list of possible values.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -1041,3 +1049,106 @@ type TrackWarnings struct {
...
+	// +optional
+	Algorithm *string `json:"algorithm,omitempty"`
+
+	// +optional
+	StoreType *string `json:"store_type,omitempty"`
+
+	// +optional
+	CipherSuites *[]string `json:"cipher_suites,omitempty"`
+
+	// default: false
```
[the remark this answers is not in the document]
adejanovski: 👍

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
adejanovski: sure thing, I've removed pointers that were not absolutely necessary.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
adejanovski: actually, client to node encryption doesn't have a field that would use this enum, it's only used by server encryption.

adejanovski: I've added the enum kubebuilder annotation 👍

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adejanovski: sure thing

apis/reaper/v1alpha1/reaper_types.go
adejanovski: 👍

apis/reaper/v1alpha1/reaper_types.go
adejanovski: 👍

apis/stargate/v1alpha1/stargate_types.go
adejanovski: 👍

controllers/k8ssandra/cleanup.go
adejanovski: done 👍

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+	ctx context.Context,
+	remoteClient client.Client,
+	kc *api.K8ssandraCluster,
+	desiredConfig map[string]interface{},
+	userConfigMapContent,
+	namespace string,
+	logger logr.Logger,
+) result.ReconcileResult {
+	logger.Info("Reconciling Stargate Cassandra yaml configMap on namespace : " + namespace)
+	filteredCassandraConfig := filterYamlConfig(desiredConfig["cassandra-yaml"].(map[string]interface{}))
```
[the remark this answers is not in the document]
adejanovski: These values always need to be the same as what's configured in the Cassandra datacenter, otherwise Stargate won't be able to communicate with the nodes. Even when we'll expose the Cassandra settings in a structured manner, encryption options should not be exposed.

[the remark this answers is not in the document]
adejanovski: I've added a comment explaining this on the encryption stores structure of the CRD.

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+		separator := "\n"
+		if strings.HasSuffix(userConfigMapContent, "\n") {
+			separator = ""
+		}
+		configYamlString = userConfigMapContent + separator + configYamlString
+	}
+
+	configMapKey := client.ObjectKey{
+		Namespace: namespace,
+		Name:      stargate.CassandraConfigMap,
```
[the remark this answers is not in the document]
adejanovski: 👍

controllers/k8ssandra/stargate.go
```diff
@@ -192,3 +198,104 @@ func (r *K8ssandraClusterReconciler) removeStargateStatus(kc *api.K8ssandraClust
...
+
+	if err := remoteClient.Get(ctx, configMapKey, actualConfigMap); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Stargate configMap doesn't exist, creating it")
+			if err := remoteClient.Create(ctx, desiredConfigMap); err != nil {
+				logger.Error(err, "Failed to create Stargate ConfigMap")
+				return result.Error(err)
+			} else {
+				// Let's wait for the configMap to be created
+				return result.RequeueSoon(r.DefaultDelay)
```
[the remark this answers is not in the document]
adejanovski: sounds good

test/testdata/fixtures/client-encryption-secret.yaml
adejanovski: I simplified this to only use 2 secrets, containing both the store and the associated password.

pkg/reaper/deployment.go
adejanovski: After giving this a try, it turns out Cassandra won't allow providing no truststore, despite setting `require_client_auth` to `false`.
Apparently, the call to [this method](https://github.com/apache/cassandra/blob/bfab1fda882c368be0d2c9dad38cb5dbee501bfa/src/java/org/apache/cassandra/security/SSLFactory.java#L156) is made [here](https://github.com/apache/cassandra/blob/cassandra-3.11/src/java/org/apache/cassandra/security/SSLFactory.java#L61) with a hardcoded `true` for `buildTruststore`.
Cassandra then fails to start if the truststore path isn't provided.

pkg/stargate/deployments.go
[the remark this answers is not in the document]
adejanovski: This was moved to the Stargate controller now.

pkg/stargate/deployments.go
```diff
@@ -332,28 +338,62 @@ func computeHeapSize(template *api.StargateTemplate) resource.Quantity {
...
 				},
 			},
-		}}
+		},
+	})
+
+	return volumes
+}
+
+func computeEncryptionVolumes(template *api.StargateTemplate, spec api.StargateSpec) []corev1.Volume {
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/stargate/deployments.go
adejanovski: Right, I'm reworking that part to remove string manipulations.

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type EncryptionStoresYaml struct {
+	Keystore           string `json:"keystore,omitempty"`
+	KeystorePassword   string `json:"keystore_password,omitempty"`
+	Truststore         string `json:"truststore,omitempty"`
+	TruststorePassword string `json:"truststore_password,omitempty"`
+}
+
+type EncryptionStoresPasswords struct {
```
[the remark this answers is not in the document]
adutra: I think I see a better option: we can get rid of this struct and store the passwords directly in `DatacenterConfig`:

```go
type DatacenterConfig struct {
...
	ServerEncryptionStores   *encryption.Stores
	ClientEncryptionStores   *encryption.Stores
	ClientKeystorePassword   string
	ClientTruststorePassword string
	ServerKeystorePassword   string
	ServerTruststorePassword string
}
```

Then you can simplify `ReadEncryptionStoresSecrets`:

```go
func ReadEncryptionStoresSecrets(ctx context.Context, klusterKey types.NamespacedName, template *DatacenterConfig, remoteClient client.Client, logger logr.Logger) error {
...
		if password, err := ReadEncryptionStorePassword(...); err != nil {
			return err
		} else {
			template.ClientKeystorePassword = password
		}
// etc
```

This will make it very easy to create the volume mounts.

For the JSON payload, you'd need to amend `CreateJsonConfig` but it seems easy.

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+				logger.Info(fmt.Sprintf("Read truststore password %s", password))
+				encryptionStoresPasswords.ServerTruststorePassword = password
+			}
+		}
+	}
+
+	return encryptionStoresPasswords, nil
+}
+
+func ReadEncryptionStorePassword(ctx context.Context, namespace string, remoteClient client.Client, secretName string, storeName encryption.StoreName) (string, error) {
```
adutra, line 207: FYI I generally think that functions like this belong into controller code/package. Because all they do is take a client and use it to read Kubernetes objects. Same for `ReadEncryptionStoresSecrets`. These 2 functions have little to do with Cassandra configuration.

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+func addJmxEncryptionOptions(template *DatacenterConfig, encryptionStores encryption.Stores, encryptionStoresSecrets encryption.EncryptionStoresPasswords) {
+	addOptionIfMissing(template, "-Dcom.sun.management.jmxremote.ssl=true")
+	addOptionIfMissing(template, "-Dcom.sun.management.jmxremote.ssl.need.client.auth=true")
+	addOptionIfMissing(template, fmt.Sprintf("-Djavax.net.ssl.keyStore=%s/%s", StoreMountFullPath(encryption.StoreTypeClient, encryption.StoreNameKeystore), encryption.StoreNameKeystore))
+	addOptionIfMissing(template, fmt.Sprintf("-Djavax.net.ssl.trustStore=%s/%s", StoreMountFullPath(encryption.StoreTypeClient, encryption.StoreNameTruststore), encryption.StoreNameTruststore))
+	addOptionIfMissing(template, fmt.Sprintf("-Djavax.net.ssl.keyStorePassword=%s", encryptionStoresSecrets.ClientKeystorePassword))
+	addOptionIfMissing(template, fmt.Sprintf("-Djavax.net.ssl.trustStorePassword=%s", encryptionStoresSecrets.ClientTruststorePassword))
+}
+
+func addOptionIfMissing(template *DatacenterConfig, option string) {
```
adutra, line 227: This could maybe be turned into a utility function `utils.AddIfAbsent`.

pkg/cassandra/config.go
```diff
@@ -26,11 +27,23 @@ type config struct {
...
+}
+
+type ClientEncryptionOptionsYaml struct {
+	encryption.ClientEncryptionOptions `json:",inline"`
+	encryption.EncryptionStoresYaml    `json:",inline"`
+}
+
 // CassYamlIR is an internal representation of the cassandra.yaml. It is required because we want to make some options (esp. start_rpc) invisible to the user,
 // but some of those options still need to be rendered out into the final cassandra.yaml.
 type CassYamlIR struct {
```
adutra, line 42: Just FYI. I see that this type has been added recently. This file is becoming a bit messy and would need a bit of tidying. E.g. this type should not be exported. Also, the `config` type above was already meant as the "internal representation of the cassandra.yaml", so now we have a bit of a duplication here. But we can look into it in another PR. Just something to keep in mind.

pkg/cassandra/config.go
```diff
@@ -50,54 +63,75 @@ func isCassandra4(version string) bool {
...
 		}
 	}
 
-	return json.Marshal(&config)
+	return json.Marshal(&cassConfig)
 }
 
 // NewIRYaml returns a new Cassandra Yaml with all internal fields set to their mandatory values. For example, StartRpc must always be set false,
 // NewIRYaml ensures this happens.
 func NewIRYaml(cassAPIYaml api.CassandraYaml) CassYamlIR {
```
adutra, line 102: Just FYI. This function should not be exported either.

controllers/k8ssandra/stargate.go
```diff
@@ -105,7 +105,18 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
 }
 
 // TODO move to stargate package
-func (r *K8ssandraClusterReconciler) newStargate(stargateKey types.NamespacedName, kc *api.K8ssandraCluster, stargateTemplate *stargateapi.StargateDatacenterTemplate, actualDc *cassdcapi.CassandraDatacenter) *stargateapi.Stargate {
+func (r *K8ssandraClusterReconciler) newStargate(stargateKey types.NamespacedName, kc *api.K8ssandraCluster, stargateTemplate *stargateapi.StargateDatacenterTemplate, actualDc *cassdcapi.CassandraDatacenter, dcTemplate api.CassandraDatacenterTemplate) *stargateapi.Stargate {
+	cassandraEncryption := stargateapi.CassandraEncryption{}
+	if dcTemplate.CassandraConfig != nil {
+		if dcTemplate.CassandraConfig.CassandraYaml.ClientEncryptionOptions != nil && kc.Spec.Cassandra.ClientEncryptionStores != nil {
```
adutra, line 111: What if `ClientEncryptionOptions` is present but `ClientEncryptionOptions.Enabled = false`? Cf. the function `ClientEncryptionEnabled` – maybe you should use it here.

controllers/k8ssandra/stargate.go
```diff
@@ -105,7 +105,18 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
...
+			cassandraEncryption.ClientEncryptionStores = kc.Spec.Cassandra.ClientEncryptionStores
+		}
+
+		if dcTemplate.CassandraConfig.CassandraYaml.ServerEncryptionOptions != nil && kc.Spec.Cassandra.ServerEncryptionStores != nil {
```
adutra, line 115: Same here, what if encryption is `none`?

controllers/stargate/stargate_controller.go
```diff
@@ -66,7 +73,6 @@ func (r *StargateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			logger.Info("Stargate resource not found", "Stargate", req.NamespacedName)
 			return ctrl.Result{}, nil
 		} else {
-			logger.Error(err, "Failed to fetch Stargate", "Stargate", req.NamespacedName)
```
adutra, line 69: Why this change?

adutra (changes requested):

pkg/cassandra/encryption.go
adejanovski: done

pkg/cassandra/encryption.go
adejanovski: As we've seen, it's not the case :(

pkg/cassandra/encryption.go
adejanovski: Done

pkg/encryption/encryption_stores.go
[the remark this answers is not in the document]
adejanovski: I added the annotation in the end as DeepCopy was required.

pkg/cassandra/config.go
adejanovski: I don't fully understand what you're saying. Did you mean Cassandra 4?

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,21 @@ type StargateSpec struct {
 	// +optional
 	// +kubebuilder:default=true
 	Auth *bool `json:"auth,omitempty"`
+
+	CassandraEncryption *CassandraEncryption `json:"cassandraEncryption,omitempty"`
+}
+
+// Encryption settings are inherited from the CassandraDatacenter resource where Stargate is being deployed and do not need to be defined in the spec.
```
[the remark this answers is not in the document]
adejanovski: damn! it slipped through the cracks...

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,21 @@ type StargateSpec struct {
...
+// Still it is required to pass the encryption stores secrets to the Stargate pods, so that they can be mounted as volumes.
+type CassandraEncryption struct {
+	// Client encryption stores which are used by Cassandra and Reaper.
+	// These should not be set explicitly by the user, they will be copied from the K8ssandraCluster object.
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+
+// +kubebuilder:object:generate=true
+type Stores struct {
+	// ref to the secret that contains the keystore and its password
+	// the expected format of the secret is a "keystore" entry and a "keystore-password" entry
+	KeystoreSecretRef corev1.LocalObjectReference `json:"keystoreSecretRef"`
+
+	// ref to the secret that contains the truststore and its password
+	// the expected format of the secret is a "truststore" entry and a "truststore-password" entry
+	// +optional
```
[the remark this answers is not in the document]
adejanovski: not anymore indeed.

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// +kubebuilder:object:generate=true
+type Stores struct {
+	// ref to the secret that contains the keystore and its password
+	// the expected format of the secret is a "keystore" entry and a "keystore-password" entry
+	KeystoreSecretRef corev1.LocalObjectReference `json:"keystoreSecretRef"`
```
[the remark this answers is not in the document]
adejanovski: right

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// ref to the secret that contains the truststore and its password
+	// the expected format of the secret is a "truststore" entry and a "truststore-password" entry
+	// +optional
+	TruststoreSecretRef *corev1.LocalObjectReference `json:"truststoreSecretRef"`
+}
+
+type EncryptionStoresYaml struct {
```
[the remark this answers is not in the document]
adejanovski: I'm moving `EncryptionStoresYaml` to config.go but `EncryptionStoresPasswords` is used in the controllers as well and cannot be moved (needs to be exported anyway).

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+// +kubebuilder:object:generate=true
+type ServerEncryptionOptions struct {
+	// +kubebuilder:default=false
+	// +optional
+	Optional *bool `json:"optional,omitempty"`
+
+	// +kubebuilder:validation:Enum=none;dc;rack;all
+	// +kubebuilder:default=none
+	// +optional
+	InternodeEncryption string `json:"internode_encryption,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: I didn't do it because there was the assumption that we could use it somewhere else, which isn't the case.

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+	ServerTruststorePassword string
+}
+
+// See the cassandra.yaml file for explanations on how to set these options: https://github.com/apache/cassandra/blob/cassandra-4.0/conf/cassandra.yaml#L1091-L1183
+// +kubebuilder:object:generate=true
+type ClientEncryptionOptions struct {
+	Enabled bool `json:"enabled"`
+
+	// +optional
+	Optional bool `json:"optional,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: yes, because in 3.11, server encryption cannot be set to optional, it's a 4.x feature.
For client encryption, it exists in both versions.

pkg/encryption/encryption_stores.go
```diff
@@ -0,0 +1,102 @@
...
+
+	EncryptionSettings `json:",inline"`
```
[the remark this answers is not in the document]
adejanovski: gotcha

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	StoresMountPath = "/mnt"
+)
+
+// HandleEncryptionOptions sets up encryption in the datacenter config template.
+// The keystore and truststore config maps are mounted into the datacenter pod and the secrets are read to be set in the datacenter config template.
+func HandleEncryptionOptions(template *DatacenterConfig, encryptionStoresSecrets encryption.EncryptionStoresPasswords) error {
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+
+func StoreMountFullPath(storeType encryption.StoreType, storeName encryption.StoreName) string {
+	return fmt.Sprintf("%s/%s-%s", StoresMountPath, storeType, storeName)
+}
+
+func ClientEncryptionEnabled(template *DatacenterConfig) bool {
+	return template.CassandraConfig.CassandraYaml.ClientEncryptionOptions != nil && template.CassandraConfig.CassandraYaml.ClientEncryptionOptions.Enabled
+}
+
+func ServerEncryptionEnabled(template *DatacenterConfig) bool {
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+	secretKey := types.NamespacedName{Namespace: namespace, Name: secretName}
+	if err := remoteClient.Get(ctx, secretKey, clientKeystoreSecret); err != nil {
+		return "", err
+	}
+	password := string(clientKeystoreSecret.Data[fmt.Sprintf("%s-password", storeName)])
+	return password, nil
+}
+
+// Add JVM options required for turning on encryption
+func addJmxEncryptionOptions(template *DatacenterConfig, encryptionStores encryption.Stores, encryptionStoresSecrets encryption.EncryptionStoresPasswords) {
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		logger.Info("Client to node encryption is enabled, reading client encryption stores secrets")
+		// Read client keystore password
+		if password, err := ReadEncryptionStorePassword(ctx, klusterKey.Namespace, remoteClient, template.ClientEncryptionStores.KeystoreSecretRef.Name, encryption.StoreNameKeystore); err != nil {
+			return encryptionStoresPasswords, err
+		} else {
+			encryptionStoresPasswords.ClientKeystorePassword = password
+		}
+
+		encryptionStoresPasswords.ClientTruststorePassword = ""
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+
+		if password, err := ReadEncryptionStorePassword(ctx, klusterKey.Namespace, remoteClient, template.ServerEncryptionStores.KeystoreSecretRef.Name, encryption.StoreNameKeystore); err != nil {
+			return encryptionStoresPasswords, err
+		} else {
+			logger.Info(fmt.Sprintf("Read keystore password %s", password))
+			encryptionStoresPasswords.ServerKeystorePassword = password
+		}
+
+		// Read server truststore password
+		encryptionStoresPasswords.ServerTruststorePassword = ""
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/encryption.go
```diff
@@ -0,0 +1,234 @@
...
+		if template.ClientEncryptionStores.TruststoreSecretRef != nil {
```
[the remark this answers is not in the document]
adejanovski: ouch... that check was removed anyway, but still 😅

pkg/cassandra/config.go
adejanovski: 👍

controllers/k8ssandra/stargate.go
adejanovski: switched to using `ClientEncryptionEnabled`

controllers/k8ssandra/stargate.go
adejanovski: switched to using `ServerEncryptionEnabled`

controllers/stargate/stargate_controller.go
adejanovski: To comply with John's requirement of not adding error loggings when we already return the error, which will make it printed at a higher level.

pkg/cassandra/encryption.go
adejanovski: Noted.
It would involve quite a few changes here so I'd like to skip for this time and do this as part of a future refactoring.

pkg/cassandra/encryption.go
adejanovski: it does as it is now used in the controller as well.

pkg/cassandra/encryption.go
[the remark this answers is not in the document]
adejanovski: it does as it is now used in the controller as well.

pkg/cassandra/config.go
adutra: I was meaning that `optional` is defined in 4.0.1:

https://github.com/apache/cassandra/blob/cassandra-4.0.1/conf/cassandra.yaml#L1109

But not in 3.11.11:

https://github.com/apache/cassandra/blob/cassandra-3.11.11/conf/cassandra.yaml#L1045-L1057

But the code looks correct, I might have misread this.

pkg/encryption/encryption_stores.go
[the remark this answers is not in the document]
adutra: Yeah that's probably enough for now. I have the feeling that `EncryptionStoresPasswords`  could be removed too with a bit of refactoring, but we're in a hurry now.

pkg/encryption/encryption_stores.go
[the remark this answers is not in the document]
adutra: Note: as this involves a refactoring, we can do this in a follow-up task.

pkg/cassandra/encryption.go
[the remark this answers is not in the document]
adutra: Hmm where? I only see 2 call sites, both inside the `cassandra` package:

* HandleEncryptionOptions in github.com/k8ssandra/k8ssandra-operator/pkg/cassandra/encryption.go
* ReadEncryptionStoresSecrets in github.com/k8ssandra/k8ssandra-operator/pkg/cassandra/encryption.go

pkg/cassandra/encryption.go
adutra: Agreed 👍

controllers/stargate/stargate_controller.go
adutra: Ah I remember now. Thanks.

adutra:
I can't browse the files in this PR anymore :-( 

A few more remarks from my latest review:

* All encryption e2e tests are using C* 3.11. How about running the same tests on 4.0 as well?
* In test/testdata/fixtures/single-dc-reaper-encryption/k8ssandra.yaml line 37: the field `enabled` does not exist and should be removed.
* The file suite_test.go is already very large, how about moving the new encryption e2e tests to a dedicated file?

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,19 @@ type StargateSpec struct {
 	// +optional
 	// +kubebuilder:default=true
 	Auth *bool `json:"auth,omitempty"`
+
+	CassandraEncryption *CassandraEncryption `json:"cassandraEncryption,omitempty"`
+}
+
+// Still it is required to pass the encryption stores secrets to the Stargate pods, so that they can be mounted as volumes.
+type CassandraEncryption struct {
+	// Client encryption stores which are used by Cassandra and Reaper.
```
adutra, line 198:
```suggestion
	// Client encryption stores which are used by Cassandra and Stargate.
```

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -188,6 +189,19 @@ type StargateSpec struct {
 	// +optional
 	// +kubebuilder:default=true
 	Auth *bool `json:"auth,omitempty"`
+
+	CassandraEncryption *CassandraEncryption `json:"cassandraEncryption,omitempty"`
+}
+
+// Still it is required to pass the encryption stores secrets to the Stargate pods, so that they can be mounted as volumes.
```
adutra, line 196:
```suggestion
// Information required to pass the encryption stores secrets to the Stargate pods, so that they can be mounted as volumes.
```

pkg/cassandra/config.go
```diff
@@ -317,8 +387,8 @@ func AllowAlterRfDuringRangeMovement(dcConfig *DatacenterConfig) {
 
 // CreateJsonConfig parses dcConfig into a raw JSON base64-encoded string. If config is nil
 // then nil, nil is returned
-func CreateJsonConfig(config api.CassandraConfig, cassandraVersion string) ([]byte, error) {
-	cfg, err := newConfig(config, cassandraVersion)
+func CreateJsonConfig(template *DatacenterConfig) ([]byte, error) {
```
adutra, line 390: 👍 exactly what I had in mind.

pkg/cassandra/datacenter.go
```diff
@@ -86,23 +89,29 @@ func (r *Replication) ReplicationFactor(dc, ks string) int {
...
+	Racks                    []cassdcapi.Rack
+	CassandraConfig          api.CassandraConfig
+	AdditionalSeeds          []string
+	Networking               *cassdcapi.NetworkingConfig
+	Users                    []cassdcapi.CassandraUser
+	PodTemplateSpec          *corev1.PodTemplateSpec
+	MgmtAPIHeap              *resource.Quantity
+	ServerEncryptionStores   *encryption.Stores
+	ClientEncryptionStores   *encryption.Stores
+	ClientKeystorePassword   string
```
adutra, line 111: 👍 glad to see that it worked.

adutra (approved):
I'm going to approve this so that we can move forward. There are some minor things that could be improved IMO wrt API design and package boundaries, but the essential work is done. Congrats on this PR, it was a difficult one :-)

PR #319, opened 2022-01-25
StartRpc should be explicitly false by default  [K8SSAND-1189]

### LINKED ISSUE #315: K8SSAND-1189 ⁃ Thrift should be disabled for 3.11.x clusters
**What happened?**
#249 introduced a structured config for cassandra.yaml. `start_rpc` was left out which makes sense. Unfortunately cass-config-definitions sets it to `true` by default. It should be disabled.

I came across this while debugging some e2e test failures the other day. Tests previously were explicitly setting `start_rpc: false` to tune things down. That's no longer possible now. Maybe this is contributing to some of the recent test flakes.

Thrift should be turned off, or we should expose it and turn it off be default. Right now, it is enabled and the user cannot disable it. This should be fixed for 1.0.0.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1189) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1189
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Cass config builder has logic which sets `start_rpc` to true by default. The current config logic in k8ssandra-operator does not allow the user to set `start_rpc` (because it was considered deprecated), however propagation of the nil value leads to config builder enabling `start_rpc` by default.

This PR to explicitly set `start_rpc` false if not defined, or propagate the value; for Cassandra 3.11 clusters only.

Fixes #315

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
+++ b/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
@@ -863,6 +863,10 @@ type CassandraYaml struct {
 	// +optional
 	SstablePreemptiveOpenIntervalInMb *int `json:"sstable_preemptive_open_interval_in_mb,omitempty"`
 
+	// Exists in 3.11
+	//+optional
+	StartRpc *bool `json:"start_rpc,omitempty"`
+
 	// Exists in: 4.0, trunk
 	// +optional
 	StreamEntireSstables *bool `json:"stream_entire_sstables,omitempty"`
--- a/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
+++ b/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
@@ -863,10 +863,6 @@ type CassandraYaml struct {
 	// +optional
 	SstablePreemptiveOpenIntervalInMb *int `json:"sstable_preemptive_open_interval_in_mb,omitempty"`
 
-	// Exists in 3.11
-	//+optional
-	StartRpc *bool `json:"start_rpc,omitempty"`
-
 	// Exists in: 4.0, trunk
 	// +optional
 	StreamEntireSstables *bool `json:"stream_entire_sstables,omitempty"`
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -267,6 +267,7 @@ func filterConfigForVersion(cassandraVersion string, filteredConfig *api.Cassand
 		filteredConfig.CassandraYaml.RequestScheduler = nil
 		filteredConfig.CassandraYaml.RequestSchedulerId = nil
 		filteredConfig.CassandraYaml.RequestSchedulerOptions = nil
+		filteredConfig.CassandraYaml.StartRpc = nil
 		filteredConfig.CassandraYaml.StreamingSocketTimeoutInMs = nil
 		filteredConfig.CassandraYaml.ThriftFramedTransportSizeInMb = nil
 		filteredConfig.CassandraYaml.ThriftMaxMessageLengthInMb = nil
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -80,6 +80,11 @@ func (c config) MarshalJSON() ([]byte, error) {
 func newConfig(apiConfig api.CassandraConfig, cassandraVersion string) (config, error) {
 	// Filters out config element which do not exist in the Cassandra version in use
 	filteredConfig := apiConfig.DeepCopy()
+	if filteredConfig.CassandraYaml.StartRpc == nil {
+		// We need to set StartRpc false because start_rpc defaults true in the config builder.
+		f := false
+		filteredConfig.CassandraYaml.StartRpc = &f
+	}
 	filterConfigForVersion(cassandraVersion, filteredConfig)
 	cfg := config{CassandraYaml: filteredConfig.CassandraYaml, cassandraVersion: cassandraVersion}
 	err := validateConfig(filteredConfig)
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -20,10 +20,15 @@ const (
 // config is an internal type that is intended to be marshaled into JSON that is a valid
 // value for CassandraDatacenter.Spec.Config.
 type config struct {
-	api.CassandraYaml
+	CassYamlIR
 	cassandraVersion     string
 	jvmOptions           jvmOptions
 	additionalJvmOptions []string
+}
+
+type CassYamlIR struct {
+	api.CassandraYaml `json:",inline,omitempty"`
+	StartRpc          *bool `json:"start_rpc,omitempty"`
 }
 
 // jvmOptions is an internal type that is intended to be marshaled into JSON that is valid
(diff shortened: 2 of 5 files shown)


Miles-Garnsey:
@adejanovski assigning to you for review because you were the one originally behind the new config logic. 

Hoping to minimise load on you @jsanda ahead of the release but you're welcome to comment also.

pkg/cassandra/config.go
```diff
@@ -267,6 +272,7 @@ func filterConfigForVersion(cassandraVersion string, filteredConfig *api.Cassand
 		filteredConfig.CassandraYaml.RequestScheduler = nil
 		filteredConfig.CassandraYaml.RequestSchedulerId = nil
 		filteredConfig.CassandraYaml.RequestSchedulerOptions = nil
+		filteredConfig.CassandraYaml.StartRpc = nil
```
adejanovski, line 275: Is it equivalent to `false` ? 🤔

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -863,6 +863,10 @@ type CassandraYaml struct {
 	// +optional
 	SstablePreemptiveOpenIntervalInMb *int `json:"sstable_preemptive_open_interval_in_mb,omitempty"`
 
+	// Exists in 3.11
+	//+optional
+	StartRpc *bool `json:"start_rpc,omitempty"`
```
adejanovski, line 868: I would have liked that this is not exposed in the CRD, but maybe that's not possible due to how we're creating the JSON?

adutra: Agreed, I'd rather see this added internally to the json blurb that we feed into config-builder:

https://github.com/k8ssandra/k8ssandra-operator/blob/21332d9b6060cd1ea3159e5d1f1f664a1264126a/pkg/cassandra/config.go#L58

adutra: You'd probably need to marshal `c.CassandraYaml` to `map[string]interface{}` first, then add the `start_rpc=false` entry.

pkg/cassandra/config.go
Miles-Garnsey: It is nil, which I think means for Cassandra 4.0 it will not be included in the YAML. This is desirable because Cassandra 4.0 does not allow this parameter.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
Miles-Garnsey: Some customers are still using Thrift (yes, they should migrate off, but legacy applications are a thing). 

Is there some reason why we wouldn't want these customers to be able to migrate to k8ssandra? It doesn't seem to do us any harm to include an option to enable it.

Miles-Garnsey: NB: it is definitely technically possible to remove it from the CRD. Not a problem at all.

jsanda: >Some customers are still using Thrift (yes, they should migrate off, but legacy applications are a thing).

That's a fair point. I suspect though that those customers are probably not running in Kubernetes. Supporting legacy features increases the maintenance burden on us with a not so great risk/reward ratio.

Miles-Garnsey: > I suspect though that those customers are probably not running in Kubernetes.

I suspect the same, but neither of us have done a survey or have any data, so I don't think we should block their migration path based on our subjective mental image of a Thrift user. ([Image](https://twitter.com/QuinnyPig/status/1227135904839417857/photo/1) of likely Thrift user).

Maintenance: what maintenance would be required?  Worth noting that we are an open source project and take PRs. Anyone still using Thrift is welcome to contribute as they find things that are broken, I'd think?

adejanovski: @Miles-Garnsey, I was the one pushing to disable thrift altogether. While some users may still be using thrift (they have all my empathy), since it was removed in 4.0 and we only support Cassandra since 3.11, I don't see a lot of value in exposing the ability to start the RPC port.
We could reconsider this if we got a wave of folks wanting to use K8ssandra and Thrift, but at this stage I don't think we should spend time on this.
While maintenance could be minimal for such a thing, anything more in the code means tech debt. Removing this will also allow us to expose one less port in our pods and services (which should be a follow up ticket to this one).

jdonenine: I'm going to second what @adejanovski is saying here. We're in the position as a project that we're purposefully opinionated with what we provide. I think the overwhelming opinion of the community as a whole is that we're moving on from thrift. So we can take that opportunity here to simplify. 

It's worth noting that making the decision to omit the option here doesn't mean we couldn't add it later if we saw a strong demand for it, highly unlikely I'd think, but we've got that door. If we put it in now, we can't really unopen that box so to speak. 

Let's take the simpler path, we will not support thrift.

Miles-Garnsey: Closing this since it seems to be agreed that we are moving to disable Thrift completely. I'll comment on what that means in terms of additional changes below.

Miles-Garnsey:
Removing the ability to enable Thrift means that we will need to create a new internal representation of the cassandra.yaml type, because the existing process is to directly serialise the `api.CassandraYaml` type. We need that type not to contain a `StartRpc` field, so a new type is needed.

My commit immediately above does all of this, just waiting for tests to run to see if it all works the way I'd expect.

Miles-Garnsey:
Results of test run:

* `CreateMultiReaper` is failing. I think this is related to the issue fixed in PR #321.
* `ClusterScoped/MultiDcMultiCluster` is passing. Normally this fails with an error relating to the slash in the test name. I don't understand why it is passing here. There appear to be error messages in the test run too. 
*  `CreateMultiStargateAndDatacenter` is [failing](https://github.com/k8ssandra/k8ssandra-operator/runs/4999948818?check_suite_focus=true#logs). I haven't seen this error before, but I suspect a timing issue so I'm re-running it.

I think PR #321 should be merged before we merge this one, to get `CreateMultiReaper` passing. 

I will raise another PR to remove the slashes from `ClusterScoped/MultiDcMultiCluster` so that we have consistent results there (but that shouldn't delay merging this PR as it is passing anyway). 

I will wait to see if `CreateMultiStargateAndDatacenter` passes when I re-run it, and raise an issue if it isn't passing/failing consistently.

jsanda:
>I will raise another PR to remove the slashes from ClusterScoped/MultiDcMultiCluster so that we have consistent results there (but that shouldn't delay merging this PR as it is passing anyway).

Go adds the `/` with subtests. GHA doesn't doesn't allow `/` in the artifact name, so uploading artifacts fails. Of course, there is another failure in the actual test; otherwise artifacts don't get uploaded.

I was investigating an e2e failure last Friday and saw this in the Cassandra logs,

```
Caused by: io.netty.channel.ChannelException: eventfd() failed: Too many open files
...
ERROR [pool-2-thread-1] 2022-01-28 13:12:13,300 LifecycleResources.java:140 - Error checking pid
java.lang.RuntimeException: java.io.IOException: Cannot run program "/bin/bash": error=24, Too many open files
```
I hope disabling thrift again will get us some more stability. 

There are definitely some other flakes though :(

Miles-Garnsey:
> Go adds the / with subtests. GHA doesn't doesn't allow / in the artifact name, so uploading artifacts fails. Of course, there is another failure in the actual test; otherwise artifacts don't get uploaded.

@jsanda thanks, it didn't click for me that it was the artefact upload process that was causing the problem. I'll raise a new ticket about this and move discussion there.

`MultiDcAuthOnOff` is still failing, but also fails on main. All other tests are now passing. Who needs to approve this?

adejanovski (approved):

PR #320, opened 2022-01-25
reset image tag for development

jsanda:
This resets the image tag in `config/deployments/default/kustomization.yaml` back to latest. In the future, releases should be done out of branches so we can avoid the changes in main.


burmanm (approved):

jsanda:
I am going to go ahead and merge since the `MutiDcAuthOnOff` test has been flaky. It takes a long time to complete so I don't want to wait for a retry.

PR #322, opened 2022-01-25
Remove Reaper DC template

### LINKED ISSUE #308: K8SSAND-1177 ⁃ Consider removing per-DC Reaper templates

adutra:
**What this PR does**:
Removes Reaper DC template, instead offers two deployment modes for K8ssandraClusters: each DC with availability EACH, or just the first one with availability ALL.

Note that this is based on #312. 

Fixes #308 

**Checklist**
- [ ] Changes manually tested


apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -149,33 +175,12 @@ type AutoScheduling struct {
...
-	// and "password".
-	// +optional
-	JmxUserSecretRef corev1.LocalObjectReference `json:"jmxUserSecretRef,omitempty"`
-
-	// Defines the secret which contains the username and password for the Reaper UI and REST API authentication.
-	// +optional
-	UiUserSecretRef corev1.LocalObjectReference `json:"uiUserSecretRef,omitempty"`
+	// +kubebuilder:default="PER_DC"
+	// +kubebuilder:validation:Enum:=PER_DC;SINGLE
+	DeploymentMode string `json:"deploymentMode,omitempty"`
```
adutra, line 183: This does not offer users the ability to chose the DC where Reaper is going to be deployed in SINGLE mode. We could add a `SingleDeploymentDcRef` field for that.

adejanovski: Is it going to be the first DC each time? It could probably be enough this way as long as we can predict which DC is going to get Reaper.
Also, what will happen if the DC that has Reaper gets decommissioned?

adejanovski (approved):
It's much simpler this way. Very nice! Who doesn't love a PR that removes 2k+ lines and only adds 200? 😄 
I have a non blocking comment on which DC will get Reaper and how it behaves during decommission.
Also, the tests didn't run so we're a little blind on the impacts, but I guess we can see that in the follow up PR as this one doesn't target `main`. Right?

apis/reaper/v1alpha1/reaper_types.go
adutra: > Is it going to be the first DC each time?

Yes currently it is always the first DC in the `kc.Spec.Cassandra.Datacenters` list that gets Reaper deployed.

> what will happen if the DC that has Reaper gets decommissioned?

We don't have that feature yet, but presumably, the DC will be removed from the list, another DC will become the "first" one in the list, and the controller will detect that Reaper should be deployed there, and will make the necessary changes.

PR #328, opened 2022-01-27
add support for removing a DC from a cluster

### LINKED ISSUE #284: K8SSAND-1154 ⁃ Support decommissioning of a datacenter from a K8ssandraCluster
Support for the decomissioning of a datacenter has been added in Cass Operator and needs to be built upon at the K8ssandra Operator level to support the operation across the full mulit-datacenter deployment.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-1154) by [Unito](https://www.unito.io)
┆epic: Decomission Datacenters
┆friendlyId: K8SSAND-1154
┆priority: Medium

jsanda:
**What this PR does**:
See https://github.com/k8ssandra/k8ssandra-operator/issues/284#issuecomment-1018717590 for detailed description.

Fixes #284

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/cleanup.go
```diff
@@ -119,3 +123,193 @@ func (r *K8ssandraClusterReconciler) checkFinalizer(ctx context.Context, kc *api
...
+	}
+
+	if dc != nil {
+		if dc.GetConditionStatus(cassdcapi.DatacenterDecommission) == corev1.ConditionTrue {
+			logger.Info("CassandraDatacenter decommissioning in progress", "CassandraDatacenter", kcKey)
+			// There is no need to requeue here. Reconciliation will be trigger by updates made by cass-operator.
+			return result.Done()
+		}
+
+		if err = remoteClient.Delete(ctx, dc); err != nil && !errors.IsNotFound(err) {
```
burmanm, line 189: If Datacenter is deleted, the opt-in annotation should be set first (if we want to decommission). And if we want to decommission, shouldn't this requeue and wait for it to finish decommissioning before you delete it from the Status? It could be part of the cluster for a while until it has finished decommissioning. The instant delete from K8ssandraCluster could give the user indication that the process has finished and they could shutdown the target Kubernetes cluster for example (thus ruining the decommission). Not to mention, it's not guaranteed the decommission is able to start instantly.

go.mod
```diff
@@ -42,6 +42,7 @@ require (
 )
 
 replace (
+	github.com/k8ssandra/cass-operator => github.com/burmanm/cass-operator v1.6.1-0.20220118133905-b20824af6b5d
```
burmanm, line 45: Should probably target master and not my branch.

test/e2e/suite_test.go
```diff
@@ -741,6 +745,124 @@ func addDcToCluster(t *testing.T, ctx context.Context, namespace string, f *fram
...
+		assert.Eventually(func() bool {
+			output, err := f.ExecuteCql(ctx, k8sCtx0, namespace, "test", "test-dc1-default-sts-0",
+				fmt.Sprintf("SELECT replication FROM system_schema.keyspaces WHERE keyspace_name = '%s'", ks))
+			if err != nil {
+				t.Logf("replication check for keyspace %s failed: %v", ks, err)
+				return false
+			}
+			return strings.Contains(output, "'dc1': '1'") && !strings.Contains(output, "'dc2': '1'")
+		}, 1*time.Minute, 5*time.Second, "failed to verify replication updated for keyspace %s", ks)
+	}
```
burmanm, line 863: This test doesn't really test if the Datacenter was actually removed from the cluster, only if replication was modified. Thus, it does not catch the situation that the dc is infact not decommissioned but is just down in the ring (which is what happens in reality in this test).

jsanda: I can add another `checkNodeToolStatusUN` after this to make sure the expected number of nodes is reported.

controllers/k8ssandra/cleanup.go
jsanda: I'll add the annotation.

As for waiting for the decommission to complete, immediately before the `Delete` there is a check,

```go
		if dc.GetConditionStatus(cassdcapi.DatacenterDecommission) == corev1.ConditionTrue {
			logger.Info("CassandraDatacenter decommissioning in progress", "CassandraDatacenter", kcKey)
			// There is no need to requeue here. Reconciliation will be trigger by updates made by cass-operator.
			return result.Done()
		}

		if err = remoteClient.Delete(ctx, dc); err != nil && !errors.IsNotFound(err) {
			return result.Error(fmt.Errorf("failed to delete CassandraDatacenter (%s): %v", dcName, err))
		}
		logger.Info("Deleted CassandraDatacenter", utils.GetKey(dc))
```
Should I be checking differently?

jsanda: I see it now :) I am missing a requeue.

controllers/k8ssandra/schemas.go
```diff
@@ -66,6 +81,10 @@ func (r *K8ssandraClusterReconciler) checkSchemaAgreement(mgmtApi cassandra.Mana
 		return result.Continue()
 	}
 
+	if _, found := versions["UNREACHABLE"]; len(versions) == 2 && found {
```
burmanm, line 84: This could use a comment, I can't make out from the code the reason for this check (why the magic word unreachable, why len check etc)

jsanda: @adutra has a better version of this function in #331. I'll update the impl in my branch accordingly.

burmanm (approved):

PR #331, opened 2022-01-27
Expose the stopped property from the CassandraDatacenter spec [K8SSAND-1168]

### LINKED ISSUE #296: K8SSAND-1168 ⁃ Expose the stopped property from the CassandraDatacenter spec
The CassandraDatacenter spec has a `stopped` property which will scale down all StatefulSets to zero but retain all PVCs. This can be helpful in a couple different scenarios. First, it can help reduce costs. If you are running a cluster in a cloud provider and not actively using the cluster, you can set `stopped` to reduce the amount of cloud resources being used. Secondly there are times when it is necessary to set `stopped` in order to recover from a bad configuration.

Note that exposing the property in the K8ssandraCluster spec is only part of the work involved. We have lots of logic based on whether the CassandraDatacenter is ready as determined by the `Ready` condition. We will need to visit each of these places in the code and see if/how they should handle the case of the CassandraDatacenter being stopped. 

For example, let's say we have a K8ssandraCluster with 2 DCs - dc1 and dc2. We decide to deploy Stargate in both DCs. The operator first reconciles all DCs, and only after they are ready will it proceed to deploy Stargate. Now let's say the user has stopped dc1 and wants to deploy Stargate to dc2. We probably want to proceed and deploy Stargate.

This should be done for 1.0.0.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1168) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1168
┆priority: Medium

adutra:
**What this PR does**:
Expose the stopped property from the CassandraDatacenter spec.

Fixes #296 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/datacenters.go
+++ b/controllers/k8ssandra/datacenters.go
@@ -122,6 +122,12 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 					logger.Error(err, "SuperuserSecretName is immutable, reverting to existing value in CassandraDatacenter")
 				}
 
+				if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) && desiredDc.Spec.Stopped {
+					desiredDc.Spec.Stopped = false
+					err = fmt.Errorf("tried to stop a datacenter that is being rebuilt")
+					logger.Error(err, "Stopped cannot be set to true until the CassandraDatacenter is fully rebuilt")
+				}
+
 				if err := cassandra.ValidateConfig(desiredDc, actualDc); err != nil {
 					return result.Error(fmt.Errorf("invalid Cassandra config: %v", err)), actualDcs
 				}
@@ -136,27 +142,42 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 				}
 			}
 
-			if !cassandra.DatacenterReady(actualDc) {
-				logger.Info("Waiting for datacenter to become ready")
-				return result.RequeueSoon(r.DefaultDelay), actualDcs
+			if actualDc.Spec.Stopped {
+				if !cassandra.DatacenterStopped(actualDc) {
+					logger.Info("Waiting for datacenter to satisfy Stopped condition")
+					return result.Done(), actualDcs
+				}
+			} else {
+				if !cassandra.DatacenterReady(actualDc) {
+					logger.Info("Waiting for datacenter to satisfy Ready condition")
+					return result.Done(), actualDcs
+				}
 			}
 
-			logger.Info("The datacenter is ready")
+			logger.Info("The datacenter is reconciled")
 
 			actualDcs = append(actualDcs, actualDc)
 
-			if recResult := r.checkSchemas(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
-				return recResult, actualDcs
-			}
+			if !actualDc.Spec.Stopped {
 
-			if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) {
-				if recResult := r.reconcileDcRebuild(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
+				if recResult := r.checkSchemas(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
 					return recResult, actualDcs
 				}
+
+				if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) {
+					if recResult := r.reconcileDcRebuild(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
+						return recResult, actualDcs
+					}
+				}
+
 			}
 
 		} else {
 			if errors.IsNotFound(err) {
+				if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) && desiredDc.Spec.Stopped {
+					err := fmt.Errorf("cannot add a datacenter in stopped state to an existing cluster")
+					return result.Error(err), actualDcs
+				}
 				// cassdc doesn't exist, we'll create it
 				if err = remoteClient.Create(ctx, desiredDc); err != nil {
 					logger.Error(err, "Failed to create datacenter")
--- a/controllers/k8ssandra/k8ssandracluster_controller.go
+++ b/controllers/k8ssandra/k8ssandracluster_controller.go
@@ -160,7 +160,7 @@ func (r *K8ssandraClusterReconciler) afterCassandraReconciled(ctx context.Contex
 			return recResult
 		} else if recResult := r.reconcileStargate(ctx, kc, dcTemplate, dc, logger, remoteClient); recResult.Completed() {
 			return recResult
-		} else if recResult := r.reconcileReaper(ctx, kc, dcTemplate, dc, i, logger, remoteClient); recResult.Completed() {
+		} else if recResult := r.reconcileReaper(ctx, kc, dcTemplate, dc, logger, remoteClient); recResult.Completed() {
 			return recResult
 		}
 	}
--- a/controllers/k8ssandra/schemas.go
+++ b/controllers/k8ssandra/schemas.go
@@ -157,7 +157,7 @@ func (r *K8ssandraClusterReconciler) updateReplicationOfSystemKeyspaces(
 		return recResult
 	}
 
-	datacenters := cassandra.GetDatacentersForReplication(kc)
+	datacenters := cassandra.GetDatacentersForSystemReplication(kc)
 	replication := cassandra.ComputeReplicationFromDcTemplates(3, datacenters...)
 
 	logger.Info("Preparing to update replication for system keyspaces", "replication", replication)
@@ -313,7 +313,7 @@ func getInternalKeyspaces(kc *api.K8ssandraCluster) []string {
 // been deployed. The replication argument may include DCs that have not yet been deployed.
 func getReplicationForDeployedDcs(kc *api.K8ssandraCluster, replication *cassandra.Replication) *cassandra.Replication {
 	dcNames := make([]string, 0)
-	for _, template := range cassandra.GetDatacentersForReplication(kc) {
+	for _, template := range kc.GetInitializedDatacenters() {
 		dcNames = append(dcNames, template.Meta.Name)
 	}
 
--- a/controllers/k8ssandra/schemas.go
+++ b/controllers/k8ssandra/schemas.go
@@ -157,7 +157,7 @@ func (r *K8ssandraClusterReconciler) updateReplicationOfSystemKeyspaces(
 		return recResult
 	}
 
-	datacenters := cassandra.GetDatacentersForSystemReplication(kc)
+	datacenters := cassandra.GetDatacentersForReplication(kc)
 	replication := cassandra.ComputeReplicationFromDcTemplates(3, datacenters...)
 
 	logger.Info("Preparing to update replication for system keyspaces", "replication", replication)
@@ -313,7 +313,7 @@ func getInternalKeyspaces(kc *api.K8ssandraCluster) []string {
 // been deployed. The replication argument may include DCs that have not yet been deployed.
 func getReplicationForDeployedDcs(kc *api.K8ssandraCluster, replication *cassandra.Replication) *cassandra.Replication {
 	dcNames := make([]string, 0)
-	for _, template := range kc.GetInitializedDatacenters() {
+	for _, template := range cassandra.GetDatacentersForReplication(kc) {
 		dcNames = append(dcNames, template.Meta.Name)
 	}
 
--- a/controllers/k8ssandra/stargate.go
+++ b/controllers/k8ssandra/stargate.go
@@ -187,7 +187,7 @@ func (r *K8ssandraClusterReconciler) reconcileStargateAuthSchema(
 		return recResult
 	}
 
-	datacenters := cassandra.GetDatacentersForReplication(kc)
+	datacenters := kc.GetInitializedDatacenters()
 	replication := cassandra.ComputeReplicationFromDcTemplates(3, datacenters...)
 
 	if err := stargate.ReconcileAuthKeyspace(mgmtApi, replication, logger); err != nil {
--- a/controllers/k8ssandra/stargate.go
+++ b/controllers/k8ssandra/stargate.go
@@ -187,7 +187,7 @@ func (r *K8ssandraClusterReconciler) reconcileStargateAuthSchema(
 		return recResult
 	}
 
-	datacenters := kc.GetReadyDatacenters()
+	datacenters := cassandra.GetDatacentersForReplication(kc)
 	replication := cassandra.ComputeReplicationFromDcTemplates(3, datacenters...)
 
 	if err := stargate.ReconcileAuthKeyspace(mgmtApi, replication, logger); err != nil {
--- a/controllers/k8ssandra/stargate.go
+++ b/controllers/k8ssandra/stargate.go
@@ -37,6 +37,11 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
 	actualStargate := &stargateapi.Stargate{}
 	logger = logger.WithValues("Stargate", stargateKey)
 
+	if actualDc.Spec.Stopped && stargateTemplate != nil {
+		logger.Info("DC is stopped: skipping Stargate deployment")
+		stargateTemplate = nil
+	}
+
 	if stargateTemplate != nil {
 		logger.Info("Reconcile Stargate")
 		desiredStargate := r.newStargate(stargateKey, kc, stargateTemplate, actualDc, dcTemplate, logger)
--- a/controllers/k8ssandra/stop_dc_test.go
+++ b/controllers/k8ssandra/stop_dc_test.go
@@ -199,10 +199,10 @@ func stopExistingDc(t *testing.T, f *framework.Framework, ctx context.Context, k
 	}, timeout, interval, "timed out waiting for dc condition check")
 
 	t.Log("check that stargate sg1 was deleted")
-	verifyObjectDoesNotExist(ctx, t, f, sg1Key, &stargateapi.Stargate{})
+	f.AssertObjectDoesNotExist(ctx, t, sg1Key, &stargateapi.Stargate{}, timeout, interval)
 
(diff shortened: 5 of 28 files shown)


controllers/k8ssandra/datacenters.go
```diff
@@ -126,23 +126,39 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
+			if !actualDc.Spec.Stopped {
 
-			if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) {
-				if recResult := r.reconcileDcRebuild(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
+				if recResult := r.checkSchemas(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
 					return recResult, actualDcs
 				}
+
+				if annotations.HasAnnotationWithValue(kc, api.RebuildDcAnnotation, dcKey.Name) {
+					if recResult := r.reconcileDcRebuild(ctx, kc, actualDc, remoteClient, logger); recResult.Completed() {
```
adutra, line 157: @jsanda if a dc is added in stopped state, should we call `reconcileDcRebuild` for it?

burmanm: We should throw a validation error in that case. cass-operator will not finish building the new DC if it's in stopped state.

adutra: Hm that's crazy indeed. I just created a new cassdc in stopped state, and cass-operator happily created the service and even the empty statefulsets for me. No validation errors. Conditions Initialized and Stopped are true. This is not going to work, I will add a validation error to forbid creating a dc in stopped state.

controllers/k8ssandra/stop_dc_test.go
```diff
@@ -0,0 +1,362 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+// stopDc tests scenarios that involve stopping a CassandraDatacenter.
+func stopDc(t *testing.T, ctx context.Context, f *framework.Framework, _ string) {
+	t.Run("StopExistingDatacenter", stopDcTest(f, ctx, stopExistingDc))
+	t.Run("AddAndStopDatacenter", stopDcTest(f, ctx, addAndStopDc))
```
adutra, line 23: I'm not sure this test is useful, I might remove it. There is nothing particular about stopping a dc that wasn't part of the initial cluster.

controllers/k8ssandra/datacenters.go
```diff
@@ -125,23 +125,39 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 				}
 			}
 
-			if !cassandra.DatacenterReady(actualDc) {
-				logger.Info("Waiting for datacenter to become ready")
+			if actualDc.Status.CassandraOperatorProgress != cassdcapi.ProgressReady {
```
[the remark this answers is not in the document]
adutra: But we were already doing this. Prior to this change we would call `DatacenterReady` here. It is implemented as follows:

```go
func DatacenterReady(dc *cassdcapi.CassandraDatacenter) bool {
	return dc.GetConditionStatus(cassdcapi.DatacenterReady) == corev1.ConditionTrue && dc.Status.CassandraOperatorProgress == cassdcapi.ProgressReady
}
```

The only reason why I extracted this comparison out of `DatacenterReady` is because there are two functions, `DatacenterStopped` and `DatacenterStopping` that differ only by the handling of `CassandraOperatorProgress`. 

Do you advocate instead for changing how `DatacenterReady` is implemented? I wanted to avoid this.

controllers/k8ssandra/datacenters.go
```diff
@@ -125,23 +125,39 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
+				logger.Info("Waiting for datacenter to finish updating")
 				return result.RequeueSoon(r.DefaultDelay), actualDcs
```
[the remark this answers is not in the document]
adutra: OK I will replace `result.RequeueSoon(r.DefaultDelay)` with `result.Done()` – even if this wasn't changed by the current PR.

controllers/k8ssandra/datacenters.go
```diff
@@ -125,27 +131,47 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
 			}
 
-			logger.Info("The datacenter is ready")
+			if actualDc.Spec.Stopped {
+				if !cassandra.DatacenterStopped(actualDc) {
+					logger.Info("Waiting for datacenter to satisfy Stopped condition")
+					return result.RequeueSoon(r.DefaultDelay), actualDcs
```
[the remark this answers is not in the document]
adutra: > And if the datacenter is stopped, can't we just move on to the next one?

Here we are waiting for the Stopped condition to be true, since that's what we want. When it becomes true, we would move to the next dc.

controllers/k8ssandra/schemas.go
```diff
@@ -62,6 +62,14 @@ func (r *K8ssandraClusterReconciler) checkSchemaAgreement(mgmtApi cassandra.Mana
 		return result.Error(err)
 	}
 
+	for uid := range versions {
+		// a key named UNREACHABLE may appear in the results when nodes are unreachable.
+		// We exclude these keys from the check.
+		if uid == "UNREACHABLE" {
```
adutra, line 68: I had to introduce this to allow the operator to make progress when a dc is stopped.

When any node is unreachable, the management api returns a map that looks like below:

```json
{
  "UNREACHABLE": [
    "172.18.0.2"
  ],
  "aa573028-7c70-3b8f-a247-13bbca2011d2": [
    "172.18.0.9",
    "172.18.0.6"
  ]
}
```

If you don't exclude the UNREACHABLE entry, this method will always return an error and the operator will get stuck.

Note: the output of `nodetool describecluster` is similar, so it's not a problem with the management API.

test/e2e/stargate_apis_test.go
```diff
@@ -256,7 +265,7 @@ func createKeyspaceAndTableNative(t *testing.T, connection *client.CqlClientConn
 	))
 	require.IsType(t, &message.SchemaChangeResult{}, response.Body.Message, "Expected CREATE TABLE response to be of type SchemaChangeResult")
 	response = sendQuery(t, connection, fmt.Sprintf("TRUNCATE %s.%s", keyspaceName, tableName))
-	assert.IsType(t, &message.VoidResult{}, response.Body.Message, "Expected TRUNCATE response to be of type VoidResult")
+	require.IsType(t, &message.VoidResult{}, response.Body.Message, "Expected TRUNCATE response to be of type VoidResult")
```
adutra, line 268: Just fixing a few panics I've seen while working on this PR.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
jsanda: I might be in favor of changing `DatacenterReady`. It doesn't have to be done in this PR (if it is done at all).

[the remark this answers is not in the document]
adutra: OK I removed the if block and came back to something closer to what we have in main. We can then see later to change DatacenterReady, but it is called in many places, so definitely something we want to do in another PR.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -160,7 +160,7 @@ func (r *K8ssandraClusterReconciler) afterCassandraReconciled(ctx context.Contex
 			return recResult
 		} else if recResult := r.reconcileStargate(ctx, kc, dcTemplate, dc, logger, remoteClient); recResult.Completed() {
```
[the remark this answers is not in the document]
jsanda: I am inclined to say we do nothing. There won't be any pods matching the selector so I would assume Prometheus won't try to scrape anything.

jsanda (changes requested):
I did some manual testing and it reminded me of some additional changed that I forgot to include in the issue. When I stop a DC, it gets removed from the replication strategy for all keyspaces. We shouldn't be removing replicas.

`reconcileStargateAuthSchema` and `reconcileReaperSchema` both call `kc.GetReadyDatacenters` followed by `cassandra.ComputeReplicationFromDcTemplates`. Stopped DCs should be included here when computing replication. 

Similarly for system keyspaces, `cassandra.GetDatacentersForSystemReplication` should be updated to include stopped as well as ready DCs.

We do update replication of user-defined keyspaces during a rebuild. I checked that code. I don't think it needs any changes because it only checked if the `Initialized` condition is set on the CassandraDatacenter.

Lastly, tests should be updated to make sure replication changes are not made.

adutra:
> When I stop a DC, it gets removed from the replication strategy for all keyspaces. We shouldn't be removing replicas.

By not removing the stopped dc from replication strategies, some CLs like QUORUM would start failing. Is that OK in your opinion?

> Similarly for system keyspaces, cassandra.GetDatacentersForSystemReplication should be updated to include stopped as well as ready DCs.

👍 

> reconcileStargateAuthSchema and reconcileReaperSchema both call kc.GetReadyDatacenters followed by cassandra.ComputeReplicationFromDcTemplates. Stopped DCs should be included here when computing replication.

In this case is it enough to replace `kc.GetReadyDatacenters` with `cassandra.GetDatacentersForSystemReplication`? It seems it's going to be the same algorithm.

controllers/k8ssandra/stargate.go
```diff
@@ -36,6 +36,11 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
 	actualStargate := &stargateapi.Stargate{}
 	logger = logger.WithValues("Stargate", stargateKey)
 
+	if actualDc.Spec.Stopped && stargateTemplate != nil {
```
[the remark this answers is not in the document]
adutra: I've asked myself this question too. I believe a Stargate node can operate normally even if the local DC is down. It could be a good thing to keep it up and running.

adutra:
@jsanda I believe I addressed all your comments.

Regarding `GetDatacentersForSystemReplication` I believe it can be used for all keyspaces, in all situations. I updated the function description. Let me know if that makes sense or if I've gone completely nuts.

controllers/k8ssandra/k8ssandracluster_controller.go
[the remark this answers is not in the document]
Miles-Garnsey: Sorry, I didn't spot this earlier. @jsanda is right, nothing will be scraped. 

From what I can understand `stopped` is intended as a temporary state and we expect the DC to eventually start again. On that basis, I agree with John in saying that nothing should happen with telemetry.

pkg/cassandra/util.go
```diff
@@ -33,35 +33,39 @@ func DatacenterStopping(dc *cassdcapi.CassandraDatacenter) bool {
...
-			if found && status.Cassandra.GetConditionStatus(cassdcapi.DatacenterReady) == corev1.ConditionTrue {
-				datacenters = append(datacenters, dc)
+// GetDatacentersForReplication determines the DCs that should be included for replication. This function works for both
+// system and non-system keyspaces. Replication for system keyspaces is initially done before the cluster has been
+// initialized, and is set through the management-api, not CQL. This allows us to specify non-existent DCs for
+// replication even though Cassandra 4 does not allow that. Once the cluster is initialized however, replication is done
+// through CQL, and we cannot include non-existing DCs anymore; this is going to be the case: 1) when updating system
+// keyspaces; 2) when configuring keyspaces for Stargate and Reaper; and 3) when updating user keyspaces during a DC
+// rebuild.
+func GetDatacentersForReplication(kc *api.K8ssandraCluster) []api.CassandraDatacenterTemplate {
```
jsanda, line 43: I don't think combining the functions is going to work. We do the schema checks after reconciling each DC. Let's say we are creating a cluster with 3 DCs - dc1, dc2, dc3. The `CassandraInitialized` condition won't get set until all three DCs are ready, but the schema checks happen after we finish reconciling each DC (i.e., each DC is ready or stopped). With this change we will include non-existing DCs.

I did a quick test to verify. I created this cluster:

```yaml
kind: K8ssandraCluster
metadata:
  name: test
spec:
  cassandra:
    serverVersion: "4.0.1"
    storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
    config:
      cassandraYaml:
        auto_snapshot: false
        memtable_flush_writers: 1
        commitlog_segment_size_in_mb: 2
        concurrent_compactors: 1
        compaction_throughput_mb_per_sec: 0
        sstable_preemptive_open_interval_in_mb: 0
        key_cache_size_in_mb: 0
        thrift_prepared_statements_cache_size_mb: 1
        prepared_statements_cache_size_mb: 1
        slow_query_log_timeout_in_ms: 0
        counter_cache_size_in_mb: 0
        concurrent_reads: 2
        concurrent_writes: 2
        concurrent_counter_writes: 2
      jvmOptions:
        heapSize: 512Mi
    networking:
      hostNetwork: true
    resources:
      limits:
        memory: 1024Mi
    datacenters:
      - metadata:
          name: dc1
        k8sContext: kind-k8ssandra-0
        size: 1
      - metadata:
          name: dc2
        k8sContext: kind-k8ssandra-1
        size: 1
  stargate:
    size: 1
    heapSize: 512Mi
    resources:
      limits:
        memory: 1024Mi
  reaper:
    autoScheduling:
      enabled: true
```
And as expected the schema checks for Stargate fail. Here is the error from Cassandra's logs:

```
INFO  [epollEventLoopGroup-7-3] 2022-02-04 03:52:22,178 RpcExecutionException.java:39 - Failed to execute method NodeOps.createKeyspace
org.apache.cassandra.exceptions.ConfigurationException: Unrecognized strategy option {dc2} passed to NetworkTopologyStrategy for keyspace data_endpoint_auth
```

jsanda (changes requested):

adutra:
Putting this back to draft. It's becoming way more complex than I initially thought. I might give up on finishing this by the way.

Also the latest rebase broke the e2e test. Reaper is not able anymore to start with one DC dead:

```
INFO   [2022-02-04 21:19:32,966] [main] c.d.d.c.Cluster - New Cassandra host cluster1-dc2-service/172.18.0.7:9042 added
com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: cluster1-dc2-service/172.18.0.7:9042 (com.datastax.driver.core.exceptions.UnavailableException: Not enough replicas available for query at consistency QUORUM (2 required but only 1 alive)))
	at com.datastax.driver.core.exceptions.NoHostAvailableException.copy(NoHostAvailableException.java:83)
	at com.datastax.driver.core.exceptions.NoHostAvailableException.copy(NoHostAvailableException.java:37)
	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:35)
	at com.datastax.driver.core.DefaultResultSetFuture.getUninterruptibly(DefaultResultSetFuture.java:293)
	at com.datastax.driver.core.AbstractSession.execute(AbstractSession.java:58)
	at org.cognitor.cassandra.migration.Database.getVersion(Database.java:195)
	at io.cassandrareaper.storage.CassandraStorage.initializeCassandraSchema(CassandraStorage.java:275)
	at io.cassandrareaper.storage.CassandraStorage.initializeAndUpgradeSchema(CassandraStorage.java:257)
	at io.cassandrareaper.storage.CassandraStorage.<init>(CassandraStorage.java:245)
	at io.cassandrareaper.storage.InitializeStorage.initializeStorageBackend(InitializeStorage.java:65)
	at io.cassandrareaper.ReaperDbMigrationCommand.run(ReaperDbMigrationCommand.java:50)
	at io.cassandrareaper.ReaperDbMigrationCommand.run(ReaperDbMigrationCommand.java:28)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:98)
	at io.dropwizard.cli.Cli.run(Cli.java:78)
	at io.dropwizard.Application.run(Application.java:94)
	at io.cassandrareaper.ReaperApplication.main(ReaperApplication.java:102)
```

I will investigate on Monday.

adutra:
I take it back: it might be a quick win to just alter Reaper's keyspace. I will try that route first.

pkg/reaper/resource.go
```diff
@@ -64,6 +65,12 @@ func NewReaper(
 			desiredReaper.Spec.UiUserSecretRef.Name = DefaultUiSecretName(kc.Name)
 		}
 	}
+	// If the cluster is already initialized and some DCs are flagged as stopped, we cannot achieve QUORUM in the
+	// cluster for Reaper's keyspace. In this case we simply skip schema migration, otherwise Reaper wouldn't be able to
+	// start up.
+	if kc.Status.GetConditionStatus(k8ssandraapi.CassandraInitialized) == corev1.ConditionTrue && kc.HasStoppedDatacenters() {
```
adutra, line 71: I'm hesitant here. What conditions should we check to decide to skip schema migration?

test/e2e/stargate_apis_test.go
```diff
@@ -255,8 +255,6 @@ func createKeyspaceAndTableNative(t *testing.T, connection *client.CqlClientConn
 		tableName,
 	))
 	require.IsType(t, &message.SchemaChangeResult{}, response.Body.Message, "Expected CREATE TABLE response to be of type SchemaChangeResult")
-	response = sendQuery(t, connection, fmt.Sprintf("TRUNCATE %s.%s", keyspaceName, tableName))
-	assert.IsType(t, &message.VoidResult{}, response.Body.Message, "Expected TRUNCATE response to be of type VoidResult")
```
adutra, line 259: This TRUNCATE query fails randomly with TruncateError. I think we can safely remove it.

adutra:
This is again more or less ready for another review. I'm not entirely happy with the workaround for the Reaper issue, but the test is passing.

pkg/reaper/resource.go
jsanda: I deployed a cluster with Reaper `deploymentMode: SINGLE`. Reaper is deployed to dc1. I go back and forth stopping/starting dc2. It causes an update of the Reaper Deployment. That should be avoidable. I'm ok with fixing it in a follow up issue if you prefer since it is arguably an optimization.

jsanda (approved):
I left a few comments in my last review, but I think they can be addressed in subsequent issues.

PR #338, opened 2022-01-31
[DO NOT MERGE] Demo of GHA failures for multiDcMultiCluster test

Miles-Garnsey:
A demo of the way that multiDcMultiCluster can't upload artefacts when it fails. It tries to upload to a directory with a `/` in the path, due to naming it equal to the Go subtest name.


Miles-Garnsey:
I'm thinking we may be able to fix this by using GHA's [expressions](https://docs.github.com/en/actions/learn-github-actions/expressions#format) to manipulate the path and replace the slashes with something GHA will accept.

The change would be on [this](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/demo-multiDcMultiCluster-failure/.github/workflows/kind_e2e_tests.yaml#L140) line.

PR #341, opened 2022-02-01
Update kustomize and controller-gen, fix CRD_OPTIONS, fix kuttl tests

burmanm:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/kuttl_tests.yaml
+++ b/.github/workflows/kuttl_tests.yaml
@@ -67,6 +67,16 @@ jobs:
         ref: ${{ github.event.pull_request.head.sha }}
     - uses: actions/checkout@v2
       if: github.event_name != 'pull_request'
+    - name: Set up Go 1.16
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.16
+    - uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
     - name: Cache Docker layers
       uses: actions/cache@v2
       with:
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE):latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd"
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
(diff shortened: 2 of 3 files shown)


burmanm:
@Miles-Garnsey

Makefile
```diff
@@ -41,7 +41,7 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE):latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd"
```
jsanda, line 44: Does `preserveUnknownFields` default to `false`?

jsanda (commented):
Shouldn't this PR include the update CRD manifests?

Can you also create an issue to reference?

Makefile
burmanm: Yes and those options are no longer present in newer controller-gen versions. Leaving those options makes the build fail.

burmanm:
@Miles-Garnsey I'll move this discussion here for just the kustomize version, but this applies to PR #286 requirements. Requiring Kustomize 4.4.1 will break usage of `kubectl apply -k` for most users. The default version for some cloud providers is 1.21, but people might use even older ones.

1.21 kubectl ships with Kustomize 4.0.5 integrated. https://github.com/kubernetes-sigs/kustomize#kubectl-integration

I think we should keep the ability to install k8ssandra-operator with just `kubectl apply -k` instead of requiring user to download Kustomize separately before installation. And we can't expect every user to be running the newest kubectl client especially if they're using a server based of 1.21. kubectl will actually even complain about the version skew. 

So if those replacement changes require newer version of Kustomize, we might need to think about doing them in post-1.0 and instead now concentrate on some approach (ugly or not) that would work with older kustomize version also.

.github/workflows/kuttl_tests.yaml
```diff
@@ -75,6 +75,16 @@ jobs:
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+    - name: Set up Go 1.16
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.16
+    - uses: actions/cache@v2
```
adutra, line 82: Could you give a name to this step? Otherwise it shows up as "Run actions/cache@v2".

adutra (approved):

.github/workflows/kuttl_tests.yaml
burmanm: Done. Although I'm not sure what value the name has in this case ;)

.github/workflows/kuttl_tests.yaml
```diff
@@ -75,6 +75,17 @@ jobs:
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+    - name: Set up Go 1.16
```
Miles-Garnsey, line 78: Thanks for identifying and fixing this @burmanm, I can see what you meant about these tests using the wrong Go version now.

jsanda:
@burmanm I think we can ignore the test failures if that is holding up merging.

PR #347, opened 2022-02-03
Expose AllowMultipleNodesPerWorker [K8SSAND-1203]

### LINKED ISSUE #342: K8SSAND-1203 ⁃ Expose AllowMultipleWorkersPerNode from CassandraDatacenter spec
**What is missing?**
The `AllowMultipleWorkersPerNode` allows multiple C* pods to be co-located on the same k8s worker node.

**Why do we need it?**
There are times during dev/testing when it is convenient to be able to co-locate C* pods particularly if there are a limited number of worker nodes. Some folks may use this for production deployments as well.

**Environment**

* K8ssandra Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1203) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1203
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Expose AllowMultipleNodesPerWorker property from the CassandraDatacenter CRD.

Fixes #342

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/cassandra/datacenter.go
+++ b/pkg/cassandra/datacenter.go
@@ -2,6 +2,7 @@ package cassandra
 
 import (
 	"fmt"
+
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	"github.com/k8ssandra/cass-operator/pkg/reconciliation"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/cassandra/datacenter.go
```diff
@@ -103,6 +104,7 @@ type DatacenterConfig struct {
 	Users                 []cassdcapi.CassandraUser
 	PodTemplateSpec       *corev1.PodTemplateSpec
 	MgmtAPIHeap           *resource.Quantity
+	SoftPodAntiAffinity   *bool
```
jsanda, line 107: Do we need a pointer here? I think it would be simpler to just use the value type.

jsanda (commented):
Just one minor question/suggestion.

pkg/cassandra/datacenter.go
jsanda: Talked with @Miles-Garnsey offline and the pointer is indeed needed here.

jsanda (approved):
Thanks for knocking this out!

pkg/cassandra/datacenter.go
Miles-Garnsey: We need to use a pointer to distinguish false from unset here so that users can override the cluster-level setting with the DC level setting. 

If the cluster level is true but DC level is false, we need to ensure that the DC-level false value is really false, not just unset.

PR #366, opened 2022-02-07
check for schema agreement before making any schema changes [K8SSAND-1167]

### LINKED ISSUE #295: K8SSAND-1167 ⁃ Check for schema agreement before applying schema changes
**What is missing?**
This is a follow up to #21. With the changes being introduced in #262, there several places/scenarios where we can apply schema changes in Cassandra. Specifically we change the replication of several keyspaces. We should be checking for schema agreement in the C* cluster before we apply each change.

**Why do we need it?**
Applying schema changes when the version is not consistent across the cluster can introduce instability. Its a good practice in general to check for schema version agreement before applying changes.

This should be done as a follow up to #262.

**Environment**

* K8ssandra Operator version:

```
0538e957
```

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1167) by [Unito](https://www.unito.io)
┆Reviewer: 5f9072f1950f8c007638ef6d
┆friendlyId: K8SSAND-1167
┆priority: High

jsanda:
**What this PR does**:

Fixes #295

**Checklist**
- [ ] Changes manually tested


pkg/cassandra/management.go
```diff
@@ -312,3 +331,32 @@ func (r *defaultManagementApiFacade) GetSchemaVersions() (map[string][]string, e
...
+		//        ],
+		//        "aa573028-7c70-3b8f-a247-13bbca2011d2": [
+		//            "172.18.0.9",
+		//            "172.18.0.6"
+		//        ]
+		//    }
+		//
+		// We exclude these keys from the check.
+		if uid == "UNREACHABLE" {
+			delete(versions, uid)
```
emerkle826, line 357: This is more or a question than a request for change, but is it safe to remove an element from the `versions` map while iterating over the map? I'm not as familiar with how this works in Golang, but in some other languages you may create a concurrent modification error doing this.

burmanm: This is safe operation in golang.

jsanda:
Note that I am planning to add/update integration tests but got sidetracked investigated the `MultiDcAuthOnOff` e2e failure. This is one of the tests that we see fail periodically and often enough. It led to my creating https://github.com/thelastpickle/cassandra-reaper/issues/1164.

jsanda:
I will create a separate ticket for writing additional integration tests. I won't have time to do it now. We do have some coverage though.

Miles-Garnsey (commented):
This PR implements schema-agreement checking logic to avoid contention and messaging storms in Cassandra's schema propagation mechanisms (gossip, I believe) which can arise due to concurrent schema updates. It helps us avoid some well-known [issues](https://issues.apache.org/jira/browse/CASSANDRA-9424).

## General questions

### Should this be in cass-operator?

Because this PR only protects mutations issued by k8ssandra-operator from being issued in the absence of schema agreement, I wonder if it wouldn't be better to implement exactly this logic in `cass-operator`, so that `k8ssandra-operator` can continue to import `httphelper` and both operators can enjoy the same protection from concurrent schema updates.

### Do we still foresee a thundering herd problem?

While this PR is a net-good in that it protects Cassandra from concurrent schema updates, we should acknowledge that it moves some risk onto the operators themselves. 

We will likely observe a thundering herd problem as multiple retries are issued at bootstrap. We are already "mitigating" this by performing most operations in serial within k8ssandra-operator, but I'd be concerned that it might become problematic for large clusters.

We should also expect the problem to become more pronounced as we move to a more concurrent model of execution. Mitigations might include adding exponential backoff and jitter to requeued requests. This might be possible via the RateLimiter in controller-runtime (I have not fully researched what jitter/backoff options are available and this could be a wholly separate ticket).

## Suggestions

I think it would be good to see:

1. Some rationale as to why we want to implement this functionality here rather than in cass-operator, and potentially a plan to migrate this functionality into cass-operator over the longer term.
2. Tickets opened to implement some sort of rate limiting and jittering of the requests.
3. A plan/tickets to move to a middleware model, or discussion as to why we shouldn't do that _(see inline comments)_.
4. A decision on how we want to package errors _(see inline comments)_.
3. John has mentioned: `There is one type of test we are missing -  a negative test where we have schema disagreement` I think we should ensure that this is added. It may be as simple as adding some tests in the controllers (e.g. [Stargate](https://github.com/jsanda/k8ssandra-operator/blob/schema-agreement/controllers/k8ssandra/stargate.go#L172)) where we stub out the `cassandra.ManagementApiFacade` unit with one that always returns a `kerrors.IsSchemaDisagreement(err) = true` to make sure that they requeue. 

If merge is urgent, all of the above can be put in the backlog rather than considered blocking. 

Also happy to prove assistance on any of the above if that would be helpful.

jsanda:
I have added a couple tests where `EnsureKeyspaceReplication` returns an error. It's hard to do much more than that right now. 

Replying to suggestions from @Miles-Garnsey...

>A decision on how we want to package errors 

The errors packages are not part of public facing APIs, so I am not overly concerned about that right now. It's certainly a lower priority at the moment. 

>Tickets opened to implement some sort of rate limiting and jittering of the requests.

The controller already handles this for us. See [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L46) and [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L120).

>Some rationale as to why we want to implement this functionality here rather than in cass-operator, and potentially a plan to migrate this functionality into cass-operator over the longer term.

This functionality isn't used in cass-operator. There have been prior discussions about moving the management-api client out into its own project, separate from cass-operator. I am in favor of doing that. It will allow us to more easily test those APIs directly which I think is a good thing.

>A plan/tickets to move to a middleware model, or discussion as to why we shouldn't do that

I suppose it is worth a discussion post-1.0. As for the suggestion about `AgreementCheckedMgmtAPIFacade`, I don't think it is necessarily a good idea because we always want to check for schema agreement for making changes, just like the drivers do.

adejanovski (approved):
LGTM

I've checked all e2e test failures and they either pass locally or aren't related to these changes (timeout trigger right before Reaper gets to ready state for two of them).

PR #367, opened 2022-02-08
configure mocks with testing.T instance to avoid panics [K8SSAND-1197]

### LINKED ISSUE #332: K8SSAND-1197 ⁃ integration tests don't clean up when there is a panic
**What happened?**
We use the `envtest` package from controller-runtime for integration tests. Its a kube api server and etcd which run in their own processes. When there is a panic in the tests, cleanup does not happen, and we are left with running kube and etcd processes. If you just run `make test` this can easily result in a large number of processes. A separate test environment is created per controller, and multiple servers started per environment.

We use the `mocks` package from https://github.com/stretchr/testify, and I have found it to be the biggest culprit in this. When a mock receives an unexpected method calls it throws a panic. There is the `Test(t TestingT)` method for Mock objects. If you call it and pass the `testing.T` instance, the Mock will fail the test instead of panicking. 

There is another problem. We make extensive use of subtests. When subtest or one of subtests panics, remaining sibling subtests don't get executed. Here is a simple example to illustrate:

```go
func TestFailures(t *testing.T) {
	t.Run("Fail1", ftest(t, "one", fail))
	t.Run("Fail1", ftest(t, "two", fail))
}

type testFunc func(t *testing.T, msg string)

func ftest(t *testing.T, msg string, f testFunc) func(t *testing.T) {
	return func(t *testing.T) {
		f(t, msg)
	}
}

func fail(t *testing.T, msg string) {
	panic(msg)
}
```

Tests should clean up such that there is no lingering processes. If there is a panic it should just be treated as a failure. Further more remaining subtests should continue execution.

**Environment**

* K8ssandra Operator version:

```
1.0.0-alpha.3
```

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1197) by [Unito](https://www.unito.io)
┆fixVersions: k8ssandra-operator-v1.0.1
┆friendlyId: K8SSAND-1197
┆priority: Medium

jsanda:
**What this PR does**:
There is some indirection involved for setting the `testing.T` instance on the mocks because the factory objects for producing the mocks are configure as member variables of the Reconciler.

Fixes #332

The scope of this PR has been reduced to handling panics with mocks. I created #376 to address handling panics in subtests.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.0.md
+++ b/CHANGELOG/CHANGELOG-1.0.md
@@ -15,10 +15,9 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
-## v0.5.0 2022-02-10
-
-* [CHANGE] [#120](https://github.com/k8ssandra/k8ssandra-operator/issues/120) Remove `systemLoggerResources` from K8ssandraCluster spec
-* [FEATURE] [#296](https://github.com/k8ssandra/k8ssandra-operator/issues/296) Expose the stopped property from the CassandraDatacenter spec
+* [FEATURE] [#296](https://github.com/k8ssandra/k8ssandra-operator/issues/296) Expose the stopped property from the
+  CassandraDatacenter spec
+* [TESTING] [#332](https://github.com/k8ssandra/k8ssandra-operator/issues/332) Configure mocks to avoid panics
 
 ## v0.4.0 2022-02-04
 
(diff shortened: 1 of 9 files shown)


jsanda:
Aside from a couple small formatting changes, this PR only makes changes to integration test code, so I'm not worried about the e2e test failures.

jsanda:
This PR only makes changes to test code, specifically integration test code, so the kuttl failures are unrelated.

adejanovski (approved):

PR #380, opened 2022-02-10
Add ConfigBuilderContainerImage and SystemLoggerContainerImage APIs [K8SSAND-1233] 

### LINKED ISSUE #378: K8SSAND-1233 ⁃ Provide support for custom system-logger and config-builder image coordinates
**What is missing?**
It should be possible to specify the complete image coordinates for all of the images to be used by the elements deployed by K8ssandra Operator so that users can provide images via their own private repositories as required for security or other reasons.

Currently there's a known lack of configurability for the system-logger image and the config-builder image.

**Why do we need it?**
To remove barriers in certain potential use cases such as security rules and regulations.  To provide feature parity with k8ssndra 1.x.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1233) by [Unito](https://www.unito.io)
┆fixVersions: k8ssandra-operator-v1.1.0
┆friendlyId: K8SSAND-1233
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Adds two new fields to the Cassandra portions of the APIs to control which images are used for each container.

Fixes #378 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -536,6 +536,32 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace 
 	require.Equal(t, k8ssandra.Spec.Cassandra.SystemLoggerContainerImage, expectedSystemLoggerContainerImage)
 	require.Equal(t, k8ssandra.Spec.Cassandra.ConfigBuilderContainerImage, expectedConfigBuilderContainerImage)
 
+	// Confirm non-default images are being applied as expected and merged with defaults.
+	partialSystemLoggerContainerImage := &images.Image{
+		Tag: "d94231ba",
+	}
+	partialConfigBuilderContainerImage := &images.Image{
+		Tag: "1.0.4",
+	}
+
+	patch := client.StrategicMergeFrom(k8ssandra.DeepCopy())
+	k8ssandra.Spec.Cassandra.SystemLoggerContainerImage = partialSystemLoggerContainerImage
+	k8ssandra.Spec.Cassandra.ConfigBuilderContainerImage = partialConfigBuilderContainerImage
+	if err := f.Client.Patch(ctx, k8ssandra, patch); err != nil {
+		require.FailNow("could not patch K8ssandraCluster with new images")
+	}
+	expectedSystemLoggerContainerImage.Tag = partialSystemLoggerContainerImage.Tag
+	expectedConfigBuilderContainerImage.Tag = partialConfigBuilderContainerImage.Tag
+	if err := f.Client.Get(ctx, kcKey, k8ssandra); err != nil {
+		require.FailNow("could not Get K8ssandraCluster to check images")
+	}
+	require.Equal(t, k8ssandra.Spec.Cassandra.SystemLoggerContainerImage, expectedSystemLoggerContainerImage)
+	require.Equal(t, k8ssandra.Spec.Cassandra.ConfigBuilderContainerImage, expectedConfigBuilderContainerImage)
+
+	// Patch back to defaults to proceed with tests
+	// TODO
+	// End image tests
+
 	stargateKey := framework.ClusterKey{K8sContext: "kind-k8ssandra-0", NamespacedName: types.NamespacedName{Namespace: namespace, Name: "test-dc1-stargate"}}
 	checkStargateReady(t, f, ctx, stargateKey)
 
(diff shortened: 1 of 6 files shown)


Miles-Garnsey:
This is a bit of a frustrating one to test, since I think we'll need to hook into the e2e tests. From what I can see, we can't unit test the behaviour of the `default` struct tag because it interacts with the API server to produce the functionality we want.

This will slow down the tests by a fair bit because it will require yet another rolling restart 😔

test/e2e/suite_test.go
```diff
@@ -518,6 +520,45 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace
...
+		Repository: "datastax",
+		Name:       "cass-config-builder",
+		Tag:        "1.0.4-ubi7",
+	}
+	require.Equal(t, k8ssandra.Spec.Cassandra.SystemLoggerContainerImage, expectedSystemLoggerContainerImage)
+	require.Equal(t, k8ssandra.Spec.Cassandra.ConfigBuilderContainerImage, expectedConfigBuilderContainerImage)
+
+	// Confirm non-default images are being applied as expected and merged with defaults.
+	partialSystemLoggerContainerImage := &images.Image{
+		Tag: "latest",
```
jsanda, line 541: Don't we want something other than `latest` here since that is the default?

jsanda (changes requested):
Changelog update needed

test/e2e/suite_test.go
```diff
@@ -518,6 +520,45 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace
 		return cassandraDatacenterReady(kdcStatus.Cassandra)
 	}, polling.k8ssandraClusterStatus.timeout, polling.k8ssandraClusterStatus.interval, "timed out waiting for K8ssandraCluster status to get updated")
 
+	// Confirm default images are being applied as expected.
+	expectedSystemLoggerContainerImage := &images.Image{
```
jsanda, line 524: It probably would be sufficient to test this via unit tests. Granted some form of coverage is better than no coverage.

jsanda: I looked some more at the test update, and it does feel out of place in the e2e test. What about adding/updating tests in `datacenter_test.go`?

test/e2e/suite_test.go
Miles-Garnsey: Yes, you're entirely right. I was half way through updating when my laptop crashed. Docker on Mac! 😩

test/e2e/suite_test.go
Miles-Garnsey: Are you saying to use an envtest instead? That's possible, but I don't think we can do it via a unit test because the `default` struct tags are acted on somewhere in the API server as far as I can tell.

Miles-Garnsey: I initially wanted it in `k8ssandracluster_types_test.go` but there's no API server there.

jsanda: I am in fact suggesting unit tests, even if less than ideal :) In the tests we would have to copy from the annotations what the expected defaults are. That's a tradeoff I can live with. 

If you look at the what was done for the Reaper image in `pkg/reaper/deployment.go` the image defaults are defined as constants.

Miles-Garnsey: It **sounds** like that would make my life a lot easier, so I'm very supportive of a plan in this direction :)  But I'm not 100% clear on what you're proposing? 

Can you fill out what you're thinking the unit under test is, the desired inputs and outputs? Would this test just amount to another test of the DC/Cluster level merge logic in `datacenter.go`?

Miles-Garnsey: One of the main reasons I'm doing a test against the API server here is just that I'm actually not sure how the `default` struct tag is processed. I'm assuming that the API server does this work, but I'm not certain. 

Equally, I'm making a big assumption that using this tag will actually result in merge-like behaviour (where the user can just set a single field in `Image` and all other fields will fall back to their defaults). But what if I'm wrong and setting `SystemLoggerContainerImage` to any `Image` doesn't result in a merge, but just an overwrite of the whole `Image` instance?

Miles-Garnsey: > I'm making a big assumption that using this tag will actually result in merge-like behaviour (where the user can just set a single field in Image and all other fields will fall back to their defaults). But what if I'm wrong and setting SystemLoggerContainerImage to a partially definedImage doesn't result in a merge, but just an overwrite of the whole Image instance?

So I think I was right. Using the `default` tag doesn't appear to solve the issue as we were hoping. This is the part of the test which checks that patching with only a partial `Image` instance will result in the defaults being applied to the other fields. This is basically how it ends:
```
                                expected  : &images.Image{Registry:"docker.io", Repository:"k8ssandra", Name:"system-logger", Tag:"d94231ba", PullPolicy:"", PullSecretRef:(*v1.LocalObjectR
                                actual: &images.Image{Registry:"docker.io", Repository:"", Name:"", Tag:"d94231ba", PullPolicy:"", PullSecretRef:(*v1.LocalObjectReference)(nil)}
```

As we can see, it appears the image only has one field defined. Using the `default` struct tag will not avoid the issue I was concerned about where changing any part of the image (e.g. tag) meant you had to define every field in the image.

adejanovski:
Closing as stale and most probably out of sync.

PR #384, opened 2022-02-10
add AdditionalSeeds property [K8SSAND-1235]

### LINKED ISSUE #381: K8SSAND-1235 ⁃ Datacenter migrations with K8ssandra Operator clusters isn't possible because additional seed access isn't provided
┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1235) by [Unito](https://www.unito.io)
┆affectedVersions: k8ssandra-operator-v1.0.0
┆friendlyId: K8SSAND-1235
┆priority: Highest

jsanda:
**What this PR does**:
Adds an AdditionalSeeds property that can be used when migrating from an existing cluster not created or managed by k8ssandra-operator.

TODO:
* Update integration tests

Fixes #381

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/seeds.go
+++ b/controllers/k8ssandra/seeds.go
@@ -71,23 +71,34 @@ func (r *K8ssandraClusterReconciler) reconcileSeedsEndpoints(
 		// If we already have an Endpoints object but no seeds then it probably means all
 		// Cassandra pods are down or not ready for some reason. In this case we will
 		// delete the Endpoints and let it get recreated once we have seed nodes.
-		if len(seeds) == 0 {
-			logger.Info("Deleting endpoints")
-			if err := remoteClient.Delete(ctx, actualEndpoints); err != nil {
-				logger.Error(err, "Failed to delete endpoints")
+		//if len(seeds) == 0 {
+		//	logger.Info("Deleting endpoints")
+		//	if err := remoteClient.Delete(ctx, actualEndpoints); err != nil {
+		//		logger.Error(err, "Failed to delete endpoints")
+		//		return result.Error(err)
+		//	}
+		//} else {
+		//	if !annotations.CompareHashAnnotations(actualEndpoints, desiredEndpoints) {
+		//		logger.Info("Updating endpoints", "Endpoints", endpointsKey)
+		//		actualEndpoints := actualEndpoints.DeepCopy()
+		//		resourceVersion := actualEndpoints.GetResourceVersion()
+		//		desiredEndpoints.DeepCopyInto(actualEndpoints)
+		//		actualEndpoints.SetResourceVersion(resourceVersion)
+		//		if err = remoteClient.Update(ctx, actualEndpoints); err != nil {
+		//			logger.Error(err, "Failed to update endpoints", "Endpoints", endpointsKey)
+		//			return result.Error(err)
+		//		}
+		//	}
+		//}
+		if !annotations.CompareHashAnnotations(actualEndpoints, desiredEndpoints) {
+			logger.Info("Updating endpoints", "Endpoints", endpointsKey)
+			actualEndpoints := actualEndpoints.DeepCopy()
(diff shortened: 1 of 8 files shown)


controllers/k8ssandra/seeds.go
```diff
@@ -127,6 +133,10 @@ func newEndpoints(dc *cassdcapi.CassandraDatacenter, seeds []corev1.Pod) *corev1
 		})
 	}
 
+	for _, seed := range additionalSeeds {
+		addresses = append(addresses, corev1.EndpointAddress{IP: seed})
```
burmanm, line 137: Do we want to have these userdefined seeds also as endpoints or added to the additionalSeeds in the CassandraDatacenter?

jsanda: What would be the reason or advantage for doing so?

burmanm: It would separate the user input with automated processing. It would quickly show if some weird seed is because of user's values (when debugging) or because of something k8ssandra-operator pushed. Semantics only.

adejanovski:
I was able to successfully use the additionalSeeds property from this PR with the `seeds.go` additions mentioned above with 2 GKE clusters 👍

controllers/k8ssandra/seeds.go
jsanda: The user's values are already stored in the K8ssandraCluster spec, so I'm not sure I see the benefit of putting them in the CassandraDatacenter spec for debugging. More importantly though, if we set CasssndraDatacenter.Spec.AdditionalSeeds, then we will have both cass-operator and k8ssandra-operator reconciling the Endpoints. I fear that might cause problems.

adejanovski: Passing the additional seeds to the cassdc would indeed duplicate the effort (although it makes debugging easier indeed...).
It should be either/or IMO, and the way John implemented it is relevant with how k8ssandra-operator approaches seed management.

jsanda:
@adejanovski let's sync up and walk through testing together. I went back and retested two scenarios locally using two kind clusters. First I tested connecting a new K8ssandraCluster to an existing one. The pods came up without error. `nodetool status` shows that both DCs can see each other. 

Next I tested connecting a new K8ssandraCluster to a stand-alone CassandraDatacenter, i.e., one that is not part of a K8ssandraCluster. The results were the same as my first test.

>You'd need to modify line 74 with `if len(seeds) == 0 && len(additionalSeeds) == 0 {` and line 103 with: `if len(seeds) > 0 || len(additionalSeeds) > 0 {`

I am not convinced that line 74 or 103 needs to or should change. cass-operator will labels the first pod as a seed before its starts Cassandra. Said differently cass-operator won't start a C* node without having a seed in the local dc.

jsanda:
I just tested my second scenario in GKE, and it worked as expected.

controllers/k8ssandra/seeds.go
```diff
@@ -50,39 +50,56 @@ func (r *K8ssandraClusterReconciler) findSeeds(ctx context.Context, kc *api.K8ss
...
+		//		actualEndpoints := actualEndpoints.DeepCopy()
+		//		resourceVersion := actualEndpoints.GetResourceVersion()
+		//		desiredEndpoints.DeepCopyInto(actualEndpoints)
+		//		actualEndpoints.SetResourceVersion(resourceVersion)
+		//		if err = remoteClient.Update(ctx, actualEndpoints); err != nil {
+		//			logger.Error(err, "Failed to update endpoints", "Endpoints", endpointsKey)
+		//			return result.Error(err)
+		//		}
+		//	}
+		//}
```
adejanovski, line 92: Could you remove lines 71 to 92 before merging the PR?

adejanovski (approved):
Thanks for this PR and for making those last changes!

After removing the newly commented block in `seeds.go`, feel free to merge this PR.

PR #406, opened 2022-02-15
Use a LocalObjectReference for Medusa bucket secret

### LINKED ISSUE #401: K8SSAND-1246 ⁃ Medusa uses a secret ref which is inconsistent with the other secrets
The Medusa api exposes a secret ref for the storage bucket credentials, `StorageSecretRef`, as a `string` instead of a `LocalObjectReference`. This is inconsistent with all the other secret references of all CRDs in the operator.

We should probably standardize this across the APIs and switch it to a `LocalObjectReference` type.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-1246) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1246
┆priority: Medium

adejanovski:
**What this PR does**:
Changes the type of Medusa's Storage Secret Ref from a string to a LocalObjectReference.

Fixes #401 

**Checklist**
- [x] Changes manually tested


burmanm:
StorageSecretRef is not set as required, thus it is optional. It's not ever checked actually if it's there, everything is just assumed. Is this always a correct approach? What happens if I don't have "local", but do not have that field set? Is the default value alright to not break anything?

adejanovski:
Reconciliation should then fail because the secret won't be mounted correctly.
I can add a check prior to that though and fail the reconciliation with a more specific message.

burmanm (approved):

PR #409, opened 2022-02-15
Allow to specify external DCs for proper replication changes

### LINKED ISSUE #400: Keyspaces get altered incorrectly during DC migration
**What happened?**
When migrating from a cluster that isn't part of the k8ssandra-operator controlled cluster (Cassandra, cass-operator or k8ssandra 1.x), keyspaces such as `data_endpoint_auth` and `reaper_db` get altered in a way that removes replicas from the pre-existing DC.

This is due to the fact that the list of DC used to build the alter commands is picked from the K8ssandraCluster object, which isn't aware of the pre-existing DC.

Keyspaces should be altered in a way that retains the replicas on the old DC or not altered at all because the way it's currently done breaks Stargate and Reaper in the old DC.

Add a DC using k8ssandra-operator to an existing cluster, using the `additionalSeeds` property to perform the DC expansion.

**Environment**
GKE

* K8ssandra Operator version:
Custom build with the changes from [this PR](https://github.com/k8ssandra/k8ssandra-operator/pull/384).

1.21

* Kubernetes cluster kind:
GKE

adejanovski:
**What this PR does**:
Allows to specify external DCs (unmanaged by k8ssandra-operator) which will be added to the replication strategy of managed keyspaces (system_auth, data_endpoint_auth and reaper_db).
This way we can safely perform migration or run clusters in hybrid (k8ssandra 1.x/k8ssandra 2.x) mode.

Fixes #400 

**Checklist**
- [x] Changes manually tested


pkg/cassandra/util.go
```diff
@@ -76,12 +80,16 @@ func ComputeReplication(maxReplicationPerDc int, datacenters ...*cassdcapi.Cassa
 }
 
 // ComputeReplicationFromDcTemplates is similar to ComputeReplication but takes dc templates as parameters.
-func ComputeReplicationFromDcTemplates(maxReplicationPerDc int, datacenters ...api.CassandraDatacenterTemplate) map[string]int {
+func ComputeReplicationFromDcTemplates(maxReplicationPerDc int, externalDatacenters []string, datacenters ...api.CassandraDatacenterTemplate) map[string]int {
```
jsanda, line 83: Should we consider renaming the function now that it also take the `externalDatacenters` arg?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -61,6 +61,12 @@ type K8ssandraClusterSpec struct {
 	// If this is non-nil, Medusa will be deployed in every Cassandra pod in this K8ssandraCluster.
 	// +optional
 	Medusa *medusaapi.MedusaClusterTemplate `json:"medusa,omitempty"`
+
+	// During a migration or in hybrid mode the operator should alter keyspaces replication settings including these external DCs.
```
jdonenine, line 65: We chatted about this a little this afternoon... let's tweak the comment just a bit to remove the bits about hybrid mode.  I think the hybrid piece is really interesting, for a number of reasons, but for now until we think through what it looks like to run a hybrid cluster long term let's only state that we're doing this for temporary migration purposes.  Just trying to pull in our potential scope in terms of documentation/questions/support for now.

config/crd/bases/config.k8ssandra.io_clientconfigs.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.6.1
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/cassandra/util.go
adejanovski: I renamed it to `ComputeReplicationFromDatacenters`. How does that sound?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adejanovski: kk, I made the changes 👍

jdonenine (approved):

pkg/cassandra/util.go
jsanda: 👌

PR #387, opened 2022-02-11
Remove cass-operator namespace [K8SSAND-1239]

Miles-Garnsey:
**What this PR does**:

All components now install into the namespace k8ssandra-operator. But the cass-operator namespace is still created. 

This PR adds a patch to delete it.

**Which issue(s) this PR fixes**:
Fixes #385 

**Checklist**
- [ ] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [ ] CHANGELOG.md updated (not required for documentation PRs)
- [ ] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #413, opened 2022-02-16
Fix namespace removal

Miles-Garnsey:
**What this PR does**:

When I merged #387 it looked like all tests were passing but the multi-cluster tests just hadn't run. They all fail!

This is because the data plane kustomizations were wrong -  they applied the same deletion patch that was already applied in the control-plane kustomizations from which data-plane is derived. Because the namespace is already deleted, this causes an error in `kustomize build`.

Haven't raised one, this is a hotfix.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/config/components/single-namespace/kustomization.yaml
+++ b/config/components/single-namespace/kustomization.yaml
@@ -22,11 +22,11 @@ replacements:
         kind: ValidatingWebhookConfiguration
       fieldPaths:
       - webhooks.0.clientConfig.service.namespace
-# patchesStrategicMerge:
-# - |-
-#   apiVersion: v1
-#   kind: Namespace
-#   metadata:
-#     name: cass-operator
-#   $patch: delete
+patchesStrategicMerge:
+- |-
+  apiVersion: v1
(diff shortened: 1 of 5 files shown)


Miles-Garnsey:
I think this is good to merge despite some test failures:

* ClusterScoped/MultiDcMultiCluster - has been failing consistently for some time.
* CreateSingleDatacenterCluster - failed due to`INSTALLATION FAILED: Service "traefik" is invalid: spec.ports[3].nodePort: Invalid value: 30080: provided port is already allocated` This is likely something to do with a previous test not cleaning up after itself properly.
* [CreateSingleReaper] this fails due to Reaper not being cleaned up properly. It was failing as of commit `595c2c3bca6660c708f72f1b940c686b9ec76e03` which was prior to my [merge](https://github.com/k8ssandra/k8ssandra-operator/commit/8bf3fe76669db6eb30c75d54c985c87446521db5) that broke the multi-cluster e2e tests this morning.

jsanda (approved):

PR #433, opened 2022-02-24
Add HA control-plane design doc

Miles-Garnsey:
**What this PR does**:

What it says on the tin.

No issue. Migrating design docs from GDocs into MD files on-repo.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/design-roadmap/ha-control-plane.md
+++ b/docs/design-roadmap/ha-control-plane.md
@@ -1,0 +1,84 @@
+# K8ssandra-operator HA design
+
+## Background
+
+K8ssandra-operator is a Kubernetes (k8s) operator intended to simplify the deployment and management of Apache Cassandra Datacenters across multiple Kubernetes clusters. This is done to -
+1. Enable hybrid/multicloud use cases (which itself might be necessary for commercial, legal or regulatory reasons).
+2. Improve resilience.
+3. Localise data near the end user for performance reasons.
+
+It is important to note that - for the purposes of this document - we will assume that a standard k8ssandra-operator install involves several Kubernetes clusters, with Cassandra Datacenters in each. Where not otherwise specified, the word 'cluster' should be taken to mean Kubernetes cluster (not Cassandra cluster).
+
+## Problem statement
+
+One of the key reasons users might adopt k8ssandra-operator (and Kubernetes, Cassandra, and distributed systems generally) is resilience in the face of node, DC or network failures.
+
+However, at present, the failure of a single Kubernetes cluster (which would normally map to a Cassandra DC) hosting the control plane would prevent changes to the whole cluster until that specific Kubernetes cluster comes back online. 
+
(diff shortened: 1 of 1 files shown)


docs/design-roadmap/ha-control-plane.md
```diff
@@ -0,0 +1,84 @@
...
+
+Option 2 sounds attractive but may not be possible under real-world latency constraints - i.e. the HTTP connection between the client (e.g. kubectl) and the API server may timeout before the data can replicate. We would need to examine mitigations for this sort of issue.
+
+## Considerations and outstanding questions
+
+### Additional user stories that may be relevant
+
+This topic touches on a number of operational questions that users are likely to face in the real world. I'm not sure how many of these we have/want to consider, but I thought it'd be good to document them somewhere.
+
+- A user has an existing CassandraDataCenter from Cass operator. How can they join it to a new K8ssandra cluster?
```
adejanovski, line 71: The procedure is documented in this blog post: https://k8ssandra.io/blog/tutorials/how-to/how-to-migrate-an-existing-cluster-to-k8ssandra-operator-without-any-downtime/

docs/design-roadmap/ha-control-plane.md
```diff
@@ -0,0 +1,84 @@
...
+- What is the process for migrating control plane DCs and data plane DCs to new k8s clusters generally?
+- What cases do we support around running some DCs under k8ssandra and others externally? (e.g. in a traditional bare metal DC.) 
+- Do we support mixed (k8s/bare-metal) clusters only at migration or permanently.
```
adejanovski, line 74: My opinion there is that we can support them permanently, but the automation we have in k8ssandra-operator won't spill over to the external clusters. Operations spanning over both clusters will have to be handled manually, which can come with some surprises.

adejanovski: So I guess we'd recommend against it.

adejanovski (commented):
Very well written doc @Miles-Garnsey.

I think automated leader election in a multicluster environment is going to be tough to manage with simplicity (if we have 2 clusters and lose one, how can we ensure consistency in the leader election process?).
But here's my question: why would we want this to be automated? Can't we provide a manual procedure to switch control planes?
We will also need to copy over client configs to the data plane so that it can connect to the API server of remote k8s clusters. Not sure if that should be replicated from the start or done manually when needed (it could be unsafe to shuffle those around).

docs/design-roadmap/ha-control-plane.md
Miles-Garnsey: That's good from a migration standpoint. There's a question still as to whether we support it long term.

Miles-Garnsey: NB: we should also note that this ends up in a situation where the clusters are joined to the same Cassandra ring, but the cass-operator cluster isn't under the management of k8ssandra-operator. So there's two meanings to 'how can they join it to a new k8ssandra cluster' in that sentence.

docs/design-roadmap/ha-control-plane.md
Miles-Garnsey: I'm happy to recommend against. I suggest we leave the question in the doc though.

PR #440, opened 2022-02-28
Test multiple Cassandra versions

### LINKED ISSUE #360: Add matrixed C* versions support to tests
**What is missing?**
Currently we're directly setting the version of C* to be run in all test scenarios, much like what is available in k8ssandra/k8ssandra a matrixed version of testing that can support various versions of platform to be tested with the same harness would be very useful.

This work was started in https://github.com/k8ssandra/k8ssandra-operator/pull/137 but that work has become stale over time and not presents a risk to merge as is late in the 1.0.0 development cycle.  We should bring the purpose of that PR back to life post 1.0.0.

**Why do we need it?**
To provide more comprehensive automated tests.

Miles-Garnsey:
**What this PR does**:

Tests many Cassandra versions.

Fixes #360 

**Checklist**
- [ ] Changes manually tested


jsanda:
The intent for #360 was to have all of the e2e tests run against different C* versions. Don't get me wrong. I welcome this PR :) When we merge this, I want to make sure the e2e tests don't fall off the radar so I'd be inclined to keep #360 open.

Miles-Garnsey:
This is finally good to go. A lot of the issues I was observing appear to have been to do with configuring the cluster via kind instead of via kuttl. This may indicate differences in the behaviour of k8s versions that we'll need to watch out for, but they only cause timing issues, not real bugs.

Miles-Garnsey:
This PR closes tickets #404 #360 and #405.

.github/workflows/version_tests.yaml
```diff
@@ -0,0 +1,121 @@
...
+        kuttl-test:
+        - test-cassandra-versions
+        serverVersion: 
+        - 3.11.10
+        - 3.11.11
+        - 3.11.12
+        - 4.0.0
+        - 4.0.1
+        - 4.0.2
+        - 4.0.3
```
adejanovski, line 71: Changing the order allows to see which Cassandra version is tested on the summary screen of GHA.
Also, we didn't publish a 4.0.2 image and went straight to 4.0.3.
My suggestion would also be to test against the latest of each minor release. Testing all patch releases will not be sustainable in the long run.

```suggestion
      matrix:
        k8s_version: ["1.21"]
        serverVersion: 
        - 3.11.12
        - 4.0.3
        kuttl-test:
        - test-cassandra-versions
```

.github/workflows/version_tests.yaml
```diff
@@ -0,0 +1,121 @@
...
+        path: /tmp
+    - name: Load Docker images
+      run: |
+        docker load --input /tmp/k8ssandra-k8ssandra-operator.tar
+    - name: install kustomize
+      run: |
+        make kustomize
+    - name: Run kuttl-test
+      run: |
+        # We are running tests against k8s 1.20 - 1.22 currently.
```
adejanovski, line 112: Aren't we testing just 1.21?

adejanovski (changes requested):
Cool stuff Miles 👍 
I think we should restrict the matrix to testing the latest patch releases for each release branch.
Also swapping the order of variables in the matrix will make GHA's summary screen easier to read.

adejanovski:
I second John's comment on the fact that this PR doesn't fix #360 which is about e2e tests (replicating what we have in the k8ssandra project).

Miles-Garnsey:
> I second John's comment on the fact that this PR doesn't fix #360 which is about e2e tests (replicating what we have in the k8ssandra project).

That was the original scope of the ticket, although after discussion Jeff and I [agreed](https://github.com/k8ssandra/k8ssandra-operator/issues/360#issuecomment-1050918044) that starting with the kuttl tests made the most sense. 

Also, (at that time) the e2e tests were very flaky. It seemed that adding to complexity there was only going to exacerbate problems and be a time sink. That situation may have changed over the past 3 weeks.

.github/workflows/version_tests.yaml
Miles-Garnsey: Yes we are. That comment was left over from a copy + paste from another workflow. Let me fix that...

.github/workflows/version_tests.yaml
Miles-Garnsey: All good points/suggestions. I've implemented all of these.

adejanovski (approved):

PR #446, opened 2022-03-02
Run e2e tests against arbitrary context names

### LINKED ISSUE #112: K8SSAND-862 ⁃ Refactor e2e tests so they can be run against clusters other than ones created by setup script
We should be able to run the e2e tests against any clusters provided that they have sufficient resources and pods have routable IPs. There are at least a couple things that need to be fixed to support this.

The k8s context names are hard coded in a bunch of places. Consider the following example:

```go
t.Log("check that datacenter dc1 is ready")
dcKey := framework.ClusterKey{K8sContext: "kind-k8ssandra-0", NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
withDatacenter := f.NewWithDatacenter(ctx, dcKey)
```

`kind-k8ssandra-0` corresponds to what is generated by `setup-kind-multicluster.sh`. The contexts need to be configurable.

The second issue involves the kubeconfig files used by the tests. When you create a kind cluster the kubeconfig entry that is creates uses localhost for the api server address. We generate a second kubeconfig that uses the docker network address. There are some checks that make sure each of these kubeconfig files exists. If I want to run the tests against GKE or EKS clusters for instance, these extra steps shouldn't be necessary.

When these issues are resolved we can start working on setting some GHA workflows for running the tests with the various cloud providers.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-862) by [Unito](https://www.unito.io)
┆fixVersions: k8ssandra-operator-v1.0.1
┆friendlyId: K8SSAND-862
┆priority: Medium

adutra:
**What this PR does**:
Run e2e tests against arbitrary context names

Fixes #112 

**Checklist**
- [ ] Changes manually tested


test/framework/e2e_framework.go
```diff
@@ -446,17 +452,25 @@ func (f *E2eFramework) DeployK8sClientConfigs(namespace string) error {
 		return err
 	}
 
+	destCfg := fmt.Sprintf(*configFileTemplate, 0)
+	kubeconfigs := filepath.Join(baseDir, *configDir)
+	inClusterKubeconfigs := filepath.Join(baseDir, *inClusterConfigDir)
```
jsanda, line 457: I need to look more closely at this, but I want to point out that the in-cluster kubeconfig is only needed for a local cluster like kind.

jsanda (commented):
One thing that isn't addressed is the manifests in the fixtures directory. We have stuff like this:

```yaml
    datacenters:
      - metadata:
          name: dc1
        k8sContext: kind-k8ssandra-0
        size: 3
      - metadata:
          name: dc2
        k8sContext: kind-k8ssandra-1
        size: 3
```
Those context names are generated by kind. I am inclined to say we use more generic context names like `control-plane`, `data-plane-1`,  `data-plane-2`, etc. If we do that we need to update manifests in the fixture dir. We will also need to update the `setup-kind-multicluster.sh` script to modify the context names. I think this will involve some yaml parsing to change the generated context names (unless there is a way to tell kind what they should be).

Lastly, the ultimate litmus test will be running against non-local clusters 😀

adutra:
> Those context names are generated by kind. I am inclined to say we use more generic context names like `control-plane`, `data-plane-1`, `data-plane-2`, etc. If we do that we need to update manifests in the fixture dir. We will also need to update the `setup-kind-multicluster.sh` script to modify the context names. I think this will involve some yaml parsing to change the generated context names (unless there is a way to tell kind what they should be).

So after hesitating a lot I went with the idea of templatizing the fixture files:

```
kind: K8ssandraCluster
metadata:
  name: test
spec:
  cassandra:
    serverVersion: 3.11.11
    datacenters:
      - metadata:
          name: dc1
        k8sContext: {{ .DataPlane1 }}
      - metadata:
          name: dc2
        k8sContext: {{ .DataPlane2 }}
```

The template variables get replaced with the contents of the `- dataPlane1` and `-dataPlane2` flags.

It seems to be working well, however this makes the fixture files invalid YAML which is not great. I am open to better solutions.

.github/workflows/kind_multicluster_e2e_tests.yaml
```diff
@@ -122,7 +130,17 @@ jobs:
       - name: Setup kind clusters
         run: make IMG="k8ssandra/k8ssandra-operator:latest" create-kind-multicluster kind-load-image-multi
       - name: Run e2e test ( ${{ matrix.e2e_test }} )
-        run: make E2E_TEST=TestOperator/${{ matrix.e2e_test }} NODETOOL_STATUS_TIMEOUT=2m e2e-test
```
adutra, line 125: The default timeout was already 2 minutes, so I removed this variable.

test/framework/e2e_framework.go
```diff
@@ -446,17 +473,25 @@ func (f *E2eFramework) DeployK8sClientConfigs(namespace string) error {
...
 		f.logger.Info("Creating ClientConfig", "cluster", k8sContext)
-		srcCfg := fmt.Sprintf("k8ssandra-%d.yaml", i)
-		cmd := exec.Command(
-			filepath.Join("scripts", "create-clientconfig.sh"),
-			"--src-kubeconfig", filepath.Join("build", "kubeconfigs", srcCfg),
-			"--dest-kubeconfig", filepath.Join("build", "kubeconfigs", "k8ssandra-0.yaml"),
-			"--in-cluster-kubeconfig", filepath.Join("build", "kubeconfigs", "updated", srcCfg),
-			"--output-dir", filepath.Join("build", "clientconfigs", srcCfg),
-			"--namespace", namespace)
+		srcCfg := fmt.Sprintf(*configFileTemplate, i)
```
jsanda, line 482: This won't work if I want to use `~/.kube/config` to run tests against some GKE clusters for example. I recently created some GKE clusters and configured the context names to be control-plane, east, west, and central. Here is an example cmd line I used to generate a ClientConfig for the central cluster:

```
./scripts/create-clientconfig.sh --namespace k8ssandra-operator \
                                 --src-context central  \
                                 --dest-context control-plane \
                                 --output-dir clientconfigs
```
`DeployK8sClientConfigs` most other setup code assumes a directory structure and configs as created by the `setup-kind-multicluster.sh` script.

I authenticate to my GKE clusters with the `gcloud container clusters get-credentials` command. It adds a context (and other stuff) in `~/.kube/config`. Tools for other cloud providers will also update the default kubeconfig. It makes sense then to also be able to use it for e2e tests. Since there may be any number of other clusters/contexts in the default config, we should require the control plane and data plane contexts to be explicitly set.

jsanda:
>The template variables get replaced with the contents of the `-dataPlane1` and `-dataPlane2` flags.
>
>It seems to be working well, however this makes the fixture files invalid YAML which is not great. I am open to better solutions.

I'm not a fan of the templates. What happens not if but when we want to make other parts of the fixtures configurable? Memory and CPU resources are good examples. Host networking is another really good one. We don't need to use host networking when we are running in non-local clusters. Do we template out these additional properties? Wouldn't it be better to just use Kustomize or Helm?

The test fixture manifests were intentionally write as simple YAML manifests so as to provide a source of examples that we can references for folks. If we start adding stuff to them, the examples become less discernible. While I find that kustomize ca be difficult to work with in some situations, I would lean towards it because allows us to keep the test fixture manifests intact without any changes. We instead layer changes on top.

I'm still in favor of making the default context names for the control plane and data planes more generic instead of using the ones that `kind` generates. I think that would make it easier to punt on making the `k8sContext` property configurable. I say that because if I want to run tests against some GKE clusters, it is easy enough (especially with `kubectx`) for me to rename contexts to `control-plane`, `data-plane-1`, etc.

test/framework/e2e_framework.go
```diff
@@ -47,61 +78,52 @@ type E2eFramework struct {
...
 		return nil, err
 	}
 
-	controlPlaneContext := ""
 	var controlPlaneClient client.Client
 	remoteClients := make(map[string]client.Client, 0)
 	t.Logf("Using config file: %s", configFile)
 
-	for name, _ := range config.Contexts {
+	for name := range config.Contexts {
```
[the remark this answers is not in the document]
Miles-Garnsey: I'm sticking my nose in here because I wanted to warn @adutra about the issues we've previously seen where tests end up getting run on the wrong cluster. Obviously this makes things quite dangerous for users who run production clusters, and people like myself who sometimes have temporary access to client clusters.

I agree with John that the control and data planes should have to be explicitly set, but I also think that it is safer to always use `build/kubeconfig` and make sure that no `kubectl` commands ever run against contexts from `~/.kube/config`. If a small bash script is required to copy named contexts from `~/.kube/config` into `build/kubeconfig` for tests run on remote clusters, then that might bridge the difference?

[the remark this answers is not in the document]
jsanda: I am fine with allowing a configurable kubeconfig file. We could require it to always be `build/kubeconfig`, but then if I want to be lazy, I might just copy `~/.kube/config`. Tools read from and write to the default kubeconfig file. On that basis I think it is reasonable to be able to use it provided we are being explicit about which clusters we are using.

As for a script to copy contexts, there are other elements that need would need to be copied. Yes it is doable, but it is additional maintenance. IMO, that extra maintenance isn't worth it as long as we are being explicit in our tests.

adutra:
> The test fixture manifests were intentionally write as simple YAML manifests so as to provide a source of examples that we can references for folks.

I wasn't aware of that, to me they were just test fixtures. I'd really suggest that we better separate test data from samples. We do have a `test/testdata/samples` folder but 1) I wonder why it was placed there and 2) I didn't change the files in that folder anyways.

> I'm not a fan of the templates. What happens not if but when we want to make other parts of the fixtures configurable? [...] I'm still in favor of making the default context names for the control plane and data planes more generic instead of using the ones that `kind` generates. I think that would make it easier to punt on making the `k8sContext` property configurable. 

OK, I will remove the templates and check if kind can generate the context names we want to use.

test/framework/e2e_framework.go
adutra: I am sorry, but I am failing to understand how you would like this to be. In fact, I'm failing to understand when/why the in-cluster config is required, and how tests are going to be executed against a cloud provider like GKE. I would suggest that we leave this part out of this PR, then someone else more skilled than me could tackle that part later.

jsanda:
>I wasn't aware of that, to me they were just test fixtures.

They are just test fixtures; however, I think tests can and should be treated like a form a documentation. That's what I really meant. The target audience will vary based on the type of test involved. The fact that the code involved is regularly executed (and for the moment let's assume stable), should give the user confidence that it is an accurate form of docs. That's how I tend to think of it.

test/framework/e2e_framework.go
jsanda: When you create a kind cluster it updates `~/.kube/config` with an entry like this:

```yaml
- cluster:
    certificate-authority-data: ...
    server: https://127.0.0.1:55055
  name: kind-k8ssandra-0
```
We don't want a localhost address for the api server, so we generate a kubeconfig using the in-cluster IP of the api server. This doesn't need to be done when using real clusters.

I'm fine though with addressing this in a follow up PR. It makes the work more incremental and PRs smaller which I think is a good thing.

test/framework/e2e_framework.go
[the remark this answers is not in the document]
adutra: I am going to add an allow-list of context names to create clients for.

adutra:
Ready for review. Summary of changes for reviewers:

- The kubeconfig file used to create remote clients and ClientConfigs is now configurable via a flag; default is `./build/kubeconfig` (the file generated by `setup-kind-multicluster.sh`). 
- The context names (control plane and data plantes) to use in e2e tests must now be provided explicitly, in order to avoid loading unwanted contexts.
- There is no more "in-cluster" kubeconfig files: the `create-clientconfig.sh` script detects when the server address is localhost and fixes that on the fly before creating the ClientConfig;
- Control plane and data plane names can _in theory_ be configured via flags. In practice however, their names cannot change as context names in fixture files are currently not configurable. We decided to fix it in a follow-up task.
- Context names in fixture files are currently hard-coded to `k8ssandra-0` and `k8ssandra-1`: these must be the names of the data planes to use.
- `setup-kind-multicluster.sh` has been modified to generate contexts with the hard-coded names `k8ssandra-0` and `k8ssandra-1`. If we implement configurable context names in fixtures, then we could revert this change.

adutra:
@jsanda and other reviewers: I discussed this PR this morning with @adejanovski and wee agreed that it's a bit frustrating to stop here and not being able to provide fully configurable context names.

I decided then to put this PR back in draft and I will work today towards adding a kustomization layer on top of the fixtures, as @jsanda suggested.

adutra:
Ready again for review, sorry for the delay. New summary of changes for reviewers:

* The kubeconfig file used to create e2e remote clients and ClientConfigs is now configurable via a flag; default is `./build/kind-kubeconfig` (the file generated by `setup-kind-multicluster.sh`).
* The context names (control plane and data plantes) to use in e2e tests must now be provided explicitly, in order to avoid loading unwanted contexts that might exist in the kubeconfig file.
* There is no more "in-cluster" kubeconfig files: the `create-clientconfig.sh` script detects when the server address is localhost and fixes that on the fly before creating the ClientConfig;
* Control plane and data plane names are fully configurable via flags.
* Context names in fixture files are kustomized to match what the user provided via the flags.
* Controller tests were also modified; context names can also be modified if required, although for now we use hard-coded names `cluster-1`, `cluster-2`, `cluster-3`.

test/framework/e2e_framework.go
```diff
@@ -35,98 +37,64 @@ import (
...
+	var validDataPlanes []string
+	for _, name := range dataPlanes {
+		if remoteClient, err := newRemoteClient(config, name); err != nil {
 			return nil, err
-		}
-
-		remoteClient, err := client.New(restCfg, client.Options{Scheme: scheme.Scheme})
-		if err == nil {
+		} else if remoteClient == nil {
+			t.Logf("ignoring invalid data plane context %v: %v", name, err)
```
jsanda, line 71: If the data plane context is invalid, do you think we should return an error here?

test/framework/fixtures.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	destDir := filepath.Join(fixturesDestDir, fixture.Name)
+	return f.kustomizeAndApply(destDir, namespace, f.ControlPlaneContext)
+}
+
+// getNumDcContexts computes the exact number of contexts to kustomize by counting the number of dcs/contexts in the
+// K8ssandraCluster spec for this fixture, if any. This is a bit hacky but unfortunately kustomize errors out if we
+// attempt to patch a non-existing path.
+func getNumDcContexts(fixture *TestFixture) (int, error) {
+	files, err := utils.ListFiles(fixture.fixtureDir(), "k8ssandra.yaml")
```
jsanda, line 54: Would it be simpler to use a yaml parse to do this?

test/e2e/suite_test.go
```diff
@@ -50,10 +50,48 @@ var (
...
 		datacenterUpdating      pollingConfig
 	}
+)
 
-	logKustomizeOutput = flag.Bool("logKustomizeOutput", false, "")
-	logKubectlOutput   = flag.Bool("logKubectlOutput", false, "")
-	imageTag           = flag.String("imageTag", "latest", "")
+var (
+	configFileFlag = flag.String(
+		"configFile",
```
jsanda, line 57: What do you think about renaming this to make it clear it is a kubeconfig file?

test/e2e/suite_test.go
```diff
@@ -50,10 +50,48 @@ var (
...
+	)
+	imageTag = flag.String(
+		"imageTag",
+		"latest",
+		"The k8ssandra-operator image to use.",
+	)
+)
+
+var (
+	configFile   string
```
jsanda, line 92: What do you think about renaming `configFile` to `kubeconfigFile` (or something similar)?

jsanda (approved):
I left some questions in my previous review but those are minor. I tested the changes with the setup-kind-multicluster.sh script. Looks good. I was able to kick off test against GKE clusters with the following cmd line:

`make E2E_TEST="TestOperator/CreateMultiDatacenterCluster" TEST_ARGS="-configFile=/path/to/my/kubeconfig -controlPlane=control-plane -dataPlanes=east" e2e-test
`
The test is going to fail though because we use a standard storage class in all the test fixtures. I don't have a standard storage class in my GKE clusters. The storage classes will vary from cluster to cluster.

I talked with @adutra agreed that the storage class customization can be addressed in a subsequent PR.

test/framework/e2e_framework.go
adutra: I was returning an error previously, then I bumped into this situation: I created a single cluster with `make single-up`, then I ran the `single-dc` test. It failed, because by default `dataPlanes` is `k8ssandra-1`, and so the framework attempts to create a client for `k8ssandra-1` (even if it is not required for the test being executed).

So I changed this as a convenience for developers. Otherwise you would have to run single dc tests with `-dataPlanes ''` each time.

test/framework/fixtures.go
adutra: I think the best tool for this would be yq. The following gives you the total number of `k8sContext` fields declared in the spec:

```
yq ea '.spec.cassandra.datacenters.[] | .k8sContext as $item ireduce (0; . +1)' k8ssandra.yaml
```

test/e2e/suite_test.go
adutra: Indeed, I will change this to `kubeconfigFile`.

test/framework/fixtures.go
adutra: The following is even better as it would give me which dcs have a k8sContext defined:

```
 > yq e '.spec.cassandra.datacenters.[] | .k8sContext ' k8ssandra.yaml 
null
kind-k8ssandra-1
```

With this I know that the kustomization must target only `.spec.cassandra.datacenters[1].k8sContext`.

I can create a ticket to implement that if you like.

jsanda: That would be great!

adutra: done, see #462

adutra:
> The storage classes will vary from cluster to cluster.

Created #463 for that.

Issue #282, opened 2022-01-10
Set verbosity level in logging statements


jsanda:
**What is missing?**
Most (maybe all) logging throughout the operator are done with `logger.Info()` and `logger.Error()`. A verbosity should be set for each statement, e.g., `logger.V(5).Info()`.

**Why do we need it?**
As the code base has grown there is a lot more logging output from the operator. A lot of the log statements are really only needed during dev/testing or debugging. By setting verbosity levels we give the user the ability to configure and tune logging output. 

This is critical from a usability standpoint IMO. 

It probably makes sense to configure the logging level via the operator configuration which means we need to do #63.

**Environment**

* K8ssandra Operator version:

    `Insert image tag or Git SHA here`
    v1.0.0-alpha.2

burmanm:
Kubernetes recommends max V(1) for debugging, not deeper.

jsanda:
So nothing greater than V(1)?

jsanda:
Here are logging guidelines for kubernetes dev docs https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md. It suggests V(4) for debug and a default level of V(2).

Issue #281, opened 2022-01-10
Do not log errors when returning errors from reconciliation


jsanda:
**What is missing?**
If we are aborting reconciliation with an error, then we should not call `logger.Error()` to log the error (in general as there might always be exceptions).

We have a common pattern throughout the operator with respect to error handling. If we have an error, we log the error and a message by calling `logger.Error(err, "some message")` and then abort reconciliation by returning the error.

Here are a couple examples. First from reaper_controller.go:

```go
// reaper_controller.go
logger.Info("Reconciling Reaper Deployment")

authVars, err := r.collectAuthVars(ctx, actualReaper, logger)
if err != nil {
	logger.Error(err, "Failed to collect Reaper auth variables")
	return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
}
```

Next from k8ssandracluster_controller.go:

```go
if err = r.setStatusForDatacenter(kc, actualDc); err != nil {
	logger.Error(err, "Failed to update status for datacenter")
	return result.Error(err), actualDcs
}
```

**Why do we need it?**
Here is example log output from the second example when there is an error:

```
2022-01-10T19:36:55.874Z	ERROR	controller.k8ssandracluster	Failed to update status for datacenter	{"reconciler group": "k8ssandra.io", "reconciler kind": "K8ssandraCluster", "name": "test", "namespace": "k8ssandra-operator", "K8ssandraCluster": "k8ssandra-operator/test", "CassandraDatacenter": "k8ssandra-operator/dc1", "K8SContext": "", "error": "status updated failed due to system failure"}
github.com/k8ssandra/k8ssandra-operator/controllers/k8ssandra.(*K8ssandraClusterReconciler).Reconcile
	/workspace/controllers/k8ssandra/k8ssandracluster_controller.go:87
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.0/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.0/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
[...]

PR #454, opened 2022-03-08
Remote restore support for Medusa

### LINKED ISSUE #450: K8SSAND-521 ⁃ Add control loop to sync CassandraBackup objects with storage bucket
In order to allow remote restores, we need to be able to sync the local CassandraBackup resources with the configured storage backend.

The sync should allow:

* Creating local missing resources from existing backups
* Deleting local resources for missing backups

A single control loop should be able to handle all 3 operations. The sync should be required explicitly, meaning that a new CRD should be created to allow requesting synchronization to the medusa-operator, and that CRD should contain a prefix so that a cluster can list backups stored for another cluster in a multi-tenant storage bucket.

The sync operation would:

* List the backups in the storage bucket, through a gRPC call
* Compare that list to the existing CassandraBackup resources
* Create a CassandraBackup object for each missing backup
* Delete local resources that don’t exist in the list of backups

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-521) by [Unito](https://www.unito.io)
┆epic: Remote Cluster Restore
┆friendlyId: K8SSAND-521
┆priority: Medium

### LINKED ISSUE #451: K8SSAND-522 ⁃ Remove the CassandraDatacenter spec from CassandraBackup objects
CassandraBackup resources shouldn’t store the CassandraDatacenter spec anymore but rather the topology object generated by Medusa.
This object will contain the list of live nodes at the time of the backup along with their token assignments. Sample topology object:

```java
{"ip-179-31-16-38.us-west-2.compute.internal": {"tokens": [-1051417974279770378, -112095142485141927, 9217379683620055011, 939597040143840721], "is_up": true}, 
 "ip-179-31-23-54.us-west-2.compute.internal": {"tokens": [-1080182033668782299, -1156652990105327691, 9134535641516355167, 9216775765860996507], "is_up": true}, 
 "ip-179-31-30-139.us-west-2.compute.internal": {"tokens": [-1018393144648419592, -1342532530495729638, 8846603596658721956, 8849342512812637165], "is_up": true}}
```

The CRD should also contain the storage bucket prefix, which is controlled by the namespace name.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-522) by [Unito](https://www.unito.io)
┆epic: Remote Cluster Restore
┆friendlyId: K8SSAND-522
┆priority: Medium

### LINKED ISSUE #452: K8SSAND-524 ⁃ Check restore mapping is valid before triggering restore
When using NetworkTopologyStrategy, the placement of replicas is dependent on the DC and rack layout. In this context, the replica for a token range will be placed on the node that owns the next token range in the next rack. We can't restore any node to any other node as we may end up putting several replicas in the same rack, and the nodes we'd restore to would possibly end up not containing data they're supposed be replica for. We can be rigid about the mapping and expect the same dc and rack names exclusively, or do a 1:1 mapping between DCs/Racks despite having different names (but expect the same number of both).

This check should be performed before attempting to restore a backup.

The backup definition will be used to map the backup hosts with the restore hosts and perform checks on whether or not the restore is possible:

* Same number of DCs
* Same number of nodes per DC
* Same rack distribution per DC

Nodes will be mapped 1:1 between the source and destination topologies (sorted alphabetically). If the mapping fails to associate nodes 1:1, the restore operation must fail.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-524) by [Unito](https://www.unito.io)
┆epic: Remote Cluster Restore
┆friendlyId: K8SSAND-524
┆priority: Medium

### LINKED ISSUE #453: K8SSAND-525 ⁃ Add bucket prefix and in-place/remote env variables for restores
The Medusa restore init containers need following new env variables to be set:

* Bucket prefix: path prefix to point to another cluster's backups in the storage bucket
* In Place/Remote: flag indicating whether the performed restore is in-place or remote.

These values will then be used to invoke medusa.restore_node in the medusa-restore init container. The init-container needs to deal with mapping between source and dest topologies to self determine which backup it needs to restore. Pre-checks in the operator will guarantee that a 1:1 mapping is possible and topologies are compatible.

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-525) by [Unito](https://www.unito.io)
┆epic: Remote Cluster Restore
┆friendlyId: K8SSAND-525
┆priority: Medium

adejanovski:
**What this PR does**:
Allow restoring a backup from another cluster (including backups created with Medusa outside of K8ssandra).
Requires these medusa PR to be merged first, and a Medusa release so that we can update the tag for the medusa image: [448](https://github.com/thelastpickle/cassandra-medusa/pull/448) and [449](https://github.com/thelastpickle/cassandra-medusa/pull/449).

The design is fully documented in `docs/medusa/remote-restore-design.md`.

Fixes #450
Fixes #451 
Fixes #452 

**Checklist**
- [x] Changes manually tested


apis/medusa/v1alpha1/cassandrabackup_types.go
```diff
@@ -45,6 +45,12 @@ type CassandraBackupSpec struct {
 	// +kubebuilder:validation:Enum=differential;full;
 	// +kubebuilder:default:=differential
 	Type BackupType `json:"backupType,omitempty"`
+
+	// Specifies if the object was created by a synchronization job.
+	// If so, reconciliation will be skipped.
+	// +kubebuilder:default:=false
+	// +kubebuilder:validation:optional
+	CreatedBySync bool `json:"createdBySync,omitempty"`
```
burmanm, line 53: I don't like this idea. I think all objects should be reconciled and synced to maintain state as Kubernetes is our user interface.

apis/medusa/v1alpha1/medusatask_types.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// MedusaTaskSpec defines the desired state of MedusaTask
+type MedusaTaskSpec struct {
```
burmanm, line 27: Naming question. We have the "Cassandra<Backup/Restore>" objects as well as "Medusa<Name>" objects. Could we use a single naming system?

controllers/medusa/cassandrabackup_controller.go
```diff
@@ -75,6 +75,26 @@ func (r *CassandraBackupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+	}
+
+	pods, err := r.getCassandraDatacenterPods(ctx, cassdc, logger)
+	if err != nil {
+		logger.Error(err, "Failed to get datacenter pods")
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+	}
+
+	// If the CassandraBackup object was created by a sync operation, we mustn't trigger a backup.
```
burmanm, line 92: The issue with this approach is that I don't think "backup" definition and "backupJob" should be the same. That would solve this mentioned issue. Also, if metadata per object is necessary, we could use annotation instead of polluting the CRD with extra fields.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+	}
+
+	task := instance.DeepCopy()
+
+	// First check to see if the task is already in progress
+	if !task.Status.StartTime.IsZero() {
+		// If there is anything in progress, simply requeue the request
+		if len(task.Status.InProgress) > 0 {
+			logger.Info("Tasks already in progress")
```
burmanm, line 79: This and next logger lines, without other info in the log statement the reader can't tell which restore process this is related to. So we need some KeyValues in the .Info()

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+			return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
+		}
+
+		logger.Info("task complete")
+
+		// Set the finish time
+		// Note that the time here is not accurate, but that is ok. For now we are just
+		// using it as a completion marker.
+		patch := client.MergeFrom(task.DeepCopy())
```
burmanm, line 88: task is already DeepCopy of instance. Do we really need double copy?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+		// Schedule a sync if the task is a purge
+		if task.Spec.Operation == medusav1alpha1.OperationTypePurge {
+			r.scheduleSyncForPurge(task)
+		}
+
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	// If the task is already finished, there is nothing to do.
+	if taskFinished(task) {
```
burmanm, line 104: Shouldn't this be before the previous if statement? StartTime is not zero for finished jobs either?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+
+func (r *MedusaTaskReconciler) getCassandraDatacenterPods(ctx context.Context, cassdc *cassdcapi.CassandraDatacenter, logger logr.Logger) ([]corev1.Pod, error) {
+	podList := &corev1.PodList{}
+	labels := client.MatchingLabels{cassdcapi.DatacenterLabel: cassdc.Name}
+	if err := r.List(ctx, podList, labels); err != nil {
+		logger.Error(err, "failed to get pods for cassandradatacenter", "CassandraDatacenter", cassdc.Name)
+		return nil, err
+	}
+
+	pods := make([]corev1.Pod, 0)
```
burmanm, line 143: ``, len(podList.Items)``

pkg/medusa/reconcile.go
```diff
@@ -164,6 +167,9 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
 	setImage(medusaSpec.ContainerImage, medusaContainer)
 	medusaContainer.SecurityContext = medusaSpec.SecurityContext
 	medusaContainer.Env = medusaEnvVars(medusaSpec, dcConfig, logger, "GRPC")
+	medusaContainer.Ports = []corev1.ContainerPort{
+		{ContainerPort: 50051, Name: "grpc"},
+	}
```
adejanovski, line 172: This for debugging, remove

adejanovski: Add it as a debugging option

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+
+	if err := r.Status().Patch(ctx, task, patch); err != nil {
+		logger.Error(err, "Failed to patch status")
+		// We received a stale object, requeue for next processing
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+	}
+
+	// Invoke the purge operation on all pods, in the background
+	go func() {
```
burmanm, line 232: How is the verification done if this go func is still alive or failed? What if the k8ssandra-operator crashes while this is running?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
```
burmanm, line 166: Same as next go func() question. Liveness check.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+								CassandraDatacenter: task.Spec.CassandraDatacenter,
+								Type:                medusav1alpha1.BackupType("differential"),
+								CreatedBySync:       true,
+							},
+						}
+						if err := r.Create(ctx, backupResource); err != nil {
+							logger.Error(err, "failed to create backup", "CassandraBackup", backup.BackupName)
+							return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+						} else {
+							logger.Info("Created Cassandra Backup", "Backup", backupResource)
```
jsanda, line 311: I know you talked about why you have the for loop on our call, but I don't recall the details so bear with my question if it was already asked. I don't think we need to fetch the CassandraBackup object at all. The `Create` call above at line 307 will initialize the necessary metadata fields to apply a follow up patch. In particular `ResourceVersion` will be set. You should be able to simply follow up the `Create` with the `Status().Patch()` call.

jsanda (commented):
A partial review. More to follow tomorrow...

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+		logger.Info("task complete")
+
+		// Set the finish time
+		// Note that the time here is not accurate, but that is ok. For now we are just
+		// using it as a completion marker.
+		patch := client.MergeFrom(task.DeepCopy())
+		task.Status.FinishTime = metav1.Now()
+		if err := r.Status().Patch(ctx, task, patch); err != nil {
+			logger.Error(err, "failed to patch status with finish time")
+			return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
jsanda, line 92: No need to set `RequeueAfter` when returning a non-nil error.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+	} else {
+		return ctrl.Result{}, fmt.Errorf("unsupported operation %s", task.Spec.Operation)
+	}
+}
+
+func (r *MedusaTaskReconciler) getCassandraDatacenterPods(ctx context.Context, cassdc *cassdcapi.CassandraDatacenter, logger logr.Logger) ([]corev1.Pod, error) {
+	podList := &corev1.PodList{}
+	labels := client.MatchingLabels{cassdcapi.DatacenterLabel: cassdc.Name}
+	if err := r.List(ctx, podList, labels); err != nil {
+		logger.Error(err, "failed to get pods for cassandradatacenter", "CassandraDatacenter", cassdc.Name)
```
jsanda, line 139: In general I don't think we should log an ERROR statement when returning an error. It can create noise in the logs. See #281 for more info. If we need to log something it is probably for debugging purposes. See #282 for details.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+	patch := client.MergeFrom(task.DeepCopy())
+	task.Status.StartTime = metav1.Now()
+	for _, pod := range pods {
+		task.Status.InProgress = append(task.Status.InProgress, pod.Name)
+	}
+
+	if err := r.Status().Patch(ctx, task, patch); err != nil {
+		logger.Error(err, "Failed to patch status")
+		// We received a stale object, requeue for next processing
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
jsanda, line 162: No need to set `RequeueAfter` when returning a non-nil error.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+					task.Status.Finished = append(task.Status.Finished, purgeResponse)
+				} else {
+					task.Status.Failed = append(task.Status.Failed, pod.Name)
+				}
+			}()
+		}
+		wg.Wait()
+		logger.Info("finished task operations")
+		if err := r.Status().Patch(context.Background(), task, patch); err != nil {
+			logger.Error(err, "failed to patch status", "MedusaTask", fmt.Sprintf("%s/%s", task.Spec.Operation, task.Namespace))
```
jsanda, line 208: `fmt.Sprintf("%s/%s", task.Spec.Operation, task.Namespace)` is used in multiple places. Maybe implement the Stringer interface?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+	patch := client.MergeFrom(task.DeepCopy())
+	task.Status.StartTime = metav1.Now()
+	for _, pod := range pods {
+		task.Status.InProgress = append(task.Status.InProgress, pod.Name)
+	}
+
+	if err := r.Status().Patch(ctx, task, patch); err != nil {
+		logger.Error(err, "Failed to patch status")
+		// We received a stale object, requeue for next processing
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
jsanda, line 228: No need to set `RequeueAfter` when returning a non-nil error.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+						return ctrl.Result{}, err
+					}
+				}
+			}
+
+			// Delete backups that don't exist remotely but are found locally
+			localBackups := &medusav1alpha1.CassandraBackupList{}
+			if err = r.List(ctx, localBackups, client.InNamespace(task.Namespace)); err != nil {
+				logger.Error(err, "failed to list backups")
+				return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
jsanda, line 353: No need to set `RequeueAfter` when returning a non-nil error.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,470 @@
...
+		}
+	}()
+
+	return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
+}
+
+func (r *MedusaTaskReconciler) syncOperation(ctx context.Context, task *medusav1alpha1.MedusaTask, pods []corev1.Pod, logger logr.Logger) (reconcile.Result, error) {
+	logger.Info("Starting sync operation")
+	syncStartTime := metav1.Now()
+	for _, pod := range pods {
```
jsanda, line 279: Won't we be doing duplicate work here by iterating over pods? It seems like we should first query for remote backups, which may involve queries across multiple pods, and then iterate across the backups. That's the pattern you have implemented for querying k8s backups in the following for loop.

apis/medusa/v1alpha1/cassandrabackup_types.go
adejanovski: I'll shortly send a new PR which separates the action of taking a backup from the backup itself, introducing the MedusaBackupJob and MedusaBackup APIs.

apis/medusa/v1alpha1/medusatask_types.go
adejanovski: good point, the upcoming PR will use the `Medusa` prefix for the object names.

controllers/medusa/cassandrabackup_controller.go
adejanovski: Right, that change is coming in the next PR 👍

jsanda (commented):
I assume we want to support the scenario of restoring from one bucket and then storing subsequent backups in a different bucket. Do the API changes support that? It doesn't look like it.

adejanovski:
> I assume we want to support the scenario of restoring from one bucket and then storing subsequent backups in a different bucket. Do the API changes support that? It doesn't look like it.

Not yet, it's part of another upcoming ticket: https://github.com/k8ssandra/k8ssandra-operator/issues/453

controllers/medusa/medusatask_controller.go
adejanovski: isn't it what line 59 does?

```
logger := log.FromContext(ctx).WithValues("MedusaTask", req.NamespacedName)
```

controllers/medusa/medusatask_controller.go
adejanovski: I'm following the conventions we've used throughout the project when patching CRs.
Do you think we're wrong to do so?
I seem to remember that patches wouldn't be applied properly without this, but I could be wrong.

controllers/medusa/medusatask_controller.go
adejanovski: yes, it would be better to move it before the previous if indeed 👍 
I'll move it.

controllers/medusa/medusatask_controller.go
adejanovski: 👍

controllers/medusa/medusatask_controller.go
adejanovski: Then we lose track of the operation. For better or worse, this is still what we currently have for backups as well (I re-used the same code structure).
We could definitely improve the resiliency by making the calls non blocking with a way to check the state asynchronously, like we're about to do with backups.
The tasks we run here should be far more short lived and easy to rerun in case of failure though, so it's less worrisome than backups and could be dealt with in a future PR.

controllers/medusa/medusatask_controller.go
adejanovski: Same answer :)

controllers/medusa/medusatask_controller.go
adejanovski: I'll try again, but I remember Status wouldn't be updated when I tried to update it without reading the object back from storage. This is why I went with this solution (and the for loop was there to retry because of cache issues that wouldn't return the object from the first attempt).
I'll post back once I have some results.

controllers/medusa/medusatask_controller.go
adejanovski: `return ctrl.Result{}, err` will reschedule a reconcile later?

controllers/medusa/medusatask_controller.go
adejanovski: I saw you wanted us to remove these loggings. The problem is that the error printed out if we don't put our own logging is usually more obscure and doesn't give details allowing to know precisely what failed. Technically we know what failed, but it's harder to know what we were trying to do when it failed.
I'm ok with some repetition in the logs if that allows more clarity during troubleshooting.

controllers/medusa/medusatask_controller.go
adejanovski: 👍

controllers/medusa/medusatask_controller.go
adejanovski: sounds good 👍

controllers/medusa/medusatask_controller.go
adejanovski: 👍

controllers/medusa/medusatask_controller.go
adejanovski: 👍

controllers/medusa/medusatask_controller.go
adejanovski: We're just iterating until we find one pod that answers back correctly (should be the first one).
We return as soon as the task is completed without moving to the next pod.
`GetBackups()` will return the backups from all nodes which is why we need a single pod to respond here.

controllers/medusa/medusatask_controller.go
jsanda: Yup, with rate limiting. Here is the relevant code from the controller:

```go
	result, err := c.Reconcile(ctx, req)
	switch {
	case err != nil:
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelError).Inc()
		log.Error(err, "Reconciler error")
	case result.RequeueAfter > 0:
		// The result.RequeueAfter request will be lost, if it is returned
		// along with a non-nil error. But this is intended as
		// We need to drive to stable reconcile loops before queuing due
		// to result.RequestAfter
		c.Queue.Forget(obj)
		c.Queue.AddAfter(req, result.RequeueAfter)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelRequeueAfter).Inc()
	case result.Requeue:
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelRequeue).Inc()
	default:
		// Finally, if no error occurs we Forget this item so it does not
		// get queued again until another change happens.
		c.Queue.Forget(obj)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelSuccess).Inc()
	}
}
```
Notice that when a non-nil error is returned the controller calls `c.Queue.AddRateLimited(req)`. See [here](http://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L287-L287) for the full source.

controllers/medusa/medusatask_controller.go
jsanda: The error printed out should include your error message. What isn't included though is a stack trace which can be problematic when the same error message is used in multiple places. 

Whatever we do or change (if anything) for logging is a separate concern from this PR, so I think we're good to go.

controllers/medusa/medusatask_controller.go
adejanovski: cool, I'll do the changes 👍

controllers/medusa/medusatask_controller.go
adejanovski: You were right @jsanda, I can patch the status straight away without needing to re-read the object from the server.
The code is much simpler now 👍

adejanovski:
@jsanda, I made the adjustments [in Medusa](https://github.com/thelastpickle/cassandra-medusa/pull/449/commits/e819aaa87d043f8843958209783c9418042bd002) so that we can support both the case where IPs aren't resolved to hostnames and the case when they are.
E2E tests are all passing and the PR is ready for another review.

apis/medusa/v1alpha1/cassandrabackup_types.go
```diff
@@ -18,20 +18,13 @@ package v1alpha1
```
jsanda, line 18: If we intend to replace this api with medusabackup_types, then I think we need to discuss how we want to handle breaking API changes. That is _not_ a discussion that I want to have in the PR review 😄 Do you think though that some comments should be added to indicate that the API is deprecated?

controllers/medusa/medusabackupjob_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			}
+			if err := r.Create(ctx, backupResource); err != nil {
+				logger.Error(err, "failed to create backup", "MedusaBackup", backup.ObjectMeta.Name)
+				return err
+			} else {
+				logger.Info("Created Medusa Backup", "Backup", backupResource)
+				// Read the backup again. Multiple attempts may be necessary due to caches, which is why the Get operation is in a loop.
+				backupInstance := &medusav1alpha1.MedusaBackup{}
+				found := false
+				for stay, timeout := true, time.After(10*time.Second); stay; {
```
jsanda, line 237: Blocking like this in a controller is an anti-pattern. By default each controller is configured to use a single goroutine. More importantly the loop is unnecessary. You can simply apply the patch after creating the object.

.github/workflows/kind_e2e_tests.yaml
```diff
@@ -61,6 +61,7 @@ jobs:
           - CreateSingleReaper
           - CreateReaperAndDatacenter
           - CreateSingleMedusa
```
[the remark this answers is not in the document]
adejanovski: I think that as long as the old APIs aren't removed, we should keep the tests because we can break the features.

.github/workflows/kind_multicluster_e2e_tests.yaml
```diff
@@ -66,6 +66,7 @@ jobs:
           - CreateMultiReaper
           - ClusterScoped/MultiDcMultiCluster
           - CreateMultiMedusa
```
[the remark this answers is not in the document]
adejanovski: Same answer, we should probably keep those tests as long as we'll have the APIs available.

Makefile
```diff
@@ -439,3 +439,7 @@ install-kuttl:
 mocks:
 	mockery --dir=./pkg/cassandra --output=./pkg/mocks --name=ManagementApiFacade
 	mockery --dir=./pkg/reaper --output=./pkg/mocks --name=Manager  --filename=reaper_manager.go --structname=ReaperManager
+
+PHONY: protobuf-code-gen
+protobuf-code-gen:
```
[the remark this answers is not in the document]
adejanovski: sounds good 👍

apis/medusa/v1alpha1/cassandrabackup_types.go
adejanovski: very much so. I had looked in the kubebuilder book if there was any annotation we could put to mark stuff as deprecated, with no luck.
But I just came across this: https://github.com/kubernetes-sigs/kubebuilder/issues/2116

Turns out the annotation exists but isn't documented: //+kubebuilder:deprecatedversion:warning="example.com/v1alpha1 CronTab is deprecated"`

I'll use this.

controllers/medusa/medusabackupjob_controller.go
adejanovski: very right, I forgot I had created a loop there as well.
I'll remove it.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,442 @@
...
+	} else if task.Spec.Operation == medusav1alpha1.OperationTypeSync {
+		return r.syncOperation(ctx, task, pods, logger)
+	} else if task.Spec.Operation == medusav1alpha1.OperationTypePrepareRestore {
+		return r.prepareRestoreOperation(ctx, task, pods, logger)
+	} else {
+		return ctrl.Result{}, fmt.Errorf("unsupported operation %s", task.Spec.Operation)
+	}
+}
+
+func (r *MedusaTaskReconciler) getCassandraDatacenterPods(ctx context.Context, cassdc *cassdcapi.CassandraDatacenter, logger logr.Logger) ([]corev1.Pod, error) {
```
jsanda, line 135: Looks like this is identical to the  `getCassandraDatacenterPods` in medusabackupjob_controller.go. Can we refactor this into a common, shared method or func?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,442 @@
...
+		if err := r.Status().Patch(context.Background(), task, patch); err != nil {
+			logger.Error(err, "failed to patch status", "MedusaTask", fmt.Sprint(task))
+		}
+	}()
+
+	return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
+}
+
+func (r *MedusaTaskReconciler) prepareRestoreOperation(ctx context.Context, task *medusav1alpha1.MedusaTask, pods []corev1.Pod, logger logr.Logger) (reconcile.Result, error) {
+	logger.Info("Starting prepare restore operations")
```
jsanda, line 216: It would be good to include the datacenter name in log statement.

adejanovski:
@jsanda, I've made the requested changes. Let me know what you think.
Whatever happens, I'll need to merge the associated PRs in the Medusa project and changes the tag accordingly here.

Also, I'd like to point out I've removed the backup name in the spec of MedusaBackupJob as I feel it's redundant with the `meta.name` for that same object. Hence, the MedusaBackupJob name will be the same as the MedusaBackup name (which only has a `meta.name` as well and no more `spec.backupName`) and also as the backup name in the object storage backend.

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,450 @@
...
+
+	// Invoke the purge operation on all pods, in the background
+	go func() {
+		wg := sync.WaitGroup{}
+
+		// Mutex to prevent concurrent updates to the Status object
+		backupMutex := sync.Mutex{}
+		patch := client.MergeFrom(task.DeepCopy())
+
+		for _, p := range pods {
```
jsanda, line 239: The control flow in this loop in nearly identical to the control flow in `purgeOperation` as well as in `MedusaBackupJobReconciler.Reconcile`. The status updates appear to be the same as well. I realize that in one case we are updating a MedusaTaskStatus and in other case we are updating a MedusaBackupJobStatus. How about moving the common fields into shared, common struct? And maybe define an interface through which the status updates are made. Both MedusaTaskStatus and MedusaBackupJobStatus would implement the interface.

.github/workflows/kind_e2e_tests.yaml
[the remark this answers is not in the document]
jsanda: Agreed. When I wrote my initial comment I was thinking that apis are being removed, but that would come later.

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -0,0 +1,294 @@
...
+	cassdcKey := types.NamespacedName{Namespace: request.RestoreJob.Namespace, Name: request.Datacenter.Name}
+	cassdc := &cassdcapi.CassandraDatacenter{}
+	err = r.Get(ctx, cassdcKey, cassdc)
+	if err != nil {
+		logger.Error(err, "failed to get cassandradatacenter", "CassandraDatacenter", cassdcKey)
+		return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
+	}
+
+	// Prepare the restore by placing a mapping file in the Cassandra data volume.
+	if !request.RestoreJob.Status.RestorePrepared {
```
jsanda, line 79: We don't need this for an in-place restore do we? I haven't tested but if I recall I want to say it is fine to do; however, that is a good bit of overhead that could otherwise be avoided.

controllers/medusa/medusatask_controller.go
adejanovski: sounds good

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	pods := make([]corev1.Pod, 0)
+	pods = append(pods, podList.Items...)
+
+	return pods, nil
+}
+
+func (r *MedusaTaskReconciler) purgeOperation(ctx context.Context, task *medusav1alpha1.MedusaTask, pods []corev1.Pod, logger logr.Logger) (reconcile.Result, error) {
+	logger.Info("Starting purge operations")
```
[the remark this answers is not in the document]
adejanovski: k, I need to move the logging further in the code so that we can access the info after reading the task object.

controllers/medusa/medusarestorejob_controller.go
jsanda: >In our future design, it's the operator that will compute it based on the host names

Will these changes be included in this PR or a follow up?

controllers/medusa/medusatask_controller.go
```diff
@@ -0,0 +1,450 @@
...
+		patch := client.MergeFrom(task.DeepCopy())
+		task.Status.FinishTime = metav1.Now()
+		if err := r.Status().Patch(ctx, task, patch); err != nil {
+			logger.Error(err, "failed to patch status with finish time")
+			return ctrl.Result{}, err
+		}
+
+		// Schedule a sync if the task is a purge
+		if task.Spec.Operation == medusav1alpha1.OperationTypePurge {
+			r.scheduleSyncForPurge(task)
```
jsanda, line 103: Need to handle the error returned here.

Something else I thought about while thinking about this part of the code is cleanup. We need an automatic cleanup of completed tasks. This is done for the Cassandra tasks in cass-operator. 

If the CassandraDacenter on which the operation was performed is deleted, should the corresponding tasks be deleted as well? I am inclined to say yes. If we are in agreement on that then we need either owner references or some new finalizer logic.

controllers/medusa/controllers_test.go
```diff
@@ -67,6 +95,66 @@ func TestMedusaBackupRestore(t *testing.T) {
...
+
+	reconcilerConfig := config.InitConfig()
+
+	reconcilerConfig.DefaultDelay = 100 * time.Millisecond
+	reconcilerConfig.LongDelay = 300 * time.Millisecond
+
+	medusaClientFactory = NewMedusaClientFactory()
+
+	err := testEnv.Start(ctx, t, func(controlPlaneMgr manager.Manager, clientCache *clientcache.ClientCache, clusters []cluster.Cluster) error {
+		err := (&k8ssandractrl.K8ssandraClusterReconciler{
```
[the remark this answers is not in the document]
adejanovski: yes, that's a lack of knowledge on my side, sorry. I couldn't figure out how to isolate them properly in the tests.

controllers/medusa/medusarestorejob_controller.go
adejanovski: Follow up PR, in this PR Medusa computes the mappings, including the in place/remote nature.

controllers/medusa/medusatask_controller.go
adejanovski: You're very right. Also, it's not a sync anymore as I've changed this part into directly creating the MedusaBackup object instead.
I'll rework that part.

controllers/medusa/medusatask_controller_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	require.Eventually(func() bool {
+		updated := &api.MedusaTask{}
+		err := f.Get(ctx, purgeKey, updated)
+		if err != nil {
+			t.Logf("failed to get purge task: %v", err)
+			return false
+		}
+
+		return !updated.Status.FinishTime.IsZero() && updated.Status.Finished[0].NbBackupsPurged == 2
```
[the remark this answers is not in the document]
adejanovski: 👍

controllers/medusa/medusatask_controller.go
adejanovski: I was confusing with some code from the backup job controller. This is still necessary and indeed I need to handle the errors.

apis/medusa/v1alpha1/cassandrabackup_types.go
jsanda: perfect 🙂

controllers/medusa/medusatask_controller.go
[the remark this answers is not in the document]
jsanda: What about adding the key/value pair to the logger in the Reconcile method?

controllers/medusa/medusatask_controller.go
jsanda: I forgot to include my main point 🤦‍♂️ The control logic is duplicated in several places. It would be good to look at deduplicating it. Doing so would handling both MedusaTaskStatus and MedusaBackupJobStatus, hence my suggestions for the common struct.

jsanda: @adejanovski I pushed some initial changes [here](https://github.com/jsanda/k8ssandra-operator/blob/medusa-pod-ops/controllers/medusa/medusatask_controller.go#L217) to illustrate what I had in mind. I introduced a `executePodOperations` which encapsulates all of the control flow and goroutines for making calls against individual pods. Some more work is needed to handle the additional status updates in `purgeOperation`. With the status-related refactoring I mentioned in my earlier [comment](https://github.com/k8ssandra/k8ssandra-operator/pull/454#discussion_r858678073) you should be able reuse this  in the backup controller as well.

To be clear though, my main observation is that the control flow for executing operations against pod is duplicated in multiple places.

Makefile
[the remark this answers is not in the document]
adejanovski: I've added a comment on line 443

controllers/medusa/medusatask_controller.go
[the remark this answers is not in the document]
adejanovski: Great idea. Done in my upcoming commit 👍

controllers/medusa/medusatask_controller.go
adejanovski: Cool, I've refactored the code using your experiment code. Haven't pushed it as far as refactoring the status because I'd like to work on using common code between task and backup operations in a subsequent PR.
There's a lot being done here already IMO and we should scope this out.
I'll create a ticket for this if you agree.

jsanda: Sorry for the delayed response. Follow up ticket works for me.

adejanovski: Cool, I've create the issue: https://github.com/k8ssandra/k8ssandra-operator/issues/541

test/e2e/medusa_test.go
```diff
@@ -36,6 +36,22 @@ func createSingleMedusa(t *testing.T, ctx context.Context, namespace string, f *
 	verifyRestoreFinished(t, ctx, f, dcKey, backupKey)
 }
 
+func createSingleMedusaJob(t *testing.T, ctx context.Context, namespace string, f *framework.E2eFramework) {
```
jsanda, line 39: I realize that the setup will be a bit more involved that existing tests, but we should have a e2e test that covers remote restores. Would you create a follow up ticket for it?

pkg/medusa/requests.go
```diff
@@ -107,6 +112,56 @@ func (f *factory) NewRestoreRequest(ctx context.Context, restoreKey types.Namesp
...
+	err := f.Get(ctx, restoreKey, restore)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, &ctrl.Result{}, nil
+		}
+		f.Log.Error(err, "Failed to get MedusaRestoreJob")
+		return nil, &ctrl.Result{RequeueAfter: 10 * time.Second}, err
+	}
+
+	backup := &api.MedusaBackup{}
```
jsanda, line 126: This part needs to be updated. I hit this error when I tried restoring to a new cluster. I had to manually create a sync task to avoid this.

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -0,0 +1,310 @@
...
+func (r *MedusaRestoreJobReconciler) prepareRestore(ctx context.Context, request *medusa.RestoreRequest) (bool, error) {
+	// Create a prepare_restore medusa task to create the mapping files in each pod.
+	// Returns true if the reconcile needs to be requeued, false otherwise.
+	prepare := &medusav1alpha1.MedusaTask{}
+	if err := r.Client.Get(context.Background(), types.NamespacedName{Name: request.RestoreJob.Status.RestoreKey, Namespace: request.RestoreJob.Namespace}, prepare); err != nil {
+		if errors.IsNotFound(err) {
+			// Create the sync task
+			prepare = &medusav1alpha1.MedusaTask{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      request.RestoreJob.Status.RestoreKey,
```
jsanda, line 246: Can we change the task name to something more descriptive, maybe something like `<cluster-name>-<dc-name>-prepare-restore`?  Seeing task objects with UUIDs for their names isn't very helpful. Alternatively we could update the CRD to include the operation type and cassdc name in the output for `kubectl get`.

apis/medusa/v1alpha1/medusarestorejob_types.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// MedusaRestoreJobSpec defines the desired state of MedusaRestoreJob
+type MedusaRestoreJobSpec struct {
```
jsanda, line 27: For remote restores shouldn't we allow the bucket from which we are restoring to be specified separately instead of just reusing what is specified for backups? I am thinking that I have some CI/CD pipeline where I provision clusters from a known, good backup I would probably want to keep their storage buckets separate.

test/e2e/medusa_test.go
adejanovski: Sure thing

adejanovski: https://github.com/k8ssandra/k8ssandra-operator/issues/542

pkg/medusa/requests.go
adejanovski: it's the process actually. Did you expect the sync to be done automatically?
Backups won't get synced unless you run a `sync` MedusaTask, then you'll have them created as MedusaBackup CR which can be restored.

apis/medusa/v1alpha1/medusarestorejob_types.go
adejanovski: The only thing we could do with the way Medusa is currently designed, is to specify a different prefix for the sync and the restore, which would then be specified at sync and restore time. It probably means that the MedusaBackup object should contain that prefix in the spec.
This is part of a follow up ticket: https://github.com/k8ssandra/k8ssandra-operator/issues/453

controllers/medusa/medusarestorejob_controller.go
adejanovski: to be fair, it'll be created in the namespace of the cluster that's being restored.
If multiple restores are done, we need a name that distinguish them, and using the restore key seemed like a good idea.
I'm afraid that we may create names that are too long if we stuff in things like the cluster name 😕 

> Alternatively we could update the CRD to include the operation type and cassdc name in the output for kubectl get.

Would you add that in the status of the CRD?

pkg/medusa/requests.go
jsanda: I think I misread the code last night and thought that the restore controller created the sync task as a preliminary step. It creates the prepare restore task. I'm good here.

controllers/medusa/medusarestorejob_controller.go
jsanda: >I'm afraid that we may create names that are too long if we stuff in things like the cluster name

Definitely a valid concern.

>Would you add that in the status of the CRD?

Not in the status. There is a `kubebuilder:printcolumn:name` annotation. See https://github.com/k8ssandra/k8ssandra-operator/blob/main/apis/stargate/v1alpha1/stargate_types.go#L283 for examples. Would make for a nice follow up enhancement. I don't think it's needed for this PR.

.github/workflows/kind_multicluster_e2e_tests.yaml
jsanda, line 68: `CreateMultiMedusa` is matching both `CreateMultiMedusaOld` and `CreateMultiMedusaJob`, so both tests are being run and `CreateMultiMedusaJob` is being run twice. It should be changed to `CreateMultiMedusaOld`.

jsanda (changes requested):

.github/workflows/kind_multicluster_e2e_tests.yaml
adejanovski: Damn, I was pretty sure I fixed this already 🤔 
Good catch!

jsanda (approved):
I'm gonna go ahead and approve so we don't have to wait until I am back online. I spent some time investigating the e2e test failure, and I think we are good to go with a small change. @adejanovski well done 👏

PR #466, opened 2022-03-15
Make fixtures cloud-ready

### LINKED ISSUE #463: K8SSAND-1322 ⁃ Make fixtures cloud-ready
Currently storage class names are set to `standard` in fixture files, and racks are set to `rack1`, `rack2` etc.. 

This won't work if we execute the tests in the cloud. The test is going to fail though because storage classes will vary from cluster to cluster, and such zones do not exist.

We need to add kustomization for storage class names, racks, and maybe more.

This could be done with a new `-storageClass` flag, or maybe with something more future-proof, e.g. `-cloudProvider GKE`.

We'll need to keep in mind that storage classes can be provided at 2 levels:

* cluster-level: `spec.cassandra.storageConfig.cassandraDataVolumeClaimSpec.storageClassName`
* dc-level: `spec.cassandra.datacenters.storageConfig.cassandraDataVolumeClaimSpec.storageClassName`

This makes the usage of yq even more appealing (see #462).

_Originally posted by @jsanda in_ [_https://github.com/k8ssandra/k8ssandra-operator/pull/446#pullrequestreview-907495606_](https://github.com/k8ssandra/k8ssandra-operator/pull/446#pullrequestreview-907495606)

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1322) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1322
┆priority: Medium

adutra:
**What this PR does**:

This PR makes it possible to run e2e tests against cloud-hosted clusters.

The tests can be run with the following make target:

```shell
make gke-e2e-test GKE_US_EAST_IP=w.x.y.z GKE_US_NORTH_IP=w.x.y.z
``` 

There are a few failures currently, most of them related to heap sizes. I'm still investigating.

Fixes #463 

**Checklist**
- [x] Changes manually tested


Makefile
```diff
@@ -439,3 +441,47 @@ install-kuttl:
 mocks:
 	mockery --dir=./pkg/cassandra --output=./pkg/mocks --name=ManagementApiFacade
 	mockery --dir=./pkg/reaper --output=./pkg/mocks --name=Manager  --filename=reaper_manager.go --structname=ReaperManager
+
+
+# The image base name used in GKE clusters.
+GKE_IMAGE_TAG_BASE ?= us-docker.pkg.dev/community-ecosystem/$(IMAGE_TAG_BASE)
```
jsanda, line 447: Is it necessary to push images into a GKE Artifact repo? @bradfordcp where are we at with the quay.io migration? Once that is complete, could we just pull from there?

Makefile
```diff
@@ -439,3 +441,47 @@ install-kuttl:
...
+# Builds the project, builds the operator image, pushes the operator image to GKE Artifact
+# Repository, then runs e2e tests using two contexts: k8ssandra-ci-us-east and
+# k8ssandra-ci-us-north.
+# The clusters must be up and running and your ~/.kube/config file must contain the corresponding
+# contexts.
+# For security reasons, instructions for setting up the clusters are not checked in version control.
+# You must invoke this target with 2 env vars, GKE_US_EAST_IP and GKE_US_NORTH_IP containing 2
+# valid and properly configured external IPs, e.g.:
+# make gke-e2e-test GKE_US_EAST_IP=w.x.y.z GKE_US_NORTH_IP=w.x.y.z
+gke-e2e-test: build manifests
```
jsanda, line 458: This almost feels like it would be better suited as a separate script since it requires a bunch of setup, a lot of which isn't even documented.

test/framework/fixtures.go
```diff
@@ -38,67 +42,100 @@ func NewTestFixture(name, k8sContext string) *TestFixture {
...
+
+func getFixtureStargate(fixture *TestFixture) (*stargateapi.Stargate, error) {
+	obj, err := evalAndUnmarshal(fixture, "Stargate", &stargateapi.Stargate{})
+	if obj == nil {
+		return nil, err
+	}
+	return obj.(*stargateapi.Stargate), err
+}
+
+func evalAndUnmarshal(fixture *TestFixture, kind string, obj interface{}) (interface{}, error) {
```
jsanda, line 130: Maybe we should make the fixture a first class citizen in the yaml 🤔

test/framework/fixtures.tmpl
```diff
@@ -0,0 +1,209 @@
...
+{{ with .K8ssandraCluster }}
+  - target:
+      kind: K8ssandraCluster
+    patch: |-
+
+      - op: replace
+        path: /spec/auth
+        value: {{ if .Auth }}{{ .Auth }}{{ else }}true{{ end }}
+
+  {{ if and $.Storage .Cassandra.StorageConfig }}
```
jsanda, line 18: Is it possible to avoid this additional templating? Could we instead just have kustomizations for each cloud provider that applies whatever changes necessary? Kustomize can be a PITA, but from a usability perspective, I think it would be a win in that users could simply run `kustomize build` to see an example manifest for GKE.

jsanda (changes requested):
Can you provide a minimal set of instructions for running a single test against a GKE cluster? For now pick a test that doesn't require ingress or backup/restore.

adutra:
> Can you provide a minimal set of instructions for running a single test against a GKE cluster? For now pick a test that doesn't require ingress or backup/restore.

Sure, you can run a single test with:

```
make gke-e2e-test GKE_US_EAST_IP=$GKE_US_EAST_IP GKE_US_NORTH_IP=$GKE_US_NORTH_IP E2E_TEST_TIMEOUT=8h E2E_TEST=TestOperator/CreateMultiReaper
```

I will send you the IPs privately as I don't want to expose them on GitHub.

Makefile
adutra: Indeed, there is room for improvement. For now, you can find documentation for the tags [here](https://github.com/adutra/k8ssandra-operator/blob/cloud-e2e/test/e2e/suite_test.go#L82-L128).

test/framework/fixtures.go
adutra: Hmm what do you mean?

test/framework/fixtures.tmpl
adutra: I think we can indeed have a few examples of kustomizations for each cloud provider. But to use those examples in our e2e tests, those kustomizations would have to contain the actual external IPs to use, and the actual zone names (racks). This would be undesirable IMO because 1) the external IPs are kind of sensitive and I'd avoid to expose them, and 2) this info could change over time if we re-create the test cluster with different IPs or different zones. One possible solution would be to investigate if kustomize can accept environment variables for these bits, which would allow us to avoid hard-coding them in the kustomize file.

adutra:
For the record, my goal here is to provide 2 ways to run the e2e tests on GKE:

* Manual with a make target (done);
* Through GitHub Actions, using the [get-gke-credentials](https://github.com/google-github-actions/get-gke-credentials) action (still in progress).

test/framework/fixtures.tmpl
```diff
@@ -0,0 +1,184 @@
...
+      - op: replace
+        path: /spec/cassandra/networking/hostNetwork
+        value: {{ $.HostNetwork }}
+  {{ end }}
+
+  {{if $.Zones }}
+    {{ range $rackIndex, $rack := .Cassandra.Racks }}
+      {{ if index $rack.NodeAffinityLabels "topology.kubernetes.io/zone" }}
+      - op: replace
+        path: /spec/cassandra/racks/{{ $rackIndex }}/nodeAffinityLabels/topology.kubernetes.io~1zone
```
jsanda, line 34: Is that a type in `topology.kubernetes.io~1zone` with the `~1`?

test/framework/fixtures.tmpl
jsanda: >the external IPs are kind of sensitive and I'd avoid to expose them

Agreed. They could be provide to GHA via secrets.

>this info could change over time if we re-create the test cluster with different IPs or different zones

I'm not too concerned about the zone labels. We don't have to use the zone labels. We can use whatever labels we want. We can configured labels on worker nodes with whatever automation we use to configure the k8s clusters.

Let's find some time to sync up on this. I would at least like to have some idea how things would look with a more declarative approach vs the templating-heavy approach.

test/framework/fixtures.tmpl
adutra: Heh. That's the way of specifying a field containing a slash. According to http://jsonpatch.com/#json-pointer :

> If you need to refer to a key with ~ or / in its name, you must escape the characters with `~0` and `~1` respectively. For example, to get "baz" from `{ "foo/bar~": "baz" }` you’d use the pointer `/foo~1bar~0`.

adutra:
> I'm not too concerned about the zone labels. We don't have to use the zone labels. We can use whatever labels we want. We can configured labels on worker nodes with whatever automation we use to configure the k8s clusters.

Oh indeed that's an awesome idea that would simplify a lot the way we configure the e2e tests to run on a cloud provider. I haven't thought of it I confess.

> Let's find some time to sync up on this. I would at least like to have some idea how things would look with a more declarative approach vs the templating-heavy approach.

Agreed 👍

adutra:
Ok I'm considering this ready-ish. 

There are plenty of opportunities to improve what we currently have but I think we could tackle them separately.

* Revisit the heavy templating to adapt the fixtures to cloud environments as per @jsanda suggestion
* Don't use zone names for topology, use fixed labels instead (increased portability) as per @jsanda suggestion
* Run e2e tests on GHA: I am not going to propose a solution here because it's going to be involved and includes at least the following:
  * Use GKE specific actions for credentials: https://github.com/google-github-actions/get-gke-credentials 
  * Need to create service accounts on GKE beforehand
  * Make ingress routes namespace-aware (allow parallelization of tests)

test/framework/fixtures.go
```diff
@@ -1,12 +1,15 @@
 package framework
 
 import (
-	"errors"
+	"fmt"
```
burmanm, line 4: What generates this type of imports? Usually the go imports are grouped in 3 from my experience, the golang core libraries, external libraries and project's internal libraries. This import structure seems to mess up everything.

test/framework/fixtures.tmpl
jsanda: Neat :)

test/framework/fixtures.go
adutra: Import order has always been a mess, each IDE is using a different style.

adutra:
I can confirm that the remaining failures are flakies. I'm going to merge this now and create the follow-up tickets.

PR #469, opened 2022-03-17
Java garbage collection properties cannot be configured correctly

### LINKED ISSUE #465: K8SSAND-1325 ⁃ Java garbage collection properties cannot be configured correctly
**What happened?**
I created a Cassandra 4.x cluster with some GC properties set in my manifest, but they were not configured correctly in the Cassandra pod. This is closely related to https://github.com/k8ssandra/k8ssandra/issues/1353 in that it has to do with how [cass]-config-definitions] handles JVM GC properties for Cassandra 4.x. 

First I tried to configure some G1 settings:

```yaml
kind: K8ssandraCluster
metadata:
  name: test
spec:
  auth: false
  cassandra:
    serverVersion: "4.0.3"
    storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
    config:
      jvmOptions:
        heapSize: 512Mi
        additionalOptions:
          - "-XX:MaxGCPauseMillis=600"
          - "-XX:ParallelGCThreads=4"
    networking:
      hostNetwork: true
    datacenters:
      - metadata:
           name: dc1
         size: 1
```

The `additionalOptions` property basically appends properties to the list of properties passed to Cassandra's JVM. `-XX:MaxGCPauseMillis=500` is already passed, so `-XX:MaxGCPauseMillis=600` is appended to the list. It isn't clear which one will take precedence, but there should only be one instance of `-XX:MaxGCPauseMillis` in the properties passed to the JVM. 

I didn't see `-XX:ParallelGCThreads` previously set so that was ok.

I tried configuring CMS with the following settings:

```yaml
    config:
      jvmOptions:
        heapSize: 512Mi
        additionalOptions:
          - "-XX:+UseConcMarkSweepGC"
          - "-XX:SurvivorRatio=4"
```

Again those properties are simply appended to the list of properties passed to the JVM. `-XX:+UseG1GC` appears earlier in the list, so presumably it is used. 

I don't think we can rely on using `additionalOptions` for configuring GC properties. 

I expect to be able to configure GC properties.

Deploy the above manifest. When the Cassandra pod is ready grab Cassandra's logs with:

`kubectl logs test-dc1-default-sts-0 -c server-system-logger > cassandra.log`

Look for the log statement that starts with `INFO  [main] 2022-03-13 22:35:40,855 CassandraDaemon.java:642 - JVM Arguments`.

**Environment**

* K8ssandra Operator version:

    `1.0.1`

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1325) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1325
┆priority: Medium

adutra:
**What this PR does**:

This PR refactors how JVM options are passed, and enables all available garbage collectors to be used.

Summary of changes:

#### New struct tag cass-config

* A new struct tag `cass-config` has been created to allow a more fine-grained control about how `CassandraYaml` and `JvmOptions` are serialized.

The contents of the tag obey to the following syntax:

```
tag := <path1>[,<path2>,...,<pathN>][;<option1>;<option2>;...;<optionN>]
path := <constraint>:[<segments>]
constraint := a valid semver constraint
segments := a series of path segments separated by a slash, e.g. foo/bar/qix
option := recurse | retainzero
```

Semantic versions and constraints are coming from the `github.com/Masterminds/semver/v3` library.

Here is an example:

```go
RingDelayMs *int cass-config:"^3.11.x:jvm-options/cassandra_ring_delay_ms,>=4.x:jvm-server-options/cassandra_ring_delay_ms"`
```

The above means:
1. if the current server version is in the range 3.11.0 inclusive to 4.0.0 exclusive, then print the field value at the `jvm-options/cassandra_ring_delay_ms` path;
2. If the current server version is equal to or greater than 4.0.0, then print the field value at the `jvm-server-options/cassandra_ring_delay_ms` path.
3. Otherwise, ignore the field value.

There are two other options: `recurse` and `retainzero`.

1. The option `recurse` tells the tag parser to recursively walk the field's contents, instead of appending its value atomically.
2. The option `retainzero` tells the tag parser to output the field value even if it is a zero value for the field type. By default, zero values are not included in the output.

Examples:

```go
        // The CassandraYaml struct will be recursively walked and all its values will be inserted at the current output path (that is, they will be merged to the current values).
	CassandraYaml CassandraYaml cass-config:"*:cassandra-yaml;recurse"`

        // The AuditLogOptions struct will be recursively walked and values will be inserted inside the audit_logging_options path in the output
	AuditLoggingOptions *AuditLogOptions cass-config:">=4.x:audit_logging_options;recurse"`

        // The field value will be inserted in the output even if it is false (false is the zero value of bool).
	Enabled bool cass-config:"*:enabled;retainzero"`

        // The field value will be included only if the version is 3.11.x, in which case it will be included even if it is false.
	StartRpc bool `cass-config:"^3.11.x:cassandra-yaml/start_rpc;retainzero"`
```

#### JVM Options revamp

Existing fields `heapSize` and `heapNewGenSize` were deprecated in favor of new ones.

CRD JVM option fields were given normalized names that start with a prefix that indicates what they are used for:

```
additionalOptions: []string
cassandra_available_processors: *int
cassandra_config_directory: *string
cassandra_disable_auth_caches_remote_configuration: *bool
cassandra_expiration_date_overflow_policy: *string
cassandra_force_3_0_protocol_version: *bool
cassandra_force_default_indexing_page_size: *bool
cassandra_initial_token: *string
cassandra_join_ring: *bool
cassandra_load_ring_state: *bool
cassandra_max_hint_ttl_seconds: *int
cassandra_metrics_reporter_config_file: *string
cassandra_replace_address: *string
cassandra_replay_list: *string
cassandra_ring_delay_ms: *int
cassandra_triggers_dir: *string
cassandra_write_survey: *bool
debug_disable_contended_annotations: *bool
debug_enable_flight_recorder: *bool
debug_listen_remote_debuggers: *bool
debug_log_compilation: *bool
debug_non_safepoints: *bool
debug_preserve_frame_pointer: *bool
debug_unlock_commercial_features: *bool
debug_unlock_diagnostic_vm_options: *bool
gc: string
gc_cms_heap_size_young_generation: *resource.Quantity
gc_cms_initiating_occupancy_fraction: *int
gc_cms_max_tenuring_threshold: *int
gc_cms_survivor_ratio: *int
gc_cms_wait_duration_ms: *int
gc_g1_conc_threads: *int
gc_g1_initiating_heap_occupancy_percent: *int
gc_g1_max_gc_pause_ms: *int
gc_g1_parallel_threads: *int
gc_g1_rset_updating_pause_time_percent: *int
gc_print_application_stopped_time: *bool
gc_print_date_stamps: *bool
gc_print_details: *bool
gc_print_flss_statistics: *bool
gc_print_heap: *bool
gc_print_log_file_size: *resource.Quantity
gc_print_number_of_log_files: *int
gc_print_promotion_failure: *bool
gc_print_tenuring_distribution: *bool
gc_print_use_log_file: *bool
gc_print_use_log_file_rotation: *bool
heap_initial_size: *resource.Quantity
heap_max_size: *resource.Quantity
jdk_allow_attach_self: *bool
jmx_connection_type: *string
jmx_port: *int
jmx_remote_ssl: *bool
jmx_remote_ssl_opts: *string
jmx_remote_ssl_require_client_auth: *bool
netty_eventloop_maxpendingtasks: *int
netty_try_reflection_set_accessible: *bool
nio_align_direct_memory: *bool
nio_maxcachedbuffersize: *resource.Quantity
vm_always_pre_touch: *bool
vm_crash_on_out_of_memory_error: *bool
vm_disable_biased_locking: *bool
vm_disable_perf_shared_mem: *bool
vm_enable_assertions: *bool
vm_enable_non_root_thread_priority: *bool
vm_enable_thread_priorities: *bool
vm_exit_on_out_of_memory_error: *bool
vm_guaranteed_safepoint_interval_ms: *int
vm_heap_dump_on_out_of_memory_error: *bool
vm_per_thread_stack_size: *resource.Quantity
vm_prefer_ipv4: *bool
vm_print_heap_histogram_on_out_of_memory_error: *bool
vm_resize_tlab: *bool
vm_string_table_size: *resource.Quantity
vm_use_numa: *bool
vm_use_tlab: *bool
```

#### Use `resource.Quantity` to express sizes

`resource.Quantity` can now be used anywhere we expect sizes. The value will be serialized as an unscaled integer (which usually corresponds to the size in bytes) so that the JVM gets the right amount.

Note that in `CassandraYaml` we have a few fields that would have been better expressed by `resource.Quantity`. I will open an issue to fix this.

#### Ability to select supported GCs and custom ones

The `gc` field can have the following values:

* G1GC: always supported
* CMS: supported in 3.11
* ZGC: supported in 4.0
* Shenandoah: supported in 4.0 but currently unusable because the mgmt-api does not include this GC.
* Graal: supported in 4.0 but currently unusable because the mgmt-api does not include this GC.
* Custom: selecting this value implies that the user is going to configure GC manually using `additionalOptions`.
* 
There is an issue to add support for Shenandoah: #490.
There are e2e tests included in this PR for all combinations of GCs and C* versions that are currently viable.

Fixes #465 

**Checklist**
- [x] Changes manually tested


pkg/encryption/encryption_stores.go
```diff
@@ -20,55 +20,55 @@ type Stores struct {
...
-	EncryptionSettings `json:",inline"`
+	EncryptionSettings `json:",inline" cass-config:"*:;recurse"`
 }
 
 // See the cassandra.yaml file for explanations on how to set these options: https://github.com/apache/cassandra/blob/cassandra-4.0/conf/cassandra.yaml#L1091-L1183
 // +kubebuilder:object:generate=true
 type EncryptionSettings struct {
 	// +kubebuilder:validation:optional
-	Protocol string `json:"protocol,omitempty"`
+	Protocol string `json:"protocol,omitempty" cass-config:"*:protocol"`
```
adejanovski, line 55: isn't the path redundant with the json tag in case of an inline setting?
Have you considered using the json tag for the setting output name, and only specifying a path in the cass-config tag?
Not sure how much you can handle one tag using info from another, but seeing the repetition of the name in both tags feels like something could be optimized.

test/kubectl/kubectl.go
```diff
@@ -243,7 +243,7 @@ func Get(opts Options, args ...string) (string, error) {
 
 	err := cmd.Run()
 
-	if logOutput {
+	if logOutput || err != nil {
```
adejanovski, line 246: 👍

test/testdata/fixtures/gc/3.11-jdk11-CMS/k8ssandra.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: k8ssandra.io/v1alpha1
```
adejanovski, line 1: that fixture name is incorrect (unless it's on purpose and there's something I'm missing).
3.11 can only work with jdk8, which is what's shipping in the images we publish and use.
Shouldn't it be `3.11-jdk8-CMS` instead?

test/testdata/fixtures/gc/3.11-jdk11-G1/k8ssandra.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: k8ssandra.io/v1alpha1
```
adejanovski, line 1: Same here, it should be `3.11-jdk8-G1`

controllers/k8ssandra/datacenters.go
```diff
@@ -66,7 +66,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 		// unauthenticated clusters.
 		cassandra.ApplySystemReplication(dcConfig, *systemReplication)
 
-		if !cassandra.IsCassandra3(dcConfig.ServerVersion) && kc.HasStargates() {
+		if dcConfig.ServerVersion.Major() != 3 && kc.HasStargates() {
```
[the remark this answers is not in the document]
adutra: I felt that the `IsCassandra3` function was now a bit useless since it's so easy to call `.Major()`. Do you prefer that I put it back?

pkg/cassandra/config_tag.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// parseCassConfigTag parses a textual representation of a cass-config tag.
+// The textual representation is of the following general form:
+// tag := <path1>[,<path2>,...,<pathN>][;<option1>;<option2>;...;<optionN>]
+// path := <constraint>:[<segments>]
+// constraint := a valid semver constraint
+// segments := a series of path segments separated by a slash, e.g. foo/bar/qix
+// option := recurse | retainzero
+func parseCassConfigTag(text string) (*cassConfigTag, error) {
```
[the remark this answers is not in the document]
adutra: I think it will be rather complex to use regex here, especially to parse version constraints. The full syntax for version contraints is explained [here](https://github.com/Masterminds/semver#basic-comparisons). If we really want something more robust, I'd rather advocate for a grammar engine like ANTLR.

pkg/encryption/encryption_stores.go
adutra: We could do that, but in fact I like decoupling the two tags, even if that leads to some repetition.

With the changes in this PR:

* The `json` tag is used only for the CRD;
* The `cass-config` tag is used only for config-builder payloads.

This allows you to specify different values: one for how the field appear in the CRD and another for how it is passed to config-builder. I did it for the JVM options for example, and thus the CRD JVM options look more "homogenous" since I harmonized the option names.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -26,1027 +26,1608 @@ import (
...
+	JvmOptions JvmOptions `json:"jvmOptions,omitempty" cass-config:"*:;recurse"`
 }
 
 // CassandraYaml defines the contents of the cassandra.yaml file. For more info see:
 // https://cassandra.apache.org/doc/latest/cassandra/configuration/cass_yaml_file.html
 type CassandraYaml struct {
 	// Exists in 3.11, 4.0, trunk
 	// +optional
-	AllocateTokensForKeyspace *string `json:"allocate_tokens_for_keyspace,omitempty"`
+	AllocateTokensForKeyspace *string `json:"allocate_tokens_for_keyspace,omitempty" cass-config:"*:allocate_tokens_for_keyspace"`
```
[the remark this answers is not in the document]
adejanovski: 👍

pkg/encryption/encryption_stores.go
adejanovski: I understand the flexibility it gives, but do we really need different names for the CRD and the config-builder options?
I went back and checked, and it seems like it lead you to specialize some settings for G1 although they are used by all collectors, such as `ParallelGCThreads` or `ConcGCThreads` (see [here](https://docs.oracle.com/en/java/javase/11/gctuning/z-garbage-collector1.html) for example, but it applies to CMS as well).

I'd be keen to retain the original jvm option name throughout the chain, otherwise folks will have to look for a mapping before setting them. We would make it a smoother experience if the names were just converted to snake case, don't you agree?

pkg/cassandra/config_tag.go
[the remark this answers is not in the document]
adejanovski: Sounds good 👍

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
adejanovski: I'm good, we can keep it your way.

pkg/encryption/encryption_stores.go
adutra: > I'd be keen to retain the original jvm option name throughout the chain [...] We would make it a smoother experience if the names were just converted to snake case

Well, if you prefer... but the names will be a mess :-)
@jsanda what would be your preference?

> it seems like it lead you to specialize some settings for G1 although they are used by all collectors, such as ParallelGCThreads or ConcGCThreads

Unfortunately those settings will only work for G1. This is because config-builder defines a conditional on the GC name for those settings to be taken into account, see [here](https://github.com/datastax/cass-config-definitions/blob/f3f5e40c780100990729abc1c62498c520c26815/resources/jvm11-server-options/cassandra/jvm11-server-options-cassandra-4.0.0.edn#L79-L90) for C* 4.0 and [here](https://github.com/datastax/cass-config-definitions/blob/7e019d117fd0324b43c072c19648276d09de9f9e/resources/jvm-options/cassandra/jvm-options-cassandra-3.11.6.edn#L222-L251) for C* 3.11. That's why I prefixed them with "gc_g1_".

adejanovski: > Unfortunately those settings will only work for G1. This is because config-builder defines a conditional on the GC name for those settings to be taken into account, see [here](https://github.com/datastax/cass-config-definitions/blob/f3f5e40c780100990729abc1c62498c520c26815/resources/jvm11-server-options/cassandra/jvm11-server-options-cassandra-4.0.0.edn#L79-L90) for C* 4.0 and [here](https://github.com/datastax/cass-config-definitions/blob/7e019d117fd0324b43c072c19648276d09de9f9e/resources/jvm-options/cassandra/jvm-options-cassandra-3.11.6.edn#L222-L251) for C* 3.11. That's why I prefixed them with "gc_g1_".

😱

adutra: I know... config-builder sucks sometimes. The only way to pass those options when using another gc is to use `additionalOptions`.

adutra: @jsanda would you mind chiming in here if you have time?

I fear that the alignment of option names as exposed to the K8ssandra user with actual option names as recognized by both config-builder and Cassandra is a pipedream. There are already cases where the option name changes with the Cassandra version (e.g. biased locking). And with the changes from CASSANDRA-15234, the config options for C* 4.1 will change even more drastically.

adejanovski: Given what we know will happen in 4.1, you can discard my comment. Having different names for both our input and output will probably be more helpful than I thought.

PR #472, opened 2022-03-21
Stabilize e2e

### LINKED ISSUE #457: K8SSAND-1314 ⁃ Experiment on CircleCI to run the e2e tests
GHA is proving to be too resource constrained to run the e2e tests under good conditions.
We need to experiment with CircleCI to see if larger instances can help us improve test execution.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1314) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1314
┆priority: Medium

adejanovski:
**What this PR does**:
Applies a set of fixes and tunings to stabilize e2e tests execution in CI:

- Removal of resource limits on memory which were incorrectly set leading to OOMKills of Cassandra pods
- Simplification of the AuthOnOff e2e test, removing the last part were auth is turned off again (that's overkill and makes the execution longer and more error prone in that environment. We can recreate another on -> off e2e test if we really hold on to this scenario).
- Addition of retries surrounding some Stargate calls along with larger timeout settings for cassandra queries
- Addition of retries in the Traefik deployment (it fails from time to time due to internet shenanigans)
- Multi DC tests with 3 nodes were downgraded from Cassandra 4.0 to 3.11 (which could be related to [this PR](https://github.com/k8ssandra/k8ssandra-operator/pull/469) and may be reverted once we merge it).*
- Creation of a CircleCI workflow that we can use in the future to more expensive tests on "large" instances.

Fixes #457

**Checklist**
- [x] Changes manually tested


test/e2e/suite_test.go
```diff
@@ -386,10 +388,13 @@ func beforeTest(t *testing.T, f *framework.E2eFramework, opts *e2eTestOpts) erro
 	}
 
 	if opts.deployTraefik {
-		if err := f.DeployTraefik(t, opts.operatorNamespace); err != nil {
-			t.Logf("failed to deploy Traefik")
-			return err
-		}
+		retry.Do(func() error {
```
burmanm, line 391: What's the benefit over assert.Eventually or whatever we otherwise use in the tests? Seems bad to introduce yet another framework for doing the same thing, only increases maintenance burden.

test/e2e/suite_test.go
```diff
@@ -864,8 +870,8 @@ func addDcToCluster(t *testing.T, ctx context.Context, namespace string, f *fram
 
 	t.Log("check that nodes in dc1 see nodes in dc2")
 	pod := "test-dc1-default-sts-0"
-	count := 4
-	checkNodeToolStatus(t, f, f.K8sContext(0), namespace, pod, count, 0, "-u", username, "-pw", password)
+	count := dcSize * 2
+	checkNodeToolStatus(t, f, "kind-k8ssandra-0", namespace, pod, count, 0, "-u", username, "-pw", password)
```
burmanm, line 874: Should not hardcode kind-k8ssandra-0 here.

test/testdata/fixtures/add-dc/k8ssandra.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: test
 spec:
   cassandra:
-    serverVersion: "4.0.1"
+    serverVersion: "3.11.11"
```
burmanm, line 7: Why downgrading from 4.0.1 to 3.11.11 series?

test/testdata/fixtures/add-dc/k8ssandra.yaml
```diff
@@ -29,28 +29,30 @@ spec:
...
+        range_request_timeout_in_ms: 60000
+        read_request_timeout_in_ms: 60000
+        request_timeout_in_ms: 60000
+        truncate_request_timeout_in_ms: 60000
+        write_request_timeout_in_ms: 60000
       jvmOptions:
         heapSize: 384Mi
     networking:
       hostNetwork: true
-    resources:
```
burmanm, line 36: If there's no resources limits at all, then this Cassandra could eat everything. It shouldn't, why isn't 1GB enough in these tests?

test/e2e/suite_test.go
adejanovski: yes, good point. That part got modified by another PR after I pushed this one I think.

test/testdata/fixtures/multi-dc-stargate/k8ssandra.yaml
```diff
@@ -1,3 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cassandra-config
+data:
+  cassandra.yaml: |
+    concurrent_reads: 32
+    concurrent_writes: 32
```
burmanm, line 8: Probably bad for performance to have so many readers and writers in a restricted CPU core count.

test/testdata/fixtures/add-dc/k8ssandra.yaml
adejanovski: 3.11.11 provided better stability. There were too many moving parts to work with and this was a simple fix.
We should open a follow up ticket to profile the footprint of C* 4.x vs 3.11 and come up with other adjustments to make 4.0 stable in multi cluster e2e tests.

test/testdata/fixtures/add-dc/k8ssandra.yaml
adejanovski: I think it's not really a concern for e2e tests. If Cassandra eats more than it should and gets killed, the test fails. Same if it eats everything and doesn't get killed, it fails.
Limits were incorrectly set in many tests and I don't think they provide value for e2e tests anyway.
Also, there's no way Cassandra can eat more than twice the heap size, plus MCAC and MAAC heap sizes*2. So there's an absolute limit, it's just not set at the k8s level.

test/testdata/fixtures/multi-dc-auth/k8ssandra.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   auth: false
   cassandra:
-    serverVersion: "4.0.1"
+    serverVersion: "3.11.11"
```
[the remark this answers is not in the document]
adejanovski: same as before, it seems like 4.0 shows a heavier footprint in our tests and switching some flaky tests to 3.11 helped stabilize the executions.
I'm not satisfied with it as we'll quickly need to run both versions at the same time using a matrix, but it can be considered outside of the scope of this work (single cluster e2e tests are still running on 4.0 FTR).

test/e2e/suite_test.go
adejanovski: Using `Eventually` was my first idea, but I seem to recall wouldn't work with retries on Stargate, as there are some places where we need to get something in return of the call.
Let me check back all the places where I'm using `retry.Do` to see if it's legit or if `Eventually` can be used everywhere.
I agree on the principle that introducing a new dependency should only be done if the ones we have don't provide the functionality we're after.

test/testdata/fixtures/multi-dc-stargate/k8ssandra.yaml
adejanovski: I don't think it matters much because we're not doing anything that's heavy on performance, but I'll lower them.

test/e2e/suite_test.go
adejanovski: After checking again, you're totally right. `require.Eventually` will do the job. I'll switch to this and remove the `retry` dependency.

adejanovski:
@burmanm, I've rebased to fix the conflicts and removed the `retry` dependency.
Let me know if you can approve the PR.

test/testdata/fixtures/add-dc/k8ssandra.yaml
burmanm: If it can't eat, then there should be nothing wrong with setting it a limit, right? That limit is never reached.

test/framework/framework.go
```diff
@@ -256,7 +256,9 @@ func (f *Framework) CreateNamespace(name string) error {
 		}
 		f.logger.Info("creating namespace", "Namespace", name, "Context", k8sContext)
 		if err := remoteClient.Create(context.Background(), namespace); err != nil {
-			return err
+			if !errors.IsAlreadyExists(err) {
```
burmanm, line 259: Are we sure we want this? If we don't anticipate it to exists, perhaps we should exit? There could be some old resources in that case that could cause issues.

test/testdata/fixtures/add-dc/k8ssandra.yaml
burmanm: Odd. I've assumed they lowered the resource overhead in 4.x series. Something to investigate later as we really need to test against 4.x

test/testdata/fixtures/add-dc/k8ssandra.yaml
adejanovski: What's wrong with setting a limit is if it's incorrectly set, which was the case for various test scenarios.
Setting the limit correctly is error prone in this case because of the various settings that can influence Cassandra's pod memory usage (Cassandra heap+off heap, Management API heap+off heap, MCAC collectd), and if heap sizes are changed, the limits must be changed accordingly.
Since this work aims at improving the test pass rate, we're better off without limits.

Anyone can work on a follow up ticket to appropriately set those limits and run tens of executions to verify they are set appropriately, but IMO removing them altogether serves the primary goal here.

test/testdata/fixtures/add-dc/k8ssandra.yaml
adejanovski: agreed 👍

adejanovski: it could be due to the fact that CMS isn't available in 4.0 and G1 is the default there. In 3.11, CMS is the default.

burmanm (approved):

PR #482, opened 2022-03-28
Telemetry webhook validation

### LINKED ISSUE #235: K8SSAND-1084 ⁃ Validate telemetry configurations + provide safety
When the user has requested ServiceMonitor deployment via the k8ssandra CRs, but Prometheus operator (and therefore the `ServiceMonitor` CRs) has not been installed, the requested resources should not be created and the user should be informed of the error.

Ideally, the validation should be implemented in a validating webhook.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1084) by [Unito](https://www.unito.io)
┆epic: Prometheus integration
┆friendlyId: K8SSAND-1084
┆priority: Medium

Miles-Garnsey:
**What this PR does**:

Validates telemetry specifications within the K8ssandraCluster, having regard for the installation status of Prometheus across each K8s DC in the cluster and the telemetry requested for it.

This is preferred to the current situation where validation occurs during reconciliation for two reasons:
1. It ensures that users get fast feedback on the failure via the resource being rejected, instead of them needing to consult logs to determine why service monitors are not created.
2. It ensures that invalid resources do not pass through into the cluster. When this happens, additional updates to the resource cannot propagate, which can cause stability risks if a broken resource is not noticed and then subsequent changes need to be made.

Fixes #235

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
I'm trying to figure out what is causing the tests to fail on this one. It isn't really straight forward, especially considering some tests are apparently still flaky. I'm uncertain if I can progress this without #472 being merged.

One source of new flakes appears to be the envtest `StopDatacenter`, which fails due to the following error at line 184:

```
                	            	Operation cannot be fulfilled on k8ssandraclusters.k8ssandra.io "stop-dc-test": the object has been modified; please apply your changes to the latest version and try again
```

This definitely should not be happening, as this is a transient failure due to optimistic locking and should never be fatal. It is occurring because the Patch has no timing tolerance. Wrapping in an eventually now to see if that improves things.

Miles-Garnsey:
I've wrapped the patch operation in an `Eventually`:

```
patch := client.MergeFromWithOptions(kc.DeepCopy(), client.MergeFromWithOptimisticLock{})
	kc.Spec.Cassandra.Datacenters[0].Stopped = true
	require.Eventually(t, func() bool {
		err := f.Client.Patch(ctx, kc, patch)
		return err == nil
	}, timeout, interval, "timeout waiting to patch dc1 while stopping")
```

The StopDatacenter test still fails on GHA, and now fails locally (BONUS!) This is going well.

jsanda:
You shouldn't need to wrap writes in Eventually calls unless you are doing so to add retry logic. It's not needed for consistency though.

Miles-Garnsey:
> You shouldn't need to wrap writes in Eventually calls unless you are doing so to add retry logic. It's not needed for consistency though.

Yes, but I need to add retry logic because of the optimistic locking error above `Operation cannot be fulfilled on k8ssandraclusters.k8ssandra.io "stop-dc-test": the object has been modified; please apply your changes to the latest version and try again`

burmanm:
Writing old object 100 times will not make a difference, it's still the old object. You need to refresh it (Get)

Miles-Garnsey:
> Writing old object 100 times will not make a difference, it's still the old object. You need to refresh it (Get)

Oh wow, how did I miss that. Thanks @burmanm you just saved me a fair bit of time honestly.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -34,11 +35,11 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
 	}
 	logger.Info("merged TelemetrySpec constructed", "mergedSpec", mergedSpec, "cluster", kc.Name)
 	// Confirm telemetry config is valid (e.g. Prometheus is installed if it is requested.)
-	promInstalled, err := telemetry.IsPromInstalled(remoteClient, logger)
+	promInstalled, err := validation.IsPromInstalled(remoteClient, logger)
 	if err != nil {
 		return result.Error(err)
 	}
-	validConfig, err := telemetry.SpecIsValid(mergedSpec, promInstalled)
+	validConfig, err := validation.TelemetrySpecIsValid(mergedSpec, promInstalled)
```
jeffbanks, line 42: Is there a chance that the `mergedSpec` goes unchecked before this validation?   That is, if the `mergedSpec` is nil does this negate the entire reconcile operation.

Miles-Garnsey: If mergedSpec is nil we probably don't need to validate whether Prometheus is installed.

Miles-Garnsey: In saying that, this logic may need another round of thought on my part to confirm that every possible telemetry specification is checked. There is a lot of complexity to do with nil pointer dereferencing given the number of different ways things can be specified between the cluster and (optional) DC levels.

jeffbanks: In this case, I think it would be fine to just check the mergedSpec before the Prometheus validation.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -85,14 +89,100 @@ func (r *K8ssandraCluster) validateK8ssandraCluster() error {
...
+
+type clientGetter interface {
+	GetRemoteClient(k8sContextName string) (client.Client, error)
+}
+
+func TelemetrySpecsAreValid(kCluster *K8ssandraCluster, clientCache clientGetter) error {
+	// Iterate through the DCs, checking Stargate and Cassandra telemetry
+	for _, dc := range kCluster.Spec.Cassandra.Datacenters {
+		//Get client and determine if prom installed.
+		dcClient, err := clientCache.GetRemoteClient(dc.K8sContext)
```
jeffbanks, line 106: This seems a bit confusing to have `clientCache` referenced here as a `clientGetter` interface type, while `clientCache` is referenced from the import using the same name.

Miles-Garnsey: You're right, it is shadowing which is poor practice. I'll fix this.

Miles-Garnsey:
I'm think this PR is ready for preliminary review. 

I've fixed 5 instances where tests were fragile due to being timing sensitive. 

There are 7 tests still failing. All fail due to timing related issues, all failures are in the multi-cluster suite. 

I will move on to making calls to check Prometheus installation status concurrent so that timeouts become less frequent.

test/webhooks/k8ssandracluster_webhook_test.go
```diff
@@ -235,6 +243,120 @@ func testNumTokens(t *testing.T) {
 	require.Error(err)
 }
 
+func testTelemetryValidation(t *testing.T) {
```
burmanm, line 246: This should be ported to envtest as this test was envtest and there's nothing in this test that would not work in a envtest. At the moment this test does not verify anything, since it's not targeting Kubernetes API.

pkg/validation/telemetry_validation.go
```diff
@@ -1,4 +1,4 @@
-package telemetry
+package validation
```
burmanm, line 1: Why the package change?

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -85,14 +89,107 @@ func (r *K8ssandraCluster) validateK8ssandraCluster() error {
...
 		}
 	}
+	if err := TelemetrySpecsAreValid(r, clientCache); err != nil {
+		return err
+	}
+	return nil
+}
+
+// clientGetter is an interface here purely to enable DI. It is anything that implements `GetRemoteClient()` mechanism to get remote clients by name.
+type clientGetter interface {
```
burmanm, line 99: This is unnecessary, we already have ``clientCache.GetRemoteClient(dc.K8sContext)`` available here and it's available in the envtest also. There isn't need for another interface that hides the implementation.

pkg/validation/telemetry_validation.go
jeffbanks: I'll let @Miles-Garnsey describe the intent here.  I can see where having our validations all in one `validation` package could be handy for generalizing and refactoring in the future.

test/webhooks/k8ssandracluster_webhook_test.go
Miles-Garnsey: As discussed, I do not believe an envtest is required for this change and do not intend to write one.

pkg/validation/telemetry_validation.go
Miles-Garnsey: The package change is required to avoid an import cycle due to the webhooks being located in `apis`.

pkg/test/fakeclient.go
```diff
@@ -52,7 +62,32 @@ func (rm fakeRESTMapper) KindsFor(resource schema.GroupVersionResource) ([]schem
 		},
 	}, nil
 }
-func NewFakeClientWRestMapper() client.Client {
-	fakeClient, _ := NewFakeClient()
-	return composedClient{fakeClient}
+
+// MockClientCache is a mock to retrieve remote clients.
```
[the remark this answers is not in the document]
Miles-Garnsey: I think I answered this above.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
Miles-Garnsey: This required to sub in a mock implementation. If this is confusing please feel free to discuss with me offline.

controllers/stargate/stargate_telemetry_reconciler.go
```diff
@@ -35,11 +36,11 @@ func (r *StargateReconciler) reconcileStargateTelemetry(
 		cfg.CommonLabels[k8ssandraapi.K8ssandraClusterNameLabel] = klusterName
 	}
 	// Confirm telemetry config is valid (e.g. Prometheus is installed if it is requested.)
-	promInstalled, err := telemetry.IsPromInstalled(remoteClient, logger)
+	promInstalled, err := validation.IsPromInstalled(remoteClient, logger)
```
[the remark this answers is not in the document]
Miles-Garnsey: I don't agree with your assumption (here and in the code more generally) that the presence of a function in a validation package implies it must only ever return error types. In fact, I think that your use of error return types to signal a failed validation is questionable as it obfuscates the difference between a failure to validate an input and an invalid input.

test/webhooks/k8ssandracluster_webhook_test.go
burmanm: Then this is considered as non-tested code and should not be merged.

Miles-Garnsey: Considering there is a test which explicitly says `testTelemetryValidation` your comment above is clearly untrue @burmanm, you should consider withdrawing it as it makes you look silly.

adejanovski:
This is largely outdated and we never managed to reach a consensus so I'll go ahead and close it.
We can rediscuss this if we think we have a need and a path forward.

Issue #495, opened 2022-03-30
Add MedusaBackupRepository type and controller


jsanda:
**What is missing?**
I created #492 which discusses making the remote storage buckets the source of truth for backups. This means we need to be able to query remote storage before and after a K8ssandraCluster exists. Currently the connection properties for the storage bucket are defined inline within the K8ssandraCluster. We need to be able to declare the connection settings separately from the K8ssandraCluster. I propose adding a new custom resource for this:

```go
type MedusaBackupRepository struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   MedusaBackupRepositorySpec   `json:"spec,omitempty"`
	Status MedusaBackupRepositoryStatus `json:"status,omitempty"`
}

type MedusaBackupRepositorySpec struct {
         // This is the same Storage struct defined in medusa_types.go
	Storage `json:",inline"`

        K8sContexts []string `json:"k8sContexts,omitempty" 
}

type MedusaBackupRepositoryStatus struct {
	// TODO
}
```
This is the same as what is currently defined inline in the K8ssandraCluster. The one difference is the `K8sContexts` property.

We will refactor medusa_reconciler.go so that it can be used by the new controller since it already does most of what is needed.

We need to add a new field in the MedusaClusterTemplate struct:

```go
type MedusaClusterTemplate struct {
	// MedusaContainerImage is the image characteristics to use for Medusa containers. Leave nil
	// to use a default image.
	// +optional
	ContainerImage *images.Image `json:"containerImage,omitempty"`

	// SecurityContext applied to the Medusa containers.
	// +optional
	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`

	// Defines the username and password that Medusa will use to authenticate CQL connections to Cassandra clusters.
	// These credentials will be automatically turned into CQL roles by cass-operator when bootstrapping the datacenter,
	// then passed to the Medusa instances, so that it can authenticate against nodes in the datacenter using CQL.
	// The secret must be in the same namespace as Cassandra and must contain two keys: "username" and "password".
	// +optional
	CassandraUserSecretRef corev1.LocalObjectReference `json:"cassandraUserSecretRef,omitempty"`

	// Provides all storage backend related properties for backups.
	StorageProperties Storage `json:"storageProperties,omitempty"`
	
[...]

PR #502, opened 2022-04-05
Create a standalone Medusa deployment

### LINKED ISSUE #483: K8SSAND-1354 ⁃ Deploy a Medusa grpc pod for operator direct interactions
In order to generate restore mappings in the operator directly, we'll need it to be able to communicate with the storage backend without having Cassandra pods up.
An additional Medusa pod then needs to be deployed for that need.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1354) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1354
┆priority: Medium

adejanovski:
Allows the operator to communicate with the Medusa storage backend without requiring a Cassandra pod to be up.

**What this PR does**:
Creates a standalone Medusa deployment with the associated service.
Allows the operator to communicate with the Medusa storage backend without requiring a Cassandra pod to be up. This will be used to compute the restore mapping in the operator but also probably to handle synchronization of custom resources with backups in remote storage.

Fixes #483

**Checklist**
- [x] Changes manually tested


burmanm:
This seems to deploy the medusa to have a standalone instance always. Why do we need it always? Wouldn't we need it only when we intend to communicate with it?

burmanm:
Well, I guess for synchronization it might make sense.

jsanda:
Yes, it would be needed to sync with remote storage. I suppose we could consider (as an optimization) not deploying it if there are Cassandra pods with Medusa enabled. Not sure I am convinced that the pros would outweigh the cons.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -137,3 +148,64 @@ func (r *K8ssandraClusterReconciler) reconcileMedusaConfigMap(
 	logger.Info("Medusa ConfigMap successfully reconciled")
 	return result.Continue()
 }
+
+func (r *K8ssandraClusterReconciler) reconcileMedusaStandalone(
```
jsanda, line 152: This doesn't make sense, at least not here in the K8ssandraClusterReconciler. The motivation of having a stand alone Medusa deployment is that the remote storage can be queried when there is no K8ssandraCluster. 

The work for this should happen as part of #495 which I have started but not gotten too far yet.

adejanovski: That's not exactly what I had understood. The motivation was to be able to communicate with the Medusa storage backend even if no Cassandra pods were up yet, in order to perform a restore on the very first start of these pods.
A K8ssandraCluster object is still our starting point IMO. It's where we'll define the name of the backup we want to restore, as well as the medusa backend settings.

Miles-Garnsey:
Made these two review comments offline, putting them here so they aren't lost.

1. Can you please rebase? It looks like this branch is 20 commits behind and there is a conflict in `reconcile.go` which makes me think a review right now may not be too useful as the rebase will lead to quite a few changes.
2. Because we're deploying a reasonably substantial new resource (a new pod is a relatively big change IMO), can we please get some e2e or kuttl tests?

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
 	}
 }
 
+func medusaMainContainer(dcConfig *cassandra.DatacenterConfig, logger logr.Logger) (int, bool, *corev1.Container) {
```
Miles-Garnsey, line 192: This function returns a lot of values, can we please use named returns to make it clearer + more self documenting?

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
...
+	medusaContainerIndex, found := cassandra.FindContainer(dcConfig.PodTemplateSpec, "medusa")
+	medusaContainer := &corev1.Container{Name: "medusa"}
+	if found {
+		logger.Info("Found medusa container")
```
Miles-Garnsey, line 196: This logging feels excessive.

test/e2e/medusa_test.go
```diff
@@ -134,6 +136,19 @@ func checkMedusaContainersExist(t *testing.T, ctx context.Context, namespace str
 	require.True(found, fmt.Sprintf("%s doesn't have medusa container", dc1.Name))
 }
 
+func checkMedusaStandalonePodExists(t *testing.T, ctx context.Context, namespace string, dcKey framework.ClusterKey, f *framework.E2eFramework, kc *api.K8ssandraCluster) {
```
Miles-Garnsey, line 139: Doesn't `ClusterKey` contain `namespace`? Do we need to pass them separately?

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -93,6 +94,26 @@ func createMultiDcClusterWithMedusa(t *testing.T, ctx context.Context, f *framew
 	dc1Key := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}, K8sContext: f.DataPlaneContexts[0]}
 	require.Eventually(f.DatacenterExists(ctx, dc1Key), timeout, interval)
 
+	t.Log("check that the standalone Medusa deployment was created")
```
Miles-Garnsey, line 97: It might be good to check that the pod becomes ready here as well as just being created.

test/e2e/medusa_test.go
Miles-Garnsey: Also, can we put this in `pkg/test`? `suite_test.go` is too long and doesn't have much coherence.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -137,3 +148,64 @@ func (r *K8ssandraClusterReconciler) reconcileMedusaConfigMap(
...
+	logger logr.Logger,
+) result.ReconcileResult {
+
+	logger.Info("Reconciling Medusa standalone deployment on namespace : " + desiredDeployment.ObjectMeta.Namespace)
+	if kc.Spec.Medusa != nil {
+		deploymentKey := utils.GetKey(desiredDeployment)
+
+		logger := logger.WithValues("MedusaDeployment", deploymentKey)
+
+		// Compute a hash which will allow to compare desired and actual configMaps
```
Miles-Garnsey, line 167: actual and desired _deployments_?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -137,3 +148,64 @@ func (r *K8ssandraClusterReconciler) reconcileMedusaConfigMap(
...
+			desiredDeployment.DeepCopyInto(actualDeployment)
+			actualDeployment.SetResourceVersion(resourceVersion)
+			if err := remoteClient.Update(ctx, actualDeployment); err != nil {
+				logger.Error(err, "Failed to update Medusa deployment resource")
+				return result.Error(err)
+			}
+			return result.RequeueSoon(r.DefaultDelay)
+		}
+
+		// Create a service for the medusa standalone deployment
```
Miles-Garnsey, line 194: This logic does not use the `CompareHashAnnotations` logic typically used in reconciliation. I'm open to not using it because `CompareHashAnnotations` doesn't actually allow for real self healing (the hash annotation is not updated when a user manually modifies an object, so an object can drift without being detected).

Just wondering why we're using two different methodologies really, I don't think it is necessarily a problem.

Miles-Garnsey (changes requested):
High level feedback here:

* I think some refactoring is needed. It looks like there is similar logic for (at least) volume reconciliation operating on both the standalone deployment and the pod within the CassDC's STS. It would be good to see this logic unified in a single function. 
*  `medusaMainContainer` is another place I'd also like to see some more clarity. There are a few functions (including `medusaMainContainer`) where either the returns or the parameters might benefit from some attention.
* Packaging - it might be good to split functions in `pkg/medusa/reconcile` into multiple files. `reconcile.go` isn't very descriptive.
* Tests: I think unit tests might be desirable for fns like `medusaMainContainer`, `StandaloneMedusaDeployment`, `StandaloneMedusaService`.`UpdateMedusaVolumes`. E2E tests should probably check that the pod becomes ready rather than just being created.

I think once the function signatures are straightened out and you've considered a refactoring to unify the logic for configuring the standalone and in-DC Medusa `Containers`, we'll be in good shape.

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
...
+	}
+	return medusaContainerIndex, found, medusaContainer
+}
+
+// Creates a deployment for a standalone Medusa pod which will be used by the operator to interact directly with the storage backend.
+// It allows such interactions before any Cassandra pod is created, to enable performing a restore on the first startup.
+func StandaloneMedusaDeployment(dcConfig *cassandra.DatacenterConfig, medusaSpec *api.MedusaClusterTemplate, clusterName string, volumes []medusaVolume, additionalVolumes []medusaAdditionalVolume, namespace string, logger logr.Logger) *appsv1.Deployment {
+	_, _, medusaContainer := medusaMainContainer(dcConfig, logger)
+
+	medusaDeployment := &appsv1.Deployment{
```
Miles-Garnsey, line 208: I might be wrong here since it isn't quite clear where medusaContainer comes from (in the worst case it seems to be an empty newly allocated Container from medusaMainContainer). So apologies if this is a silly question...

Do we have liveness and readiness checks for this container?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -52,10 +53,20 @@ func (r *K8ssandraClusterReconciler) ReconcileMedusa(
...
+		for _, volume := range volumes {
+			cassandra.AddOrUpdateVolume(dcConfig, volume.Volume, volume.VolumeIndex, volume.Exists)
+		}
+		for _, volume := range additionalVolumes {
+			cassandra.AddOrUpdateAdditionalVolume(dcConfig, volume.Volume, volume.VolumeIndex, volume.Exists)
+		}
 		cassandra.AddCqlUser(medusaSpec.CassandraUserSecretRef, dcConfig, medusa.CassandraUserSecretName(medusaSpec, kc.Name))
+		medusaStandalone := medusa.StandaloneMedusaDeployment(dcConfig, medusaSpec, kc.Name, volumes, additionalVolumes, namespace, logger)
+		medusaService := medusa.StandaloneMedusaService(dcConfig, medusaSpec, kc.Name, namespace, logger)
+		r.reconcileMedusaStandalone(ctx, kc, remoteClient, medusaStandalone, medusaService, logger)
```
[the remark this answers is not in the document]
jsanda: I haven't looked at the PR yet, but we should only need a volume for the Medusa configmap.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: In the end I decided not to use the hash for the service. There's really nothing to configure there, so once it's created then we're fine. I'm reluctant to add more code to handle this, as it's not really necessary.

controllers/k8ssandra/medusa_reconciler.go
adejanovski: fixed

pkg/medusa/reconcile.go
```diff
@@ -172,6 +171,14 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
 
 	medusaContainer.VolumeMounts = medusaVolumeMounts(medusaSpec, dcConfig, logger)
 
+	medusaContainer.Ports = []corev1.ContainerPort{
```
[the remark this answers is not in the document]
adejanovski: Let me know how it all looks now.

pkg/medusa/reconcile.go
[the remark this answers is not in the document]
adejanovski: it does :) I removed it

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
...
 
+func medusaMainContainer(dcConfig *cassandra.DatacenterConfig, logger logr.Logger) (int, bool, *corev1.Container) {
+	medusaContainerIndex, found := cassandra.FindContainer(dcConfig.PodTemplateSpec, "medusa")
+	medusaContainer := &corev1.Container{Name: "medusa"}
+	if found {
+		logger.Info("Found medusa container")
+
+		medusaContainer = dcConfig.PodTemplateSpec.Spec.Containers[medusaContainerIndex].DeepCopy()
+	}
+	return medusaContainerIndex, found, medusaContainer
```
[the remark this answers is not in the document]
adejanovski: That whole logic got refactored. Let me know how it looks.

controllers/k8ssandra/medusa_reconciler.go
Miles-Garnsey: Did we determine that changes to labels etc. would not cause an update of the service? As long as changes flowing from the CRs do not cause uncaught updates to the service I'm fine with not using a hash comparison.

Miles-Garnsey (changes requested):
You're going to love me I'm afraid here @adejanovski, I have a few more comments... 

Firstly, this is looking much better and 90% of the changes have fixed the majority of the areas I'd queried. 

I think we still need to attend to the following areas:

1. Liveness/readiness, I've suggested three options there, but I think we should define a default (raise an issue if you want this in a separate PR, I'm open to that.)
2. We need to have resource requests/limits on the Medusa container, you can make this configurable like your readiness/liveness checks if you want.
3. `UpdateMedusaVolumes` is still a bit odd to me, can we try to make the behaviour consistent between additionalvolumes/volumes?
4. One question RE hashannotation comparisons. 

Otherwise, thanks for your efforts, we're getting very close.

pkg/medusa/reconcile.go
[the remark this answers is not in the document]
Miles-Garnsey: Very much better! Resolving.

pkg/medusa/reconcile.go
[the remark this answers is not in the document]
Miles-Garnsey: Resolving subsequent to refactoring.

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
...
+				},
+			},
+		},
+	}
+
+	for _, volume := range volumes {
+		medusaDeployment.Spec.Template.Spec.Volumes = append(medusaDeployment.Spec.Template.Spec.Volumes, *volume.Volume)
+	}
+	for _, additionalVolume := range additionalVolumes {
+		// Additional volumes here are used to store backup data, which this deployment will not require.
```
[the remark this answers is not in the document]
Miles-Garnsey: Understood that this is required so we can use the same Medusa config as used in the Cass STS pods. Suggest further review at a later time but this is fine for a temporary solution.

pkg/medusa/reconcile.go
```diff
@@ -182,6 +189,100 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
...
+		medusaDeployment.Spec.Template.Spec.Volumes = append(medusaDeployment.Spec.Template.Spec.Volumes, corev1.Volume{
+			Name: additionalVolume.Volume.Name,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		})
+	}
+
+	// Create dummy additional volumes
```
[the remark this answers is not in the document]
Miles-Garnsey: Understood that this is required so we can use the same Medusa config as used in the Cass STS pods. Suggest further review at a later time but this is fine for a temporary solution.

pkg/medusa/reconcile.go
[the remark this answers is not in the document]
Miles-Garnsey: Much better!

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -143,4 +143,12 @@ type MedusaClusterTemplate struct {
 
 	// Provides all storage backend related properties for backups.
 	StorageProperties Storage `json:"storageProperties,omitempty"`
+
+	// Define the readiness probe to use for the Medusa containers.
+	// +optional
+	ReadinessProbe *corev1.Probe `json:"readinessProbe,omitempty"`
```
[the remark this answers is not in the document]
adejanovski: There are default probes defined [here](https://github.com/adejanovski/k8ssandra-operator/blob/add-medusa-controller-pod/pkg/medusa/reconcile.go#L206-L242), and merged with whatever the user sets in config.
I guess I should only give the ability to override the timeout and threshold settings. I'll come up with something that matches what we do elsewhere in the code.
Good point about the resource limits, I'll add the ability to set them and set a default.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -52,10 +54,23 @@ func (r *K8ssandraClusterReconciler) ReconcileMedusa(
...
+
+		// Generate the Medusa main container
+		medusaMainContainer := medusa.GenerateMedusaMainContainer(dcConfig, medusaSpec, logger)
+		// Add the containers definitions in the podTemplateSpec
 		medusa.UpdateMedusaInitContainer(dcConfig, medusaSpec, logger)
-		medusa.UpdateMedusaMainContainer(dcConfig, medusaSpec, logger)
-		medusa.UpdateMedusaVolumes(dcConfig, medusaSpec, logger)
+		medusa.UpdateMedusaMainContainer(dcConfig, medusaMainContainer, logger)
+		// Create required volumes for Medusa
+		additionalVolumes := medusa.UpdateMedusaVolumes(dcConfig.PodTemplateSpec, medusaSpec, dcConfig.Cluster, logger)
```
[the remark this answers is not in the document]
adejanovski: > Can we get this all consistent by doing everything (both volumes + additionalvolumes) in-place to start with. That'll be a good first step.

As much as I would want this, it's not possible without going against your other requirement of working on the pod template spec. The additional volumes aren't defined in the `podTemplateSpec`, they're under `dcConfig.StorageProperties`.

I'll rename `UpdateMedusaVolumes()` to `GenerateMedusaVolumes()` and have it return both volumes and additional volumes, then add them to the pods in an external loop. At least that behavior will be consistent for both types of volumes.

pkg/medusa/reconcile.go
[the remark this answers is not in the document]
adejanovski: 👍

controllers/k8ssandra/medusa_reconciler.go
adejanovski: > As long as changes flowing from the CRs do not cause uncaught updates to the service I'm fine with not using a hash comparison.

The K8ssandraCluster name is all that could trigger a breaking change here and it's not possible, so we're safe 👍

adejanovski:
@Miles-Garnsey, tests are all passing now, let me know if we're good to go.

test/e2e/medusa_test.go
Miles-Garnsey: This isn't even in `suite_test.go`. Ignore me.

pkg/medusa/reconcile.go
```diff
@@ -187,6 +237,116 @@ func UpdateMedusaMainContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
 	}
 }
 
+func generateMedusaProbe(configuredProbeSettings *api.ProbeSettings) *corev1.Probe {
```
Miles-Garnsey, line 240: Could/should this be done via Goalesce? Note that I can't see `TerminationGracePeriodSeconds` in here, which may be an important configurable in some cases.

apis/medusa/v1alpha1/medusa_types.go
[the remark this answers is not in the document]
Miles-Garnsey: Carrying forward comments into new review now that some of this has been addressed.

controllers/k8ssandra/medusa_reconciler.go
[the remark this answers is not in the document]
Miles-Garnsey: Agreed. I don't like using both volumes and additionalvolumes but this will do for now, pending my investigation of what additionalVolumes is doing under the hood.

apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -149,4 +149,43 @@ type MedusaClusterTemplate struct {
 	// See https://docs.datastax.com/en/developer/python-driver/latest/security/ for more information on the required files.
 	// +optional
 	CertificatesSecretRef corev1.LocalObjectReference `json:"certificatesSecretRef,omitempty"`
+
+	// Define the readiness probe settings to use for the Medusa containers.
+	// +optional
+	ReadinessProbeSettings *ProbeSettings `json:"readinessProbeSettings,omitempty"`
```
Miles-Garnsey, line 155: Can we drop `Settings` off the end of these names. This is a total nit, but I think that e.g. `ReadinessProbe` is sufficient.

Miles-Garnsey (commented):
I've made a few changes but we are truly down to brass tacks here and I am negotiable on all of these. 

I do feel that the tests could be better separated between integration and unit tests, and I'd like to think carefully about defining `ProbeSettings` given the similarities to `corev1. Probe`.

The other comments are for your consideration only.

pkg/medusa/reconcile.go
```diff
@@ -9,18 +9,30 @@ import (
...
 	"github.com/go-logr/logr"
 	cassandra "github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 const (
-	DefaultMedusaImageRepository = "k8ssandra"
+	DefaultMedusaImageRepository = "adejanovski"
```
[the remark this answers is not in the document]
adejanovski: You're right, I need to cut out a Medusa release first.

apis/medusa/v1alpha1/medusa_types.go
adejanovski: Since we're not defining the whole probe (I'm leaving out the Exec part on purpose because it shouldn't be overridden), I think it's fair to leave "settings" at the end.

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -93,6 +121,44 @@ func createMultiDcClusterWithMedusa(t *testing.T, ctx context.Context, f *framew
...
+	medusaDeploymentKey1 := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: medusa.MedusaStandalonePodName("test", "dc1")}, K8sContext: f.DataPlaneContexts[0]}
+	medusaDeployment1 := &appsv1.Deployment{}
+	require.Eventually(func() bool {
+		if err := f.Get(ctx, medusaDeploymentKey1, medusaDeployment1); err != nil {
+			return false
+		}
+		return true
+	}, timeout, interval)
+
+	require.Equal(int32(kc.Spec.Medusa.ReadinessProbeSettings.FailureThreshold), medusaDeployment1.Spec.Template.Spec.Containers[0].ReadinessProbe.FailureThreshold)
```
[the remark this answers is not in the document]
adejanovski: Honestly, as long as it's covered I think I'm happy with the way it is. I spent too much time already on this PR 😓

Miles-Garnsey:
@adejanovski, as discussed, it does appear that the sanitized clusterName field is the point of reference for downstream resource names, so that's all good. Apologies for my error there.

I've raised a ticket RE using goalesce to disentangle things a bit here, but as it is an implementation detail I won't block merge on this point.

I tested once again, changing the k8s resource name, cluster name, and rack names (but leaving the DC name unchanged). 

That tests shows that I can use cqlsh to extract data added to the original cluster from the restored cluster. So I think this is good to merge. Approving once again. Nice work on getting this monster across the line!

Miles-Garnsey (approved):

adejanovski:
Closing in favor of #980

PR #523, opened 2022-04-13
Consider Traefik as a prerequisite for e2e tests

### LINKED ISSUE #515: K8SSAND-1449 ⁃ Consider Traefik as a prerequisite for e2e tests
Traefik is currently deployed as part of the `beforeTest` steps for each e2e test that requires ingresses.

Several things are wrong with this approach:

* Multiple installs: one single install of Traefik should be enough for all tests. 
* Also, Treafik should be considered a prerequisite for tests and be installed only once, prior to the e2e test execution. This should save us some time when running multiple e2e tests locally.
* The tests themselves don't need to know that Traefik is being used: the only thing they need is a URL reachable from outside the cluster. Traefik should be an implementation detail and should be replaceable by any other ingress framework.
* Sticking to Traefik won't play well if/when our tests will be run against cloud providers that favor other ingress providers, see #466.

I suggest the following:

* Consider the deployment of an Ingress controller a pre-requisite
* Create make targets to install Traefik locally with Helm, call them when setting up a test cluster. 
  * Traefik should be configured to watch all namespaces.
  * For local clusters, the current [Helm values file](https://github.com/k8ssandra/k8ssandra-operator/blob/1dc67e484a224c0d208aae27207b98c83562a883/test/testdata/ingress/traefik.values.yaml) can be reused. 
* Remove all the code in e2e framework to deploy Traefik programmatically.
* We still need to create the ingresses as part of the e2e test, since the namespace and service names are specific to the test. For that:
  * Create new test flags: `stargateIngressUrl` and `reaperIngressUrl` with the following defaults for local clusters: `stargate.127.0.0.1.nip.io` and `reaper.127.0.0.1.nip.io`.
  * Then the e2e framework will generate 2 ingress objects based on these templates:
    * [Stargate](https://github.com/k8ssandra/k8ssandra-operator/blob/1dc67e484a224c0d208aae27207b98c83562a883/test/testdata/ingress/stargate-ingress.yaml)
    * [Reaper](https://github.com/k8ssandra/k8ssandra-operator/blob/bd8af171793933bea3f11c848d090433fb0642c2/test/testdata/ingress/reaper-ingress.yaml)
  * The namespace and service name will be kustomized to match the e2e test current namespace + service name (as it is done already); the host field will be kustomized to include the flag values (new).
  * The templates above are specific to Traefik; we will need to use another ingress type. The most generic option is to use Kubernetes Ingress, but it doesn't handle TCP very well. We will need to investigate how to proceed with testing the CQL port, but one option would be to disable CQL port testing through a flag, if the underlying Ingress controller doesn't handle it.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1449) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1449
┆priority: Medium

adutra:
**What this PR does**:

This PR abstracts away Traefik from the e2e framework.

Fixes #515 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/test/e2e/reaper_test.go
+++ b/test/e2e/reaper_test.go
@@ -38,7 +38,8 @@ func createSingleReaper(t *testing.T, ctx context.Context, namespace string, f *
 	testRemoveReaperFromK8ssandraCluster(t, f, ctx, kcKey, dcKey, reaperKey)
 
 	t.Log("deploying Reaper ingress routes in", f.DataPlaneContexts[0])
-	f.DeployReaperIngresses(t, ctx, f.DataPlaneContexts[0], namespace, "test-dc1-reaper-service")
+	reaperRestHostAndPort := ingressConfigs[f.DataPlaneContexts[0]].ReaperRest
+	f.DeployReaperIngresses(t, ctx, f.DataPlaneContexts[0], namespace, "test-dc1-reaper-service", reaperRestHostAndPort)
 	defer f.UndeployAllIngresses(t, f.DataPlaneContexts[0], namespace)
 
 	t.Run("TestReaperApi[0]", func(t *testing.T) {
@@ -62,7 +63,8 @@ func createSingleReaperWithEncryption(t *testing.T, ctx context.Context, namespa
 	checkReaperK8cStatusReady(t, f, ctx, kcKey, dcKey)
 
 	t.Log("deploying Reaper ingress routes in context", f.DataPlaneContexts[0])
-	f.DeployReaperIngresses(t, ctx, f.DataPlaneContexts[0], namespace, "test-dc1-reaper-service")
+	reaperRestHostAndPort := ingressConfigs[f.DataPlaneContexts[0]].ReaperRest
+	f.DeployReaperIngresses(t, ctx, f.DataPlaneContexts[0], namespace, "test-dc1-reaper-service", reaperRestHostAndPort)
 	defer f.UndeployAllIngresses(t, f.DataPlaneContexts[0], namespace)
 
 	t.Run("TestReaperApi[0]", func(t *testing.T) {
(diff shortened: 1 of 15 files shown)


test/testdata/ingress/stargate-ingress.yaml
```diff
@@ -2,40 +2,40 @@
...
-  namespace: default # actual namespace will be kustomized
+  name: cluster1-dc1-stargate-service-http-ingress
+  namespace: k8ssandra
 spec:
   entryPoints:
-    - web-http
+    - web
   routes:
-    - match: Host(`stargate.127.0.0.1.nip.io`) && PathPrefix(`/v1/auth`)
+    - match: Host(`stargate.example.com`) && PathPrefix(`/v1/auth`)
```
adutra, line 11: The changes in the Ingress yaml definitions are merely to make these files more illustrative. John asked me a while ago to always consider the yaml fixture files as self-documenting examples, so I chose more generic names for hosts, namespaces, service names, etc. The actual names don't matter, they will be kustomized before getting applied.

test/e2e/reaper_test.go
```diff
@@ -326,7 +338,7 @@ func testRemoveReaperFromK8ssandraCluster(
 
 func connectReaperApi(t *testing.T, ctx context.Context, k8sContext, clusterName, username, password string) reaperclient.Client {
 	t.Logf("Testing Reaper API in context %v...", k8sContext)
-	var reaperURL, _ = url.Parse(fmt.Sprintf("http://%s", ingresses[k8sContext]["reaper_rest"]))
+	var reaperURL, _ = url.Parse(fmt.Sprintf("http://%s", ingressConfigs[k8sContext].ReaperRest))
```
jeffbanks, line 341: 👍🏻 on the `ReaperRest`, `StargateRest`, etc.  approach to plug these in per context.

jeffbanks (approved):
LGTM.  Minor indent and one +1.

PR #548, opened 2022-05-15
add overlay with image tag override for cluster scoped test

### LINKED ISSUE #547: K8SSAND-1513 ⁃ Cluster scoped e2e tests need image tag override for release branches
## Problem
In e2e tests we deploy k8ssandra-operator with kustomize using the `config/deployments` kustomizations. We use the `latest` tag in tests. The tag is configured in `config/deployments/default`.

On release branches it is a bit tricky. Part of the release process involves updating the image tag in `config/deployments/default` to the version being released, e.g., v1.0.0. This won't for tests running on release branches work since if that image hasn't been published yet. To work around this there is a kustomization overlay that is generated in the e2e test framework that overrides the image tag so that it is always set to `latest`. The tests on the release branch will use the locally built image just like is done in PRs. 

This is not being done for the cluster-scoped e2e test which means it will always fail on the release branch, at least until the image version is available on Docker Hub. 

## Solution
There needs to be an overlay for the cluster-scoped tests like there is for other tests.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1513) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1513
┆priority: Medium

jsanda:
**What this PR does**:

Fixes #547
**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.0.md
+++ b/CHANGELOG/CHANGELOG-1.0.md
@@ -28,6 +28,7 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 * [TESTING] [#514](https://github.com/k8ssandra/k8ssandra-operator/issues/514) ⁃ Consider cert-manager a prerequisite for e2e tests
 * [TESTING] [#515](https://github.com/k8ssandra/k8ssandra-operator/issues/515) ⁃ Consider ingress controller a prerequisite for e2e tests
 * [TESTING] [#463](https://github.com/k8ssandra/k8ssandra-operator/issues/463) ⁃ Make fixtures cloud-ready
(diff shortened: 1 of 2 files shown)


burmanm:
Wouldn't it be simpler to just load the generated image to kind before testing? Then the v1.0.0 for example would be available.

jsanda:
>Wouldn't it be simpler to just load the generated image to kind before testing?

We would need to update the GHA workflow to determine the version. That can't be done just by looking at the branch name since we can release multiple versions from the branch. Alternatively, I suppose we could parse config/deployments/default/kustomization.yaml to determine the version. This should be pretty easy with `yq` which in turn keeps the test code simpler overall. I will submit a different PR with that approach.

PR #554, opened 2022-05-30
Medusa client to node encryption support

### LINKED ISSUE #354: K8SSAND-1215 ⁃ Add encryption support to Medusa
**What is missing?**
When encryption is turned on, Cassandra, Stargate and Reaper will all get configured appropriately but Medusa won't.
We need to configure Medusa appropriately to support client to node encryption.

**Why do we need it?**
Encrypted clusters still need to be backed up ;)

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1215) by [Unito](https://www.unito.io)
┆fixVersions: k8ssandra-operator-v1.1.0
┆friendlyId: K8SSAND-1215
┆priority: Medium

adejanovski:
**What this PR does**:
Enable Medusa support for client to node encryption when it's turned on in Cassandra. 
The way python handles SSL requires to provide the root ca, and a client certificate/key pair, which will be provided through a secret.
The encryption readme was updated with the steps to generate these files.

Fixes #354

**Checklist**
- [x] Changes manually tested


apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -143,4 +143,10 @@ type MedusaClusterTemplate struct {
 
 	// Provides all storage backend related properties for backups.
 	StorageProperties Storage `json:"storageProperties,omitempty"`
+
+	// Certificates for Medusa if client encryption is enabled in Cassandra.
+	// The secret must be in the same namespace as Cassandra and must contain three keys: "rootca.crt", "client.crt_signed" and "client.key".
+	// See https://docs.datastax.com/en/developer/python-driver/3.20/security/ for more information on the required files.
```
adutra, line 149: 3.25 is the latest version fyi. Or are you sticking to the exact version used by Medusa?

test/e2e/suite_test.go
```diff
@@ -226,13 +226,13 @@ func TestOperator(t *testing.T) {
 	}))
 	t.Run("CreateSingleMedusaJob", e2eTest(ctx, &e2eTestOpts{
```
adutra, line 227: Probably unrelated, but why some of the medusa tests are called "Old" and others are called "Job"?

adutra (approved):

apis/medusa/v1alpha1/medusa_types.go
adejanovski: not specifically. I'll put `latest` instead of a specific version, that should do it.

test/e2e/suite_test.go
adejanovski: In 1.1, we introduced a new set of CRDs named MedusaBackup, MedusaBackupJob and MedusaRestoreJob with the associated controllers. The old APIs are deprecated but still usable, hence the additional tests. They were renamed with the `Old` suffix because invoking the e2e tests with just  `TestOperator/CreateSingleMedusa` would also trigger `CreateSingleMedusaJob` (that's the way subtests work I guess, it just sees what you enter as a prefix).

PR #562, opened 2022-06-09
Cleanup medusa controllers, fix a bug here or there..

burmanm:
**What this PR does**:
Multiple things:

* Add BackupStatus call to the MedusaClient
* Remove MedusaBackupReconciler as it did not do anything, move the RBAC right annotations to MedusaBackupJobController
* MedusaTaskScheduler did not update the FinishTime correctly due to incorrectly placed Patch creation
* Make all Medusa*Controllers ignore Reconcile requests caused by the Reconciler doing a Status update (no changes to the actual reconciled object)

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


controllers/medusa/cassandrabackup_controller.go
```diff
@@ -215,6 +217,6 @@ func backupFinished(backup *medusaapi.CassandraBackup) bool {
 
 func (r *CassandraBackupReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&medusaapi.CassandraBackup{}).
+		For(&medusaapi.CassandraBackup{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
adejanovski, line 220: Does that prevent from reconciling on status changes?

burmanm: Yep, it filters all the reconcile requests that did not change the real object (as the generation stays the same if we update the status).

PR #588, opened 2022-06-27
Add the ability to set metrics filters in MCAC

### LINKED ISSUE #573: K8SSAND-1579 ⁃ Add metric filters support to exclude unwanted metrics.
**What is missing?**

The k8ssandra 1.x allows to configure [metric_filters](https://github.com/k8ssandra/k8ssandra/blob/main/charts/k8ssandra/values.yaml#L428) by which we can exclude unwanted metrics, This feature is net yet there the k8ssandra-operator which is a limiting factor as it causes issues as explained below. 

**Why do we need it?**

We recently noticed a weird thing after enabling Prometheus metrics in the k8ssandra cluster managed by the operator, The `/metrics` endpoint (`e.g http://<cassandra-pod-ip>:9103/metrics`) is way too big, around 31 MB on a 9 node cluster! :scream:

```
curl -LO http://10.244.45.116:9103/metrics                                       
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0 31.0M    0 18844    0     0  17502      0  0:30:59  0:00:01  0:30:58 17627
```

This is abnormal and it causes excessive network traffic on the minion that runs Prometheus leading to congestion and performance issues, we had to disable Prometheus monitoring for our k8ssandra clusters due to this :slightly_frowning_face: 

**Environment:**

Kubernetes v1.21.4 (self hosted)

**K8ssandra Operator version:**

k8ssandra-operator v1.1.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1579) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1579
┆priority: Medium

adejanovski:
**What this PR does**:
MCAC supports metrics filters to reduce the amount of metrics scraped by Prometheus. This PR allows to define those in the telemetry part of the K8ssandraCluster spec and sets the same default filters than K8ssandra v1.x.
There's also a small refactoring to have a single function used to search for env variables in a container throughout the code.

Fixes #573 

**Checklist**
- [x] Changes manually tested


pkg/telemetry/cassandra_metrics_filters.go
```diff
@@ -0,0 +1,50 @@
...
+// If no filters are defined and telemetry was set in the CRD, the default filters are used.
+func InjectCassandraTelemetryFilters(telemetrySpec *telemetry.TelemetrySpec, dcConfig *cassandra.DatacenterConfig) {
+	if telemetrySpec == nil || telemetrySpec.Prometheus == nil {
+		return
+	}
+
+	metricsFilters := make([]string, len(telemetrySpec.Prometheus.McacMetricFilters))
+	containerIndex, containerFound := cassandra.FindContainer(dcConfig.PodTemplateSpec, "cassandra")
+
+	if len(telemetrySpec.Prometheus.McacMetricFilters) > 0 {
```
Miles-Garnsey, line 38: Should we make it possible to NOT filter metrics by passing an empty `McacMetricFilters` slice? 

`McacMetricFilters` would need to become a slice pointer to accomodate a true nil value.

Miles-Garnsey (commented):
I want to run a manual test of this to confirm that everything passes through as expected.

A few areas in there that I think you may also want to change, in particular - 

1. Allow passing an empty slice to avoid any filtering.
2. We may want to think about whether filtering should happen by default even if telemetry is disabled.

controllers/k8ssandra/datacenters.go
```diff
@@ -77,6 +78,10 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 		if medusaResult := r.ReconcileMedusa(ctx, dcConfig, dcTemplate, kc, logger); medusaResult.Completed() {
 			return medusaResult, actualDcs
 		}
+		// Inject MCAC metrics filters if telemetry is enabled
+		if kc.Spec.Cassandra.Telemetry != nil {
```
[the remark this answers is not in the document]
Miles-Garnsey: That's an option, but remember the struct is

```
Telemetry {
  Prometheus {
    Enabled
  }
}
```
So you can already put the filters under the telemetry struct. Take a look and see what you think.

pkg/telemetry/cassandra_metrics_filters.go
Miles-Garnsey: I'd still change it to a nil pointer because that's more idiomatic for the k8s API. Nil means "do the default", empty means "empty filters". This means that the user needs less knowledge of how the filtering/MCAC config file works.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
jsanda: Why do we need the enabled flag at all? Wouldn't the presence or absence of `Prometheus` be sufficient?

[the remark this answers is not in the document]
adejanovski: Well, the `enabled` flag was there from the beginning so we probably should keep it now.
You're right @Miles-Garnsey, I should take the mcac filters out of the `prometheus` struct and just put it under `telemetry`.
Not sure I need another `MCAC` struct there though.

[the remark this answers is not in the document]
adejanovski: @jsanda, a comment in the source code explains why we need the boolean:

```
Enabled bool `json:"enabled,omitempty"` // A bool flag required here to disambiguate when e.g. the cluster should have telemetry turned on but one DC should have it explicitly turned off.
```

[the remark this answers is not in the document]
Miles-Garnsey: > Why do we need the enabled flag at all? Wouldn't the presence or absence of Prometheus be sufficient?

We went through this when the initial PR was put together. Recapping - some DCs may not have Prometheus available. In that case, it is helpful to be able to enable at cluster level and disable at DC level.

> Not sure I need another MCAC struct there though.

Your call. I'm just thinking of forward compatibility in the event that we end up supporting some other type of metrics agent. Kind of similar to the way I encapsulated Prometheus related things in `Prometheus` struct, which has now made it much easier for you to add `MCAC` related things outside that struct (but still under telemetry).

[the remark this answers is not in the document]
adejanovski: ok ok, making it as follows:

```
telemetry:
  mcac:
    metricFilters: []
```

pkg/telemetry/cassandra_metrics_filters.go
adejanovski: idiomatic we go then 👍

test/testdata/fixtures/single-dc/k8ssandra.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
 spec:
   cassandra:
     serverVersion: 3.11.11
+    telemetry:
```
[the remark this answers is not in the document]
adejanovski: Yup, removing that now that we cleared things out.

apis/telemetry/v1alpha1/telemetry_types.go
```diff
@@ -4,14 +4,26 @@ package v1alpha1
 
 type TelemetrySpec struct {
 	Prometheus *PrometheusTelemetrySpec `json:"prometheus,omitempty"`
+	Mcac       *McacTelemetrySpec       `json:"mcac,omitempty"`
```
Miles-Garnsey, line 7: Nit: `MCAC` according to https://github.com/golang/go/wiki/CodeReviewComments. (scroll up slightly, the anchor link for initialisms is broken.)

Miles-Garnsey: Also, just double checking whether `mcac` is right for the yaml, my gut feeling is that this is the right way to do it, but just drawing your attention to it, as I couldn't find a style guide for CRDs - I think they should follow whatever is preferred for JSON schemas.

pkg/telemetry/cassandra_metrics_filters.go
```diff
@@ -26,25 +26,19 @@ var (
...
-	if len(telemetrySpec.Prometheus.McacMetricFilters) > 0 {
-		// Custom filters were defined in the CRD.
-		metricsFilters = append(metricsFilters, telemetrySpec.Prometheus.McacMetricFilters...)
-	} else {
-		// No custom filters were defined in the CRD, using the defaults.
-		metricsFilters = append(metricsFilters, DefaultFilters...)
-	}
-
 	if containerFound {
+		if telemetrySpec == nil || telemetrySpec.Mcac == nil || telemetrySpec.Mcac.MetricFilters == nil {
```
Miles-Garnsey, line 34: Boom! Nice.

pkg/telemetry/cassandra_metrics_filters_test.go
```diff
@@ -83,14 +85,15 @@ func Test_InjectCassandraTelemetryFilters_Disabled(t *testing.T) {
...
+	telemetrySpec := &telemetry.TelemetrySpec{
+		Mcac: &telemetry.McacTelemetrySpec{
+			MetricFilters: &[]string{},
+		},
+	}
 	InjectCassandraTelemetryFilters(telemetrySpec, dcConfig)
 	cassandraEnvVariables := dcConfig.PodTemplateSpec.Spec.Containers[0].Env
-	assert.Equal(t, 0, len(cassandraEnvVariables), "Expected no env variable to be injected")
+	assert.Equal(t, 1, len(cassandraEnvVariables), "Expected 1 env variable to be injected")
+	assert.Equal(t, cassandraEnvVariables[0].Name, "METRIC_FILTERS", "Expected METRIC_FILTERS env variable to be injected")
```
Miles-Garnsey, line 96: Nit: This looks brittle. If the ordering of the environment variable injection changes you'll get a false positive here. Didn't you have a findEnvironmentVariable function in here somewhere that you could use to make this more robust?

Miles-Garnsey (approved):
I've only raised several nits here, your call if you want to make my suggested changes. 

I've approved anyway so you can go ahead and merge once you've decided.

pkg/telemetry/cassandra_metrics_filters_test.go
adejanovski: In this very case we should be fine since it's a unit test and no controller is involved.
In the e2e tests, I'm using a function to find the environment variable I'm looking for, because its position is unpredictable indeed.

pkg/telemetry/cassandra_metrics_filters.go
jsanda: With all due respect, I am not aware of `*[]string` being idiomatic for Kubernetes APIs. In fact I don't see much of that in Go in general. I do see `[]*Foo` some. With Kubernetes I almost always see `[]string` or `[]Foo` for example. 

If you can point me to an API conventions doc or some references/example that demonstrate otherwise that would be great.

pkg/telemetry/cassandra_metrics_filters.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"strings"
+
+	telemetry "github.com/k8ssandra/k8ssandra-operator/apis/telemetry/v1alpha1"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
+	v1 "k8s.io/api/core/v1"
+)
+
+var (
+	DefaultFilters = []string{"deny:org.apache.cassandra.metrics.Table",
```
jsanda, line 12: These should be documented, maybe both in telemetry_types.go as well as at docs.k8ssandra.io.

pkg/telemetry/cassandra_metrics_filters.go
jsanda: @Miles-Garnsey I stand corrected. I found the doc I was looking for along with another good reference:

* https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
* https://github.com/kubernetes/kubernetes/issues/34641

Ideally, I'd rather avoid the pointers, but I understand that it allows us to distinguish between default vs no filters.

PR #614, opened 2022-07-07
Fix METRIC_FILTERS env variable not being created

### LINKED ISSUE #613: K8SSAND-1633 ⁃ METRIC_FILTERS env variable isn't injected if the telemetry block is nil
**What happened?**
The PR allowing to add metric filters was merged with a bug that wouldn't inject the env variable if the telemetry block is nil.

Metric filters should always be created, even if the telemetry block isn't created in the spec.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1633) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1633
┆priority: Medium

adejanovski:
**What this PR does**:
Creates the METRIC_FILTERS env variable in all cases, even when the telemetry block isn't created in the k8c spec.
Metrics are always extracted even if no ServiceMonitor is created, so they must be filtered in all cases.

Fixes #613

**Checklist**
- [x] Changes manually tested


burmanm:
Why do we have to have the env variable?

burmanm:
And that question relates mostly to the: "If filter list is set to nil, the default filters are used, otherwise the provided filters are used."

If Telemetry is not enabled, why do we care to filter something? Does it reduce load on the collectd?

burmanm (approved):

adejanovski:
> If there's some metrics that we should always filter out on our management-api images, shouldn't we be doing that in cass-operator also?

Fair question. I guess we could move this to cass-operator indeed with a new field in the cassdc spec, and have K8ssandra-operator fill this in instead of injecting the env variable 👍 
I'll create a ticket for this.

PR #616, opened 2022-07-08
Allow injecting containers and init containers into the Cassandra pods

### LINKED ISSUE #569: Expose the podTemplateSpec for the cassdc objects
**What is missing?**
cass-operator exposes the podTemplateSpec for the statefulset pods but k8ssandra-operator doesn't allow to set it.

**Why do we need it?**
This would allow injecting containers and volumes into the Cassandra pods.

**Environment**

* K8ssandra Operator version:

    `v1.1.1`
    
    

adejanovski:
**What this PR does**:
Introduces two structs in the K8ssandraCluster CRD to allow injecting init containers and containers into the Cassandra pods.
Such containers are injected as is in first position.
A validation check was added to prevent conflicts with containers that are generated by K8ssandra-operator.

Fixes #569 

**Checklist**
- [x] Changes manually tested


pkg/cassandra/datacenter.go
```diff
@@ -348,9 +348,61 @@ func Coalesce(clusterName string, clusterTemplate *api.CassandraClusterTemplate,
 	dcConfig.ClientEncryptionStores = clusterTemplate.ClientEncryptionStores
 	dcConfig.AdditionalSeeds = clusterTemplate.AdditionalSeeds
 
+	if len(dcTemplate.DatacenterOptions.AdditionalContainers) > 0 {
```
jsanda, line 351: The `Coalesce` function previously had a single purpose - combine the cluster and DC templates with override semantics. Now the function is also modifying the PodTemplateSpec and AdditionalVolumes. It would be nice to keep the function single-purposed and do the modifications downstream.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -313,6 +313,18 @@ type DatacenterOptions struct {
...
+	// +optional
+	AdditionalContainers []corev1.Container `json:"additionalContainers,omitempty"`
+
+	// AdditionalInitContainers defines additional init containers to be deployed in each Cassandra pod.
+	// +optional
+	AdditionalInitContainers []corev1.Container `json:"additionalInitContainers,omitempty"`
+
+	// AdditionalVolumes defines additional volumes to be added to each Cassandra pod.
+	// +optional
+	AdditionalVolumes []cassdcapi.AdditionalVolumes `json:"additionalVolumes,omitempty"`
```
jsanda, line 327: I would prefer we handle volumes separately. I think it merits some more thought and discussion. AdditionalVolumes in cass-operator only supports PVs which are added to the PVC template of the StatefulSet. A volume mount is added in the cassandra container. Are we sure this is the behavior we want? Maybe it is, but I would rather take the time to discuss and be certain before we commit to the API change.

adejanovski: sounds good, let's discuss this before moving forward with that specific part.

pkg/cassandra/datacenter.go
adejanovski: The problem is that the `Coalesce` function is both:
- where the dcConfig object is initialized
- where we deal with overlaps between cluster and dc level settings

Since we need to do the latter, we end up doing the former as well.
I guess what would be right to do here is rather add `AdditionalInitContainers` and `AdditionalContainers` to the `DatacenterConfig` struct and fill them in the `Coalesce` function but only modify the `PodTemplateSpec` afterwards outside of it.
Sounds good?

adejanovski:
@jsanda, this PR is ready for review.
I've removed the additional volumes and refactored the podTemplateSpec changes out of the `Coalesce` function.
Let me know what you think.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -313,6 +313,14 @@ type DatacenterOptions struct {
...
 	// +optional
 	Telemetry *telemetryapi.TelemetrySpec `json:"telemetry,omitempty"`
+
+	// AdditionalContainers defines additional containers to be deployed in each Cassandra pod.
+	// +optional
+	AdditionalContainers []corev1.Container `json:"additionalContainers,omitempty"`
+
+	// AdditionalInitContainers defines additional init containers to be deployed in each Cassandra pod.
+	// +optional
+	AdditionalInitContainers []corev1.Container `json:"additionalInitContainers,omitempty"`
```
adejanovski, line 323: To be consistent with volumes, and since we're now allowing to specify standard k8ssandra containers (to control the ordering or add settings), should we rename these to `Containers` and `InitContainers` instead?
The presence of `Additional` feels misleading now.

jsanda: Sounds good to me

pkg/cassandra/datacenter.go
jsanda: Sounds good

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -313,6 +313,29 @@ type DatacenterOptions struct {
...
+	//  - name: server-system-logger
+	//  - name: custom-container
+	//    image: busybox
+	//  - name: cassandra
+	// Otherwise containers referenced here will be injected in first position
+	// +optional
+	Containers []corev1.Container `json:"containers,omitempty"`
+
+	// InitContainers defines init-containers to be deployed in each Cassandra pod.
+	// K8ssandra-operator and cass-operator will create their own init-containers, which can be referenced here to control ordering.
```
jsanda, line 330: How is the ordering handled with respect to init containers added by k8ssandra-operator?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2088,3 +2089,88 @@ func changeClusterNameFails(t *testing.T, ctx context.Context, f *framework.Fram
...
+
+	dc := &cassdcapi.CassandraDatacenter{}
+	err = f.Get(ctx, dc1Key, dc)
+	require.NoError(err, "failed to get CassandraDatacenter dc1")
+
+	posInit, foundInit := cassandra.FindInitContainer(dc.Spec.PodTemplateSpec, "injected-init-container")
+	require.True(foundInit, "failed to find injected-init-container")
+	require.Equal(0, posInit, "injected-init-container should be the second init container")
+
+	posMain, foundMain := cassandra.FindContainer(dc.Spec.PodTemplateSpec, "injected-container")
```
jsanda, line 2168: We don't need to worry about the position of containers, only init containers.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2088,3 +2089,88 @@ func changeClusterNameFails(t *testing.T, ctx context.Context, f *framework.Fram
 	require.NoError(err, "failed to delete K8ssandraCluster")
 	f.AssertObjectDoesNotExist(ctx, t, dc1Key, &cassdcapi.CassandraDatacenter{}, timeout, interval)
 }
+
+// Create a cluster with server and client encryption but client encryption stores missing.
+// Verify that dc1 never gets created.
+func injectContainers(t *testing.T, ctx context.Context, f *framework.Framework, namespace string) {
+	require := require.New(t)
```
jsanda, line 2096: This test should probably have Medusa enabled and then also verify positions of medusa-restore and jmx-credentials init containers in addition to user-defined ones.

pkg/cassandra/datacenter_test.go
```diff
@@ -358,6 +358,62 @@ func TestCoalesce(t *testing.T) {
 				Stopped: true,
 			},
 		},
+		{
+			name: "Additional cluster container",
```
jsanda, line 362: Can you add a test that includes init containers added by k8ssandra-operator?

test/e2e/suite_test.go
```diff
@@ -1645,3 +1646,25 @@ func checkMetricsFiltersPresence(t *testing.T, ctx context.Context, f *framework
 	}
 	return nil
 }
+
+func checkInjectedContainersPresence(t *testing.T, ctx context.Context, f *framework.E2eFramework, dcKey framework.ClusterKey) error {
```
jsanda, line 1650: It doesn't hurt to have it, but do we need these checks in e2e tests? Are the integration and unit tests sufficient?

adejanovski: integration and unit tests don't involve cass-operator. I like to make sure the new behaviors are checked in the e2e tests. It has proven useful with my last PR 😅

jsanda: >integration and unit tests don't involve cass-operator 

fair point :)

jsanda (approved):
Good to go!

PR #621, opened 2022-07-13
Add resources config settings for Medusa and Reaper containers

### LINKED ISSUE #553: K8SSAND-1524 ⁃ Add a resource configuration block for every spawned POD
**What is missing?**

Configuration for the resources for the PODs managed by the operator

**Why do we need it?**

The best practice is to set resource requirements for every POD running to allow the correct resource placement. 

> Also our operations team require every resource deployed (via the admission controller) to have such configuration otherwise it is rejected by the policy.

**Environment**

* K8ssandra Operator version: v1.1.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1524) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1524
┆priority: Medium

adejanovski:
**What this PR does**:
Allows setting resources requests and limits for the pods K8ssandra generates that are missing the option: Reaper and Medusa.
Stargate is covered already and #616 allows to set them for Cassandra pods containers that could be missing it.

Fixes #553

**Checklist**
- [x] Changes manually tested


pkg/medusa/reconcile.go
```diff
@@ -160,6 +168,22 @@ func UpdateMedusaInitContainer(dcConfig *cassandra.DatacenterConfig, medusaSpec
 	}
 }
 
+func medusaInitContainerResources(medusaSpec *api.MedusaClusterTemplate) corev1.ResourceRequirements {
+	if medusaSpec.Resources != nil {
```
burmanm, line 172: Is this intended to be ``if medusaSpec.InitContainerResources != nil`` ?

pkg/medusa/reconcile_test.go
```diff
@@ -328,3 +331,48 @@ func testMedusaIniMissingOptionalSettings(t *testing.T) {
 	assert.Contains(t, medusaIni, "secure = False")
 	assert.NotContains(t, medusaIni, "backup_grace_period_in_days =")
 }
+
+func TestInitContainerDefaultResources(t *testing.T) {
```
burmanm, line 335: There's no test to verify that the user requested resources are ever added? A simple "override works" would be fine, no need to check every detail.

adejanovski: You're right, I did it for Reaper but forgot it for Medusa.
I'll add a test for this 👍

pkg/medusa/reconcile.go
adejanovski: yep, good catch 👍

adejanovski:
@burmanm, addressed both comments. Let me know what you think.
Thanks!

burmanm (approved):

PR #626, opened 2022-07-18
Harmonize resource names using the cassdc cluster name as prefix if set

### LINKED ISSUE #619: K8SSAND-1646 ⁃ Harmonize resource names to be based on the cass dc cluster name
Cluster name overrides have been enabled which allows naming the Cassandra cluster name differently than the K8ssandraCluster name.
This leads cass-operator to create resources with the cassdc cluster name as prefix while K8ssandra-operator will create resources with the k8c name as prefix.
We'd need to harmonize resource names and use the Cassandra cluster name as prefix throughout the fleet.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1646) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1646
┆priority: Medium

adejanovski:
**What this PR does**:
Uses the Cassandra cluster name (sanitized) override if it exists to prefix the various resources created by K8ssandra-operator.

Fixes #619

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -184,7 +184,7 @@ func (in *K8ssandraCluster) SanitizedName() string {
 	if in.Spec.Cassandra.ClusterName != "" {
 		return cassdcapi.CleanupForKubernetes(in.Spec.Cassandra.ClusterName)
 	}
-	return in.ObjectMeta.Name
+	return in.Name
 }
 
 // +kubebuilder:object:root=true
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -181,7 +181,7 @@ func (in *K8ssandraCluster) GetInitializedDatacenters() []CassandraDatacenterTem
 
 // Name returns a sanitized version of the cassandra cluster name override if it exists, otherwise the k8c object name
 func (in *K8ssandraCluster) SanitizedName() string {
-	if in.Spec.Cassandra.ClusterName != "" {
+	if in.Spec.Cassandra != nil && in.Spec.Cassandra.ClusterName != "" {
 		return cassdcapi.CleanupForKubernetes(in.Spec.Cassandra.ClusterName)
 	}
 	return in.Name
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -179,7 +179,8 @@ func (in *K8ssandraCluster) GetInitializedDatacenters() []CassandraDatacenterTem
 	return datacenters
 }
 
-// Name returns a sanitized version of the cassandra cluster name override if it exists, otherwise the k8c object name
+// SanitizedName returns a sanitized version of the Cassandra cluster name override if it exists,
+// otherwise the k8c object name.
 func (in *K8ssandraCluster) SanitizedName() string {
 	if in.Spec.Cassandra != nil && in.Spec.Cassandra.ClusterName != "" {
 		return cassdcapi.CleanupForKubernetes(in.Spec.Cassandra.ClusterName)
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -179,6 +179,14 @@ func (in *K8ssandraCluster) GetInitializedDatacenters() []CassandraDatacenterTem
 	return datacenters
 }
 
+// Name returns a sanitized version of the cassandra cluster name override if it exists, otherwise the k8c object name
+func (in *K8ssandraCluster) SanitizedName() string {
+	if in.Spec.Cassandra.ClusterName != "" {
+		return cassdcapi.CleanupForKubernetes(in.Spec.Cassandra.ClusterName)
+	}
+	return in.ObjectMeta.Name
+}
+
 // +kubebuilder:object:root=true
 
 // K8ssandraClusterList contains a list of K8ssandraCluster
--- a/controllers/k8ssandra/cassandra_telemetry_reconciler.go
+++ b/controllers/k8ssandra/cassandra_telemetry_reconciler.go
@@ -30,9 +30,9 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
 		MonitoringTargetName: actualDc.Name,
 		ServiceMonitorName:   kc.SanitizedName() + "-" + actualDc.Name + "-" + "cass-servicemonitor",
 		Logger:               logger,
-		CommonLabels:         mustLabels(kc.SanitizedName(), kc.Namespace, actualDc.Name),
+		CommonLabels:         mustLabels(kc.Name, kc.Namespace, actualDc.Name),
 	}
-	logger.Info("merged TelemetrySpec constructed", "mergedSpec", mergedSpec, "cluster", kc.SanitizedName())
+	logger.Info("merged TelemetrySpec constructed", "mergedSpec", mergedSpec, "cluster", kc.Name)
 	// Confirm telemetry config is valid (e.g. Prometheus is installed if it is requested.)
 	promInstalled, err := telemetry.IsPromInstalled(remoteClient, logger)
 	if err != nil {
--- a/controllers/k8ssandra/cassandra_telemetry_reconciler.go
+++ b/controllers/k8ssandra/cassandra_telemetry_reconciler.go
@@ -28,11 +28,11 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
 	cfg := telemetry.PrometheusResourcer{
 		MonitoringTargetNS:   actualDc.Namespace,
 		MonitoringTargetName: actualDc.Name,
-		ServiceMonitorName:   kc.Name + "-" + actualDc.Name + "-" + "cass-servicemonitor",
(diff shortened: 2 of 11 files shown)


controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -28,11 +28,11 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
 	cfg := telemetry.PrometheusResourcer{
 		MonitoringTargetNS:   actualDc.Namespace,
 		MonitoringTargetName: actualDc.Name,
-		ServiceMonitorName:   kc.Name + "-" + actualDc.Name + "-" + "cass-servicemonitor",
+		ServiceMonitorName:   kc.SanitizedName() + "-" + actualDc.Name + "-" + "cass-servicemonitor",
 		Logger:               logger,
-		CommonLabels:         mustLabels(kc.Name, kc.Namespace, actualDc.Name),
+		CommonLabels:         mustLabels(kc.SanitizedName(), kc.Namespace, actualDc.Name),
```
adutra, line 33:
```suggestion
		CommonLabels:         mustLabels(kc.Name, kc.Namespace, actualDc.Name),
```

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -28,11 +28,11 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
...
 	}
-	logger.Info("merged TelemetrySpec constructed", "mergedSpec", mergedSpec, "cluster", kc.Name)
+	logger.Info("merged TelemetrySpec constructed", "mergedSpec", mergedSpec, "cluster", kc.SanitizedName())
```
adutra, line 35: Hmm not sure here. I think log messages should always refer to `kc.Name`.

controllers/k8ssandra/datacenters.go
```diff
@@ -43,10 +43,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 
 	actualDcs := make([]*cassdcapi.CassandraDatacenter, 0, len(kc.Spec.Cassandra.Datacenters))
 
-	cassClusterName := kc.Name
-	if kc.Spec.Cassandra.ClusterName != "" {
-		cassClusterName = kc.Spec.Cassandra.ClusterName
-	}
+	cassClusterName := kc.SanitizedName()
```
adutra, line 46: Very nice that you extracted this to a method 👍

adutra (changes requested):
Hmm I'm seeing some inconsistencies:

1) The `cassandra.Coalesce` function is being called:
* with `kc.SanitizedName()` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/9d8515d5f7933fce3b23e4cc8162e6f5fd37446d/controllers/k8ssandra/datacenters.go#L65);
* with `kc.Name` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/a06907f06d3dfddb8c0822a2951f9f45f1ad0f35/pkg/stargate/stargate.go#L36).

=> I think it should be `kc.SanitizedName()`.

2) What is the expectation regarding the value of `K8ssandraClusterNameLabel` ? Should it be `kc.Name` or `kc.SanitizedName()`? I found at least one inconsistency: 
* the function `k8ssandra.mustLabels` is being called with `kc.SanitizedName()` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/9d8515d5f7933fce3b23e4cc8162e6f5fd37446d/controllers/k8ssandra/cassandra_telemetry_reconciler.go#L33), which sets `K8ssandraClusterNameLabel` to `kc.SanitizedName()`;
* but it is being set to `kc.Name` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/68258b740272cac03dc4018b69fd149cabe3a0cd/pkg/cassandra/datacenter.go#L160), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L39), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L56), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L77), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/22b3324bd811bbf5d27ec80d0c21368fe4e851ef/pkg/reaper/labels.go#L18), etc.

=> I think the label should be set to `kc.Name`.

I'll be honest, imho naming derived resources after `kc.Spec.Cassandra.ClusterName` is and will always be error-prone. I'd very much preferred that we use `kc.Name` consistently. But I guess it's too late for that.

I wonder if we shouldn't introduce a type for the sanitized cluster name? This would make the code less error-prone when passing strings around. Something like:

```go
type ClusterName string

func (in ClusterName) Sanitized() string {
	return cassdcapi.CleanupForKubernetes(in)
}
```

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -179,6 +179,14 @@ func (in *K8ssandraCluster) GetInitializedDatacenters() []CassandraDatacenterTem
 	return datacenters
 }
 
+// Name returns a sanitized version of the cassandra cluster name override if it exists, otherwise the k8c object name
+func (in *K8ssandraCluster) SanitizedName() string {
+	if in.Spec.Cassandra.ClusterName != "" {
```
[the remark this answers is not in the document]
adejanovski: I think having a nil `spec.cassandra` would be an anomaly anyway, don't you?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -179,6 +179,14 @@ func (in *K8ssandraCluster) GetInitializedDatacenters() []CassandraDatacenterTem
...
+		return cassdcapi.CleanupForKubernetes(in.Spec.Cassandra.ClusterName)
+	}
+	return in.ObjectMeta.Name
```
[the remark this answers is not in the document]
adejanovski: 👌

controllers/k8ssandra/cassandra_telemetry_reconciler.go
adejanovski: Agreed, labels should carry the `kc.Name` IMO.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
adejanovski: works for me 👍

adejanovski (commented):
> The cassandra.Coalesce function is being called:
with kc.SanitizedName() [here](https://github.com/k8ssandra/k8ssandra-operator/blob/9d8515d5f7933fce3b23e4cc8162e6f5fd37446d/controllers/k8ssandra/datacenters.go#L65);
with kc.Name [here](https://github.com/k8ssandra/k8ssandra-operator/blob/a06907f06d3dfddb8c0822a2951f9f45f1ad0f35/pkg/stargate/stargate.go#L36).
=> I think it should be kc.SanitizedName().

Yes, good catch 👍 

> What is the expectation regarding the value of K8ssandraClusterNameLabel ? Should it be kc.Name or kc.SanitizedName()? I found at least one inconsistency:
the function k8ssandra.mustLabels is being called with kc.SanitizedName() [here](https://github.com/k8ssandra/k8ssandra-operator/blob/9d8515d5f7933fce3b23e4cc8162e6f5fd37446d/controllers/k8ssandra/cassandra_telemetry_reconciler.go#L33), which sets K8ssandraClusterNameLabel to kc.SanitizedName();
but it is being set to kc.Name [here](https://github.com/k8ssandra/k8ssandra-operator/blob/68258b740272cac03dc4018b69fd149cabe3a0cd/pkg/cassandra/datacenter.go#L160), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L39), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L56), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/39ce569048dfcd303a574f271d04843fa970eb7e/pkg/labels/labels.go#L77), [here](https://github.com/k8ssandra/k8ssandra-operator/blob/22b3324bd811bbf5d27ec80d0c21368fe4e851ef/pkg/reaper/labels.go#L18), etc.
=> I think the label should be set to kc.Name.

Agreed 👍 

> I'll be honest, imho naming derived resources after kc.Spec.Cassandra.ClusterName is and will always be error-prone. I'd very much preferred that we use kc.Name consistently. But I guess it's too late for that.

cass-operator doesn't allow this as it'll name resources based on the Cassandra cluster name. This led to the decision of using the same name everywhere, which you +1ed btw 😂 

> I wonder if we shouldn't introduce a type for the sanitized cluster name? This would make the code less error-prone when passing strings around. Something like:
> 
> type ClusterName string
> 
> func (in ClusterName) Sanitized() string {
>	return cassdcapi.CleanupForKubernetes(in)
>}

I'll open a follow up issue to discuss/track this.

adutra:
> cass-operator doesn't allow this as it'll name resources based on the Cassandra cluster name.

I know, and I think it is a bad choice.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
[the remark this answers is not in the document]
adutra: Yes. But it could happen e.g. in a unit test.

adutra (approved):

Issue #637, opened 2022-07-27
K8SSAND-1700 ⁃ Nil pointer exception when auth is disabled


Miles-Garnsey:
**What happened?**

When I create a cluster with `auth` enabled, then disable it, I see a NPE and the operator crashes.

**How to reproduce it (as minimally and precisely as possible)**:

1. Start a k8ssandracluster with no auth defined at all.
1. Switch to auth being explicitly disabled.
1. I think the operator crashes here.
1. If you didn’t get a crash in (3) try deleting the cluster. That was the last thing I did before I recall seeing chaos.

** K8ssandra Operator version: **

1.2.0

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1700) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1700
┆priority: Medium

Miles-Garnsey:
Here is a log of the nil pointer dereference we see: [k8ssandra-operator-75d64456d8-cs5jm.log](https://github.com/k8ssandra/k8ssandra-operator/files/9205358/k8ssandra-operator-75d64456d8-cs5jm.log)

Having reproduced this, I can confirm that you don't need to delete the pod, you simply add `spec.auth: false` and the NPE occurs.

Miles-Garnsey:
I've narrowed down to a few areas. The main one is [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/telemetry/cassandra_metrics_filters.go#L41), when auth is disabled, the PodTemplateSpec is nil and we get an NPE on the attempt to access `dcConfig.PodTemplateSpec.Spec.Containers[containerIndex].Env`.

It seemed there was also a problem [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L432), there was no guard against `dcPodTemplateSpec` being passed in nil, which then again NPEs.

jsanda:
Fixed in #640

PR #640, opened 2022-07-28
Bugfix/npe on update

Miles-Garnsey:
**What this PR does**:

Because the `cassandra.ApplyAuth` function call within `reconcileDatacenters` creates a PodTemplateSpec within the `cassandra.DatacenterConfig`, certain downstream reconciliation functions have come to depend on the creation of that struct to behave in an error free way.

When `auth==false` this causes NPEs which crash the operator in spectacular fashion. 

This PR ensures that all of those downstream functions guard against a nil `PodTemplateSpec`, consistently return errors, no longer expect to each create their own `PodTemplateSpec` and that any errors that do occur are handled gracefully.

#637 

**Checklist**
- [ ] Changes manually tested


controllers/k8ssandra/datacenters.go
```diff
@@ -63,13 +63,26 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 		// its fields are pointers, and without the copy we could end of with shared
 		// references that would lead to unexpected and incorrect values.
 		dcConfig := cassandra.Coalesce(cassClusterName, kc.Spec.Cassandra.DeepCopy(), dcTemplate.DeepCopy())
+
+		// Ensure we have a valid PodTemplateSpec before proceeding to modify it.
+		if dcConfig.PodTemplateSpec == nil {
+			dcConfig.PodTemplateSpec = &corev1.PodTemplateSpec{}
+			// we need to declare at least one container, otherwise the PodTemplateSpec struct will be invalid
```
burmanm, line 70: Why can't PodTemplateSpec stay as nil? PodTemplateSpec is not required by cass-operator.

adejanovski: We always customize things in K8ssandra-operator and always rely on the podTemplateSpec.
There's only one very edgy case where it's not true, which is when auth is disabled, medusa is disabled and metrics filters are explicitly voided with an empty list.
Doing this will simplify a lot future developments on K8ssandra-operator.

Miles-Garnsey:
@adejanovski we're seeing a single test failure here on `CreateSingleDatacenterCluster ` which (surprisingly!) appears to be legitimate.

Log: [k8ssandra-operator-b9cc64b7c-4gcp6.log](https://github.com/k8ssandra/k8ssandra-operator/files/9216210/k8ssandra-operator-b9cc64b7c-4gcp6.log)

The specific error relates to duplicate cassandra containers being created in the STS... This is likely due to the changes you asked for where we do `cassandra.UpdateCassandraContainer()` in `reconcileDatacenters()`, which has uncovered a another new bug in `AddContainersToPodTemplateSpec()`

The `single-dc` manifest is extremely complex, however, I'd say that the fact that it declares `spec.containers[0] == cassandra` may play a role. 

When the reconciliation runs, it calls `UpdateCassandraContainer` [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/bugfix/npe-on-update/controllers/k8ssandra/datacenters.go#L71), which creates the container `cassandra` in the `PodTemplateSpec`. Then, it calls `AddContainersToPodTemplateSpec()`, which adds the cassandra container again.

Because this function is defined (see [here](https://github.com/Miles-Garnsey/k8ssandra-operator/blob/bugfix/npe-on-update/pkg/cassandra/datacenter.go#L381-L389)) without the strategic merge logic we'd expect, we end up with two separate cassandra containers. FWIW, this is an error irrespective of how we've modified `reconcileDatacenters`, the fact that it was originally working is a bit of a fluke IMO.

I've dropped the STS here: [sts.txt](https://github.com/k8ssandra/k8ssandra-operator/files/9216309/sts.txt)

Here's the error it creates:

```
create Pod weirdclustername-dc1-default-sts-0 in StatefulSet weirdclustername-dc1-default-sts failed error: Pod "weirdclustername-dc1-default-sts-0" is invalid: [spec.containers[0].

image: Required value, spec.containers[1].name: Duplicate value: "cassandra"]
```

Miles-Garnsey:
@adejanovski I think this one is good to go pending tests passing (the previous round of tests did). Can you please confirm you're happy with it and I'll merge.

config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
```diff
@@ -12166,13 +12166,22 @@ spec:
                                               filters to MCAC in order to reduce the
                                               amount of extracted metrics. Not setting
                                               this field will result in the default
-                                              filters being used. Setting it to an
```
adutra, line 12169: The changes in this file and in other similar ones do not look related to the nil pointer issue at hand, do they?

k8ssandra-example.yaml
```diff
@@ -1,449 +0,0 @@
-apiVersion: v1
```
adutra, line 1: Why remove this file? Granted, it has more examples for monitoring Cassandra than for creating a K8ssandraCluster resource, but it could still be useful, don't you think?

pkg/cassandra/auth.go
```diff
@@ -13,7 +14,11 @@ import (
 const JmxInitContainer = "jmx-credentials"
 
 // ApplyAuth modifies the dc config depending on whether auth is enabled in the cluster or not.
-func ApplyAuth(dcConfig *DatacenterConfig, authEnabled bool) {
+func ApplyAuth(dcConfig *DatacenterConfig, authEnabled bool) error {
+
+	if dcConfig.PodTemplateSpec == nil {
+		return errors.New("PodTemplateSpec was nil, cannot add auth settings")
```
adutra, line 20: I'd be curious to learn what others think, but I'm torn on the idea of returning errors when `PodTemplateSpec` is nil, for a few reasons:

1. Such errors are not the user's fault but rather the developer's, and yet users would see these messages in logs, but how would they be able to fix them by themselves?
2. Such errors being the developer's fault – our fault :-) – I'd rather see them panic here, so that our unit tests can "fail spectacularly" as you say, and thus quickly detect such problems. If the issue stems from setting auth to false, maybe we'd need to improve the existing tests so that they can catch this one, and other similar nil pointer issues. But returning an error without fixing the issue would just make the controller enter an infinite loop of failed reconcile attempts.
3. Speaking of which, the real fix is in datacenters.go, and your changes look good there. IMHO the changes in that file are enough to consider the problem solved. In all places where `dcConfig.PodTemplateSpec` is read, I'd simply remove the nil checks and consider that this field is assumed to never be nil.

jsanda: I'm 👎 on returning an error in functions that work with PodTemplateSpec. I agree about the real fix being in datacenters.go and that the changes look good. 

I also agree that this is something that should be handled with test coverage vs making functions unnecessarily convoluted. If we aren't confident in our ability to ensure PodTemplateSpec will be initialized before it's use and having adequate test coverage, then maybe we shouldn't export `dcConfig.PodTemplateSpec`, i.e., change it to `dcConfig.podTemplateSpec` and provide a method on `dcConfig` for accessing it. That method would ensure it is non-nil. I'm good though with the changes in datacenters.go :)

k8ssandra-example.yaml
Miles-Garnsey: Um, so this is embarrassing, but this was a working file that I accidentally added into the repo. I was trying to (very quietly) remove it without anyone noticing. It should never have been there because k8ssandra-operator creates the SMs and the user doesn't need this example ;)

pkg/cassandra/auth.go
Miles-Garnsey: RE errors being returned: I think that's some good feedback. The invariant we are trying to maintain is that the PodTemplateSpec must never be nil. I have done this by returning a runtime error, and it is correct to say that (because of the way we handle such errors) it probably isn't quite the optimal design.

The question is how we can maintain this invariant via a compile time check, perhaps the solution is to make the PodTemplateSpec a value type instead of a pointer? That way it can never be nil and we can remove the nil checks everywhere.

The reality is that (while it may have originally made sense for the PodTemplateSpec to be a pointer as it was optional) the logic has evolved and it is no longer optional at all. We should perhaps reflect that in the types.

config/crd/bases/k8ssandra.io_k8ssandraclusters.yaml
Miles-Garnsey: I don't think they are related, no. I think the CRs got bumped automatically during the `make test` process (which has a dependency on the `manifests` target). 

Someone must have forgotten to bump them in previous iterations, but only the description is affected so no functionality will have been harmed.

Miles-Garnsey:
I've done more manual testing on this and it definitely fixes the NPEs we were seeing.

While I do appreciate the point that perhaps it is not optimal to return errors from some of the functions as this PR does, I also think that `cassandra.DatacenterConfig.PodTemplateSpec` should no longer be a pointer (since it is not optional). That's the way to fix the root cause of this problem really.

Because making PodTemplateSpec a value type is a  much bigger refactor, can I suggest that we merge this as is, then turn our hand to the refactoring tasks. Leaving this unmerged while we debate the finer points of the type system risks missing the forest for the trees a little, given that a commonly used piece of dev functionality is currently **_extremely_** broken.

k8ssandra-example.yaml
adutra: Heh :-) In this case, it's OK to remove.

adutra:
> can I suggest that we merge this as is, then turn our hand to the refactoring tasks.

Fair enough, I'll approve but please go ahead and create a follow-up issue to further refactor PodTemplateSpec.

pkg/cassandra/auth.go
adutra: For the record, we had a similar situation with `CassandraConfig` and its fields, which were all pointers until a while ago; and we had nil checks everywhere. They were subsequently turned into value types and that indeed made things much cleaner. The 2 aspects to watch when doing this change imho are: turning nilness tests into equality tests; and how things get serialized/deserialized.

adutra (approved):

PR #658, opened 2022-08-23
Add DSE support

### LINKED ISSUE #657: K8SSAND-1747 ⁃ DSE support
**What is missing?**

Cass-operator has support for DSE, but k8ssandra-operator doesn't.

**Why do we need it?**

We need it because DSE users currently cannot use K8ssandra operator and thus cannot deploy multi-dc DSE clusters using K8ssandra.

**Environment**

* K8ssandra Operator version:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1747) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1747
┆priority: Medium

adutra:
**What this PR does**:

This PR introduces basic support for DSE-specific features such as: workloads, dse.yaml.

Fixes #657 

**Checklist**
- [ ] Changes manually tested


apis/k8ssandra/v1alpha1/dseconfig_types.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type DSEYaml struct {
+
+	// TODO
+	Foo string `json:"foo,omitempty"`
```
adutra, line 22: Just a placeholder for now. Actual support for all dse.yaml options will come later.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -263,8 +268,10 @@ type CassandraDatacenterTemplate struct {
 
 // DatacenterOptions are configuration settings that are can be set at the Cluster level and overridden for a single DC
 type DatacenterOptions struct {
-	// ServerVersion is the Cassandra version.
-	// +kubebuilder:validation:Pattern=(3\.11\.\d+)|(4\.0\.\d+)
+	// ServerVersion is the Cassandra or DSE version. The following versions are supported:
+	// - Cassandra: 3.11.X and 4.0.X
+	// - DSE: 6.8.X
+	// +kubebuilder:validation:Pattern=(6\.8\.\d+)|(3\.11\.\d+)|(4\.0\.\d+)
```
adejanovski, line 274: Do we explicitly want to add a pattern on versions that is so specific?
Updating it each time a new version comes out seems a little bit overkill, doesn't it?

adejanovski (approved):
I wonder if it's a good idea to put pattern based validations at the kubebuilder level for server versions, but it's not a blocker.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Yes I agree. I copied the pattern from cass-operator, see [here](https://github.com/k8ssandra/cass-operator/blob/master/apis/cassandra/v1beta1/cassandradatacenter_types.go#L91). But I would be in favor of leaving this field free-form or just validating a general semver string.

burmanm: Hmm, we actually use this one: 

https://github.com/k8ssandra/cass-operator/blob/master/pkg/images/images.go#L27

That kubebuilder one should be removed, I don't even know if it applies..

PR #663, opened 2022-08-29
Add Cassandra workload DSE test

adejanovski:
**What this PR does**:
Adds an e2e test for the Cassandra workload in DSE.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/test/e2e/dse_test.go
+++ b/test/e2e/dse_test.go
@@ -2,7 +2,6 @@ package e2e
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
@@ -24,6 +23,6 @@ func createSingleDseDatacenterCluster(t *testing.T, ctx context.Context, namespa
 
 	t.Log("Check that we can communicate through CQL with DSE")
 	_, err = f.ExecuteCql(ctx, f.DataPlaneContexts[0], namespace, k8ssandra.SanitizedName(), DcPrefix(t, f, dcKey)+"-default-sts-0",
-		fmt.Sprintf("CREATE KEYSPACE ks1 WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : %d}", 3))
-	require.NoError(t, err, "failed to create keyspace")
+		"SELECT * FROM system.local")
+	require.NoError(t, err, "failed to execute CQL query")
 }
--- a/test/e2e/dse_test.go
+++ b/test/e2e/dse_test.go
@@ -2,15 +2,16 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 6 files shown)


test/framework/e2e_framework.go
```diff
@@ -45,14 +45,20 @@ type E2eFramework struct {
 var (
 	nodeToolStatusUN = regexp.MustCompile("UN\\s\\s")
 	nodeToolStatusDN = regexp.MustCompile("DN\\s\\s")
+	CqlshBin         = "/opt/cassandra/bin/cqlsh"
```
adutra, line 48: I think we can avoid to export this variable. Also. I'd rather see it as a private (unexported) field of the `E2eFramework` struct, rather than a global variable.

test/testdata/fixtures/single-dc-dse/k8ssandra.yaml
```diff
@@ -7,13 +7,13 @@ spec:
...
-      graphEnabled: true
-      searchEnabled: true
+      graphEnabled: false
+      searchEnabled: false
     datacenters:
       - metadata:
           name: dc1
         k8sContext: kind-k8ssandra-0
-        size: 1
+        size: 3
```
adutra, line 16: We need to be careful here, 3 DSE nodes might be too much for the GHA free tier.

test/e2e/dse_test.go
```diff
@@ -20,4 +21,9 @@ func createSingleDseDatacenterCluster(t *testing.T, ctx context.Context, namespa
 	dcKey := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
 	checkDatacenterReady(t, ctx, dcKey, f)
 	assertCassandraDatacenterK8cStatusReady(ctx, t, f, kcKey, dcKey.Name)
+
+	t.Log("Check that we can communicate through CQL with DSE")
+	_, err = f.ExecuteCql(ctx, f.DataPlaneContexts[0], namespace, k8ssandra.SanitizedName(), DcPrefix(t, f, dcKey)+"-default-sts-0",
```
adutra, line 26: This is an expensive CQL command, couldn't we just do a simple `SELECT * FROM system.peers`?

adutra (changes requested):
Looking good so far. Are you planning to run the DSE e2e test on GHA, or should we tackle this later?

test/e2e/dse_test.go
adejanovski: Done

test/framework/e2e_framework.go
adejanovski: Sure thing

adejanovski:
> Looking good so far. Are you planning to run the DSE e2e test on GHA, or should we tackle this later?

Right, I forgot that part.

All done now 👍

adutra (approved):

PR #670, opened 2022-09-07
Order datacenters by upgrade priority during reconcile

### LINKED ISSUE #669: K8SSAND-1766 ⁃ Deterministic DSE upgrades
DSE upgrades require a specific ordering depending on the DC workload, as explained [in the docs](https://docs.datastax.com/en/upgrading/docs/datastax_enterprise/upgdDsePlanning.html#_upgrade_order).

K8ssandra needs to be able to perform such upgrades appropriately, prioritizing analytics DCs, then graph/transactional ones, and lastly search DCs.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1766) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1766
┆priority: Medium

adejanovski:
**What this PR does**:
DSE requires a specific ordering for upgrades, depending on which workload (graph, transactional, search or analytics) the DC is running.
This PR assigns a priority to each type of workload and orders datacenters reconciliation by that priority.
If no workload is assigned to the datacenters, the ordering remains unchanged.

Fixes #669 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/datacenters.go
+++ b/controllers/k8ssandra/datacenters.go
@@ -441,10 +441,12 @@ func newRebuildTask(targetDc, namespace, srcDc string, numNodes int) *cassctlapi
 // The datacenters with the highest priority are first in the list.
 // The datacenters with the lowest priority are last in the list.
 func sortDatacentersByPriority(datacenters []api.CassandraDatacenterTemplate) []api.CassandraDatacenterTemplate {
-	sort.Slice(datacenters, func(i, j int) bool {
-		return dcUpgradePriority(datacenters[i]) > dcUpgradePriority(datacenters[j])
+	sortedDatacenters := make([]api.CassandraDatacenterTemplate, len(datacenters))
+	copy(sortedDatacenters, datacenters)
+	sort.Slice(sortedDatacenters, func(i, j int) bool {
+		return dcUpgradePriority(sortedDatacenters[i]) > dcUpgradePriority(sortedDatacenters[j])
 	})
-	return datacenters
+	return sortedDatacenters
 }
 
 // dcUpgradePriority returns the upgrade priority for the given datacenter.
--- a/controllers/k8ssandra/datacenters.go
+++ b/controllers/k8ssandra/datacenters.go
@@ -3,6 +3,7 @@ package k8ssandra
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -52,7 +53,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 	}
 
 	// Reconcile CassandraDatacenter objects only
-	for idx, dcTemplate := range kc.Spec.Cassandra.Datacenters {
+	for idx, dcTemplate := range sortDatacentersByPriority(kc.Spec.Cassandra.Datacenters) {
 		if !secret.HasReplicatedSecrets(ctx, r.Client, kcKey, dcTemplate.K8sContext) {
 			// ReplicatedSecret has not replicated yet, wait until it has
 			logger.Info("Waiting for replication to complete")
@@ -433,4 +434,33 @@ func newRebuildTask(targetDc, namespace, srcDc string, numNodes int) *cassctlapi
 	annotations.AddHashAnnotation(task)
 
 	return task
+}
+
+// sortDatacentersByPriority sorts the datacenters by upgrade priority.
+// The datacenters are sorted in descending order of priority.
+// The datacenters with the highest priority are first in the list.
+// The datacenters with the lowest priority are last in the list.
+func sortDatacentersByPriority(datacenters []api.CassandraDatacenterTemplate) []api.CassandraDatacenterTemplate {
+	sort.Slice(datacenters, func(i, j int) bool {
+		return dcUpgradePriority(datacenters[i]) > dcUpgradePriority(datacenters[j])
+	})
+	return datacenters
+}
+
+// dcUpgradePriority returns the upgrade priority for the given datacenter.
+func dcUpgradePriority(dc api.CassandraDatacenterTemplate) int {
+	if dc.DseWorkloads == nil {
+		// Cassandra workload
+		return 2
+	}
+
+	if dc.DseWorkloads.AnalyticsEnabled {
+		return 3
+	}
+
+	if dc.DseWorkloads.SearchEnabled {
+		return 1
+	}
+
+	return 2
 }
--- a/controllers/k8ssandra/datacenters_test.go
+++ b/controllers/k8ssandra/datacenters_test.go
@@ -1,0 +1,139 @@
+package k8ssandra
+
+import (
+	"context"
+	"testing"
+
+	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
+	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	"github.com/k8ssandra/k8ssandra-operator/test/framework"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	analyticsWorkloadDc = api.CassandraDatacenterTemplate{
+		DatacenterOptions: api.DatacenterOptions{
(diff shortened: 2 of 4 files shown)


controllers/k8ssandra/datacenters_test.go
```diff
@@ -0,0 +1,139 @@
...
+			DseWorkloads: &cassdcapi.DseWorkloads{
+				GraphEnabled:     false,
+				AnalyticsEnabled: false,
+				SearchEnabled:    false,
+			},
+		},
+	}
+)
+
+func datacentersTest(t *testing.T, ctx context.Context, f *framework.Framework, namespace string) {
```
adutra, line 63: These are actually unit tests, not integration tests (in the sense of: tests using the testEnv framework). You should change the signature here to:

```go
func TestDatacenters(t *testing.T) {
	t.Run("DcUpgradePriorityTest", dcUpgradePriorityTest)
	t.Run("SortDatacentersForUpgradeTest", sortDatacentersForUpgradeTest)
	t.Run("SortNoChangeTest", sortNoChangeTest)
}
```

The advantage is that you can run them without spinning up an API server for nothing and so it's going to be much faster.

controllers/k8ssandra/datacenters.go
```diff
@@ -434,3 +435,32 @@ func newRebuildTask(targetDc, namespace, srcDc string, numNodes int) *cassctlapi
 
 	return task
 }
+
+// sortDatacentersByPriority sorts the datacenters by upgrade priority.
+// The datacenters are sorted in descending order of priority.
+// The datacenters with the highest priority are first in the list.
+// The datacenters with the lowest priority are last in the list.
+func sortDatacentersByPriority(datacenters []api.CassandraDatacenterTemplate) []api.CassandraDatacenterTemplate {
+	sort.Slice(datacenters, func(i, j int) bool {
```
adutra, line 444: `sort.Slice` performs an in-place sort and that's going to be a problem because this means that calling `sortDatacentersByPriority(kc.Spec.Cassandra.Datacenters)` will modify the input `kc.Spec.Cassandra.Datacenters`. You should instead create a copy of the slice inside the function.

controllers/k8ssandra/datacenters.go
```diff
@@ -52,7 +53,7 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 	}
 
 	// Reconcile CassandraDatacenter objects only
-	for idx, dcTemplate := range kc.Spec.Cassandra.Datacenters {
+	for idx, dcTemplate := range sortDatacentersByPriority(kc.Spec.Cassandra.Datacenters) {
```
adutra, line 56: Open question: this means that dcs will always be reconciled in this particular order, even if no upgrades are happening. IOW, when creating a new DSE cluster, we would first create the Analytics DCs, then the Cassandra/Graph ones, then the Search ones. But is it OK to bootstrap an Analytics DC before the Cassandra ones?

controllers/k8ssandra/datacenters_test.go
```diff
@@ -0,0 +1,139 @@
...
+	datacenters = append(datacenters, searchWorkloadDc)
+	datacenters = append(datacenters, graphWorkloadDc)
+	datacenters = append(datacenters, mixedWorkloadDc)
+	sortedDatacenters := sortDatacentersByPriority(datacenters)
+	assert.Equal(5, len(sortedDatacenters), "Expected 5 datacenters")
+	// The analytics enabled DCs should be first, which includes the mixed workload DC
+	assert.True(sortedDatacenters[0].DseWorkloads.AnalyticsEnabled, "Analytics workload DCs should be first")
+	assert.True(sortedDatacenters[1].DseWorkloads.AnalyticsEnabled, "Analytics workload DCs should be first")
+	// Second and third datacenters should be Cassandra and Graph only
+	assert.False(sortedDatacenters[2].DseWorkloads.AnalyticsEnabled || sortedDatacenters[2].DseWorkloads.SearchEnabled, "Analytics workload DC should be first and search should be last")
```
adutra, line 94: The assertions are a bit weird. I have a suggestion: instead of testing the values of these boolean fields, why don't you given distinctive names to each dc, then test the names?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2216,3 +2218,177 @@ func injectContainersAndVolumes(t *testing.T, ctx context.Context, f *framework.
...
+
+	verifyFinalizerAdded(ctx, t, f, client.ObjectKey{Namespace: kc.Namespace, Name: kc.Name})
+
+	verifySuperuserSecretCreated(ctx, t, f, kc)
+
+	verifyReplicatedSecretReconciled(ctx, t, f, kc)
+
+	verifySystemReplicationAnnotationSet(ctx, t, f, kc)
+
+	t.Log("check that dc1 was created")
```
adutra, line 2289:
```suggestion
	t.Log("check that dc2 was created")
```

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2216,3 +2218,177 @@ func injectContainersAndVolumes(t *testing.T, ctx context.Context, f *framework.
...
+	dc2Key := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc2"}, K8sContext: f.DataPlaneContexts[1]}
+	require.Eventually(f.DatacenterExists(ctx, dc2Key), timeout, interval)
+
+	t.Log("update datacenter status to scaling up")
```
adutra, line 2293: IMHO this test is about how dcs are created, but the intermediary state of a dc being bootstrapped is less relevant. I would remove lines 2293 to 2333 and go straight to the point were we test that dc1 is not created until dc2 is ready.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2216,3 +2218,177 @@ func injectContainersAndVolumes(t *testing.T, ctx context.Context, f *framework.
...
+									StorageClassName: &defaultStorageClass,
+								},
+							},
+							DseWorkloads: &cassdcapi.DseWorkloads{
+								SearchEnabled: true,
+							},
+						},
+					},
+					{
+						Meta: api.EmbeddedObjectMeta{
```
adutra, line 2256: Shouldn't we add more dcs and also test a dc with Cassandra and a dc with Graph workloads?

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2216,3 +2218,177 @@ func injectContainersAndVolumes(t *testing.T, ctx context.Context, f *framework.
...
+
+	t.Log("check that dc1 was created")
+	require.Eventually(f.DatacenterExists(ctx, dc1Key), timeout, interval)
+
+	t.Log("update dc1 status to ready")
+	err = f.SetDatacenterStatusReady(ctx, dc1Key)
+	require.NoError(err, "failed to set dc1 status ready")
+
+	t.Log("check that the K8ssandraCluster status is updated")
+	require.Eventually(func() bool {
```
adutra, line 2351: Same here, I think the k8c status is well tested already, I'm not convinced that we must test it here.

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: kk 👍

adejanovski (commented):
I've pushed a new commit to address most of the changes you suggested.

controllers/k8ssandra/datacenters.go
burmanm: No, that's not how Golang slices work. You pass as value the reference to the slice. Even if you would do ``datacenters := datacenters``, you would only create a copy of the reference - not the underlying data.

https://go.dev/doc/effective_go#slices

So if you wish to copy the entire slice's data, you need to use copy() function.

```golang
datCenters := make([]type, len(datacenters))
copy(datacenters, datCenters)
```

adutra (approved):

PR #672, opened 2022-09-07
Enable code coverage in codecov

adejanovski:
**What this PR does**:
Adds code coverage reports using codecov.

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/test_and_build_image.yaml
+++ b/.github/workflows/test_and_build_image.yaml
@@ -41,6 +41,8 @@ jobs:
         run: |
           export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
           make envtest test
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
   build_image:
     name: Build and push image to Docker Hub
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Miles-Garnsey:
It is hard for me to assess if this PR provides everything we need since some of the critical features we need are going to be visible only after the `codecov.yaml` file is merged on `main`. Because this PR doesn't seem to actively break anything I am approving and hopefully some of these things (e.g. coverage stats integrated into the GH UI) will pop up once it is merged.

One note: we may want to avoid the use of `covermode=atomic` as apparently (according to [here](https://go.dev/blog/cover)) it can have performance implications. We ideally want the tests to remain fast (although our unit tests are already quite fast, so maybe this is not a concern).

Miles-Garnsey (approved):
Approving, please double check whether you really want `covermode=atomic` before merging.

adejanovski:
> Approving, please double check whether you really want `covermode=atomic` before merging.

Thanks, I checked and the execution time of the unit/integration tests didn't change. It's around 7 minutes with and without `-covermode=atomic`.

PR #678, opened 2022-09-12
Pick the right default repository for pulling DSE images

### LINKED ISSUE #677: K8SSAND-1775 ⁃ Force the docker repo in k8ssandra-operator when serverType is dse
Since we’re building new DSE images with an up to date management api, we need to enforce the docker hub repo where the images will get pulled to [Docker Hub](https://hub.docker.com/r/datastax/dse-mgmtapi-6_8) 
This will be done at the k8ssandra-operator level, as cass-operator can (and should) run both types of images without a problem.
Using dse images with an up to date mgmt api would allow us to perform multi DC deployments.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1775) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1775
┆priority: Medium

adejanovski:
**What this PR does**:
Enforces a default DSE docker hub repository to pull images with the new management api.

Fixes #677

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -240,7 +240,7 @@ type CassandraClusterTemplate struct {
 	// Server type: "cassandra" or "dse".
 	// +kubebuilder:validation:Enum=cassandra;dse
 	// +kubebuilder:default=cassandra
-	ServerType Distribution `json:"serverType,omitempty"`
+	ServerType ServerDistribution `json:"serverType,omitempty"`
 }
 
 // +kubebuilder:pruning:PreserveUnknownFields
@@ -422,9 +422,9 @@ func init() {
 	SchemeBuilder.Register(&K8ssandraCluster{}, &K8ssandraClusterList{})
 }
 
-type Distribution string
+type ServerDistribution string
 
 const (
-	DistributionCassandra = Distribution("cassandra")
-	DistributionDse       = Distribution("dse")
+	ServerDistributionCassandra = ServerDistribution("cassandra")
+	ServerDistributionDse       = ServerDistribution("dse")
 )
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -240,7 +240,7 @@ type CassandraClusterTemplate struct {
 	// Server type: "cassandra" or "dse".
 	// +kubebuilder:validation:Enum=cassandra;dse
 	// +kubebuilder:default=cassandra
-	ServerType string `json:"serverType,omitempty"`
+	ServerType Distribution `json:"serverType,omitempty"`
 }
 
 // +kubebuilder:pruning:PreserveUnknownFields
@@ -421,3 +421,10 @@ func (s *K8ssandraClusterStatus) SetCondition(condition K8ssandraClusterConditio
 func init() {
 	SchemeBuilder.Register(&K8ssandraCluster{}, &K8ssandraClusterList{})
 }
+
+type Distribution string
+
+const (
+	DistributionCassandra = Distribution("cassandra")
+	DistributionDse       = Distribution("dse")
+)
--- a/config/cass-operator/cluster-scoped/kustomization.yaml
+++ b/config/cass-operator/cluster-scoped/kustomization.yaml
@@ -4,6 +4,10 @@ kind: Kustomization
 resources:
 - github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=master
 
+
 images:
   - name: k8ssandra/cass-operator
     newTag: v1.12.0
+
+components:
+  - ../../components/cass-operator-image-config
--- a/config/cass-operator/ns-scoped/kustomization.yaml
+++ b/config/cass-operator/ns-scoped/kustomization.yaml
@@ -7,3 +7,6 @@ resources:
 images:
   - name: k8ssandra/cass-operator
     newTag: v1.12.0
+
+components:
+  - ../../components/cass-operator-image-config
--- a/config/components/cass-operator-image-config/image_config.yaml
+++ b/config/components/cass-operator-image-config/image_config.yaml
@@ -1,0 +1,22 @@
+apiVersion: config.k8ssandra.io/v1beta1
+kind: ImageConfig
+metadata:
+  name: image-config
+images:
+  system-logger: "k8ssandra/system-logger:latest"
+  config-builder: "datastax/cass-config-builder:1.0.4-ubi7"
+  # cassandra:
+  #   "4.0.0": "k8ssandra/cassandra-ubi:latest"
+  # dse:
+  #   "6.8.999": "datastax/dse-server-prototype:latest"
+# imageRegistry: "localhost:5000"
+# imagePullPolicy: Always
+# imagePullSecret:
+#   name: my-secret-pull-registry
+defaults:
+  # Note, postfix is ignored if repository is not set
+  cassandra:
+    repository: "k8ssandra/cass-management-api"
+  dse:
+    repository: "datastax/dse-mgmtapi-6_8"
+    suffix: "-jdk8"
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -56,7 +56,7 @@ func addNumTokens(template *DatacenterConfig, cfg *cassConfig) {
 	// explicitly set it because the config builder defaults to num_tokens: 1
 	if cfg.CassandraYaml.NumTokens == nil {
 		version := template.ServerVersion
-		if template.ServerType == api.DistributionCassandra && version.Major() == 3 {
+		if template.ServerType == api.ServerDistributionCassandra && version.Major() == 3 {
 			cfg.CassandraYaml.NumTokens = pointer.Int(256)
 		} else {
 			cfg.CassandraYaml.NumTokens = pointer.Int(16)
--- a/pkg/cassandra/config.go
+++ b/pkg/cassandra/config.go
@@ -28,7 +28,7 @@ func CreateJsonConfig(apiConfig *DatacenterConfig) ([]byte, error) {
 	addNumTokens(apiConfig, cfg)
 	addEncryptionOptions(apiConfig, cfg)
 	handleDeprecatedJvmOptions(cfg)
-	if out, err := preMarshalConfig(reflect.ValueOf(cfg), apiConfig.ServerVersion, apiConfig.ServerType); err != nil {
+	if out, err := preMarshalConfig(reflect.ValueOf(cfg), apiConfig.ServerVersion, string(apiConfig.ServerType)); err != nil {
 		return nil, err
 	} else {
 		return json.Marshal(out)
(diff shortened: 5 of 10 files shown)


pkg/cassandra/datacenter.go
```diff
@@ -397,6 +397,16 @@ func AddContainersToPodTemplateSpec(dcConfig *DatacenterConfig, containers []cor
...
+		if dcConfig.ServerType == api.DistributionDse {
+			index, found := FindContainer(dcConfig.PodTemplateSpec, "cassandra")
+			if found {
+				env := dcConfig.PodTemplateSpec.Spec.Containers[index].Env
+				cassandraConfEnvVar := utils.FindEnvVar(env, "CASSANDRA_CONF")
+				if cassandraConfEnvVar == nil {
+					dcConfig.PodTemplateSpec.Spec.Containers[index].Env = append(dcConfig.PodTemplateSpec.Spec.Containers[index].Env, corev1.EnvVar{Name: "CASSANDRA_CONF", Value: "/config"})
+				}
+			}
+		}
```
adejanovski, line 409: The images we're using right now aren't creating the appropriate symlinks for DSE to find the config files.
This hack will allow it to read the `cassandra.yaml` file in the right place, but not the `dse.yaml` file.

It will be enough to move forward with multi DC related work.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -416,3 +416,14 @@ func (s *K8ssandraClusterStatus) SetCondition(condition K8ssandraClusterConditio
 func init() {
 	SchemeBuilder.Register(&K8ssandraCluster{}, &K8ssandraClusterList{})
 }
+
+type Distribution string
```
jeffbanks, line 420: Curious on this one, why not Distribution as a structure with name, version, etc.?

adejanovski: Wouldn't it mean that `Cassandra 4.0` and `Cassandra 4.1` are considered as two different distributions?
I'm just trying to improve the typing of our `ServerType` in order to avoid repeating `cassandra` and `dse` throughout the code here, not do a refactor on how `ServerType`, `ServerVersion` and `ServerImage` are related to each other.

adutra (dismissed):
I'm approving to get things moving but see my last comment: I'm concerned that we will need to keep the new kustomizations in sync with what's in cass-operator.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -416,3 +416,14 @@ func (s *K8ssandraClusterStatus) SetCondition(condition K8ssandraClusterConditio
...
+
+const (
+	DistributionCassandra = Distribution("cassandra")
+	DistributionDse       = Distribution("dse")
+)
+
+func (d Distribution) String() string {
```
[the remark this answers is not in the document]
adejanovski: right, I'll use simple casts then and remove this :+1:

controllers/k8ssandra/datacenters.go
```diff
@@ -118,6 +118,16 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 			return result.Error(err), actualDcs
 		}
 
+		// Temporary hack: force Cassandra config location
+		// TODO: Remove once the DSE images are updated with the appropriate symlinks
+		// Inject MCAC metrics filters
```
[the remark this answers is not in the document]
adejanovski: > Is this comment accurate?

it is not 😅

config/cass-operator/cluster-scoped/kustomization.yaml
```diff
@@ -3,7 +3,38 @@ kind: Kustomization
 
 resources:
 - github.com/k8ssandra/cass-operator/config/deployments/cluster?ref=master
-
+patches:
```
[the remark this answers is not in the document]
adejanovski: I share all of your concerns and this was definitely tricky for me to figure out, but it seems that "this is the way" 🤷 
Happy to start a discussion on how this could be improved.

config/cass-operator/ns-scoped/kustomization.yaml
```diff
@@ -3,7 +3,38 @@ kind: Kustomization
 
 resources:
 - github.com/k8ssandra/cass-operator/config/deployments/default?ref=master
-
+patches:
```
jsanda, line 6: Is this patch the same as the one for cluster-scoped? If so then it would be nice to have the cluster-scoped and ns-scoped kustomizations share and reuse the same patch.

config/cass-operator/cluster-scoped/kustomization.yaml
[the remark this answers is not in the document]
jsanda: I'm not sure I understand the concerns. The overall approach is to extend the ComponentConfig as describe in the kubebuilder book [here](https://book.kubebuilder.io/component-config-tutorial/tutorial.html). 

Why would users need to change the kustomization file? ImageConfig declares defaults. If I want to change those defaults I can simply modify the cass-operator-manager-config ConfigMap. I would really prefer to see something like this over what we have in k8ssandra-operator. If I want images to always be pulled from a private registry, I only need to modify the ConfigMap. In k8ssandra-operator we either have to supply the image coordinates with each K8ssandraCluster (not DRY), or we have to recompile the operator to change the defaults defined in Go code. On top of that, default image coordinates are not centralized in one place in k8ssandra-operator.

Lastly, I'm not convinced of the need for a CRD; however, we really need docs to explain the structure and fields.

jsanda:
>I'm approving to get things moving but see my last comment: I'm concerned that we will need to keep the new kustomizations in sync with what's in cass-operator.

This is a valid concern, and I read that comment after I wrote this [comment](https://github.com/k8ssandra/k8ssandra-operator/pull/678#discussion_r970196040). Does the fact that the ImageConfig is declared as a multi-line scalar value require us to patch the whole thing? If so and if a more structured config would allow for a simpler patch, then I think it is certainly something to consider.

config/cass-operator/ns-scoped/kustomization.yaml
adejanovski: It is, but I don't know how to do what you're describing.
There seem to be 2 issues:
- patches need to be in a subdirectory of a kustomization (but we have two different directories here)
- Kustomize doesn't seem to read patches from a file, it only seems to support inline patches.

How would you do this specifically?

burmanm: What do you mean Kustomize doesn't seem to read patches from a file? Isn't that pretty common way of adding patches? At least strategicMergePatches can be in a file separately, for example in cass-operator in the default that's the case:

https://github.com/k8ssandra/cass-operator/tree/master/config/default

But even for json6902 patches that's doable, as seen here:

https://github.com/k8ssandra/cass-operator/blob/master/config/crd/kustomization.yaml#L21

Patches also don't need to be a in a subdirectory of that path, you can do a common directory. Key here is that the default parameters given to Kustomize build prevent this, but kustomize does allow overriding that. In cass-operator kustomize templates, there's plenty of these shared resources between different packages.

Run kustomize with: ``build --load-restrictor LoadRestrictionsNone `` and it should be happy to read files from pretty much anywhere.

config/cass-operator/cluster-scoped/kustomization.yaml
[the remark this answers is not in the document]
adutra: I think we all agree we have a problem of insufficient documentation around operator configuration.

My main concern here is that the actual requirement is to change 2 lines in image_config.yaml:

```
defaults.dse.repository = "datastax/dse-mgmtapi-6_8"
defaults.dse.suffix     = ""
```

Unfortunately, that is not possible with kustomize (or is it?). We are then forced to re-declare the entire contents of image_config.yaml here. This is imho a fragile design that will force us to keep this file in sync with cass-operator every time the original file is updated with new defaults.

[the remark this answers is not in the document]
burmanm: The structure of the imageConfig is defined here: https://github.com/k8ssandra/cass-operator/blob/master/apis/config/v1beta1/imageconfig_types.go

Same as every CRD. We simply don't generate the Kubernetes yaml presentation.

In our Helm charts, we allow modification of single values in the imageConfig, since Helm renders the file. If you have to add all the fields in a patch, then that's more a limitation of Kustomize (and why it's not necessarily a good tool to deploy anything that requires kustomization (tm)). The issue is that Kustomize does not allow modifying files before the generator reads the input. 

That said, this approach seems complex what's done here. Wouldn't it be easier to do it like we patch the other field in this same ConfigMap when webhook is enabled?

https://github.com/k8ssandra/cass-operator/tree/master/config/components/webhook

It uses a file as an input to modify the OperatorConfig.

config/cass-operator/ns-scoped/kustomization.yaml
```diff
@@ -3,38 +3,10 @@ kind: Kustomization
...
-            repository: "k8ssandra/cass-management-api"
-          dse:
-            repository: "datastax/dse-mgmtapi-6_8"
-            suffix: ""
+
 images:
   - name: k8ssandra/cass-operator
     newTag: v1.12.0
+
+components:
```
jsanda, line 11: Using the component makes it possible to reuse the ImageConfig patch across the cass-operator overlays. To do our due diligence we should also make sure it is possible for someone using k8ssandra-operator to patch and change the ImageConfig.

config/components/cass-operator-image-config/image_config.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: config.k8ssandra.io/v1beta1
+kind: ImageConfig
+metadata:
+  name: image-config
+images:
+  system-logger: "k8ssandra/system-logger:latest"
```
adutra, line 6: I did a small research and it seems it's going to be tricky to avoid recopying the whole file here.

See this issue: https://github.com/kubernetes-sigs/kustomize/issues/2962.

burmanm: But this file can be created and modified with kustomize. It just can't be modified in the same kustomize build command, but one needs to execute that kustomize command twice.

pkg/cassandra/config.go
```diff
@@ -56,7 +56,7 @@ func addNumTokens(template *DatacenterConfig, cfg *cassConfig) {
 	// explicitly set it because the config builder defaults to num_tokens: 1
 	if cfg.CassandraYaml.NumTokens == nil {
 		version := template.ServerVersion
-		if strings.ToLower(template.ServerType) == "cassandra" && version.Major() == 3 {
+		if template.ServerType == api.DistributionCassandra && version.Major() == 3 {
```
burmanm, line 59: A note on the naming.. here we have refer the Distribution with two different names. On the CRD it's "ServerType", yet on that enum it's "Distribution". This feels odd. Is it a type of server or is it just a different distribution of the same server?

burmanm: And does distribution mean deb-packaged Cassandra vs. tar-Cassandra? Perhaps an OSS Cassandra packaged by soem vendor like there used to be ?

adejanovski: The fact that we're using "ServerType" in the CRD doesn't make Distribution irrelevant here. Also, the term distribution is not a matter of how a software is packaged, but rather how vanilla or not it is compared to the upstream project.
It makes sense in the Linux world, just as it seems to me it makes sense in the Cassandra world.

It's not a hill I'll die on though, so if this is a blocker for you I can change this to `ServerTypeDSE` and `ServerTypeCassandra`.

adejanovski: @burmanm, what do you think?

adejanovski: Also, if we use `ServerType` as type, we'll have the `ServerType` field that has the `ServerType` type 😕 not really a great situation. I think that's why I came up with an alternate name.

config/components/cass-operator-image-config/image_config.yaml
adutra: How would you create this file with kustomize?

burmanm (approved):

PR #666, opened 2022-08-30
Add DSE graph workload e2e tests

adejanovski:
**What this PR does**:
Adds an e2e test for the Graph DSE workload.
This PR introduces a new dependency on gremlin-go to test the connection to the graph endpoint and execute some queries against it.

**Which issue(s) this PR fixes**:
Fixes #665 

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [ ] CHANGELOG.md updated (not required for documentation PRs)
- [ ] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #679, opened 2022-09-13
Add gRPC port 8090 to the Stargate service

### LINKED ISSUE #648: K8SSAND-1726 ⁃ Open GRPC port within the stargate deployment (&fix external access)
**What is missing?**
GRPC support in stargate is amazing but unfortunately we can't use it with the cassandra operator.

The service is not opening a port to GRPC.

**Why do we need it?**

Supporting GRPC endpoints

**Environment**

* Cass Operator version:
   helm 0.38.0
  version 1.2.0

Note: Even by creating your own service to support it, it seems like only localhost is supported, not external ip addresses.

I managed to connect to stargate through `kubectl port-forward` but not with an external IP address.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1726) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1726
┆priority: Medium

olim7t:
**What this PR does**:
Add gRPC port (8090) to the Stargate service.

Fixes #648

**Checklist**
- [x] Changes manually tested


pkg/stargate/deployments.go
```diff
@@ -154,7 +154,7 @@ func NewDeployments(stargate *api.Stargate, dc *cassdcapi.CassandraDatacenter) m
 								{ContainerPort: 8082, Name: "rest"},
 								{ContainerPort: 8084, Name: "health"},
 								{ContainerPort: 8085, Name: "metrics"},
-								{ContainerPort: 8090, Name: "http-schemaless"},
+								{ContainerPort: 8090, Name: "grpc"},
```
olim7t, line 157: Not directly part of the fix, but I think this should be changed too.

adejanovski (commented):
@olim7t, the changes look good.
It would be nice to add some checks around the grpc api in the e2e tests, which will be a little more involved.
You can use this [PR](https://github.com/k8ssandra/k8ssandra-operator/pull/666/files) to guide you. It was similar in the sense that the test required to add a new ingress in the e2e tests to test the graph API of DSE.
You won't need to add a brand new e2e test for this, but add the grpc API in the tested Stargate APIs: https://github.com/k8ssandra/k8ssandra-operator/blob/main/test/e2e/stargate_apis_test.go#L29
This way, all tests that involve Stargate will check the grpc api as well.

olim7t:
I went down a rabbit hole with the e2e test. Configuring ingress for gRPC turns out to be pretty complicated. Here are my findings so far:

### Stargate
Stargate exposes its gRPC API on HTTP port 8090, without TLS. It is implemented with the grpc-java library, which is itself based on Netty.
I can write a client with grpc-java and connect to a local Stargate instance (no proxying) over a plaintext connection. There is a [NegotiationType](https://grpc.github.io/grpc-java/javadoc/io/grpc/netty/NegotiationType.html) object that can be passed when building the client channel:
- `PLAINTEXT`: "*Just assume the connection is plaintext (non-SSL) and the remote endpoint supports HTTP/2 directly without an upgrade.*" This is the default, and it works.
- `PLAINTEXT_UPGRADE`: "*Use the HTTP UPGRADE protocol for a plaintext (non-SSL) upgrade from HTTP/1.1 to HTTP/2.*" This one I think refers to the [h2c](https://www.rfc-editor.org/rfc/rfc7540#section-3.2) token described in the spec. It doesn't work.

From what I understand, `PLAINTEXT` is used when both parties know beforehand that the protocol will be HTTP/2, whereas `PLAINTEXT_UPGRADE` is a way to discover it dynamically.

I also found a comment on a [grpc-java issue](https://github.com/grpc/grpc-java/issues/8778) that confirms that the upgrade doesn't work:
> the grpc java server supports plaintext HTTP/2, just not the HTTP/1 to HTTP/2 upgrade

### Nginx
Before I started looking at the e2e test code, I configured a cluster manually to test my change. I used NGINX for ingress and was able to configure it successfully:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
  name: stargate-grpc-ingress
spec:
  ingressClassName: nginx
  rules:
  - host: grpc.ingress.local
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: demo-dc1-stargate-service
            port:
              number: 8090
  tls:
  - secretName: grpc-tls-secret
    hosts:
    - grpc.ingress.local
```

[Ingress-nginx does not support HTTP/2 without TLS](https://stackoverflow.com/questions/67429657/how-to-enable-http2-grpc-on-port-80-i-e-without-tls-with-nginx-ingress), so I had to generate certificates and have the TLS section at the bottom. However, TLS is terminated at the proxy level, and NGINX proceeds to Stargate on the HTTP port, with what I assume is the "plaintext / no upgrade" protocol.

### Traefik
The e2e test infrastructure uses Traefik for ingress. I tried replicating the configuration for the other Stargate ports; gRPC traffic gets correctly routed to the Stargate pod, but the last section is done with HTTP1 and I see this error in the Stargate logs:
```
Sep 14, 2022 10:21:48 PM io.grpc.netty.shaded.io.grpc.netty.NettyServerTransport notifyTerminated
INFO: Transport failed
io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception: Unexpected HTTP/1.x request: POST /stargate.Stargate/ExecuteQuery
```
I tried adding a [ServersTransport](https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-serverstransport) with `disableHTTP2=false`, but that doesn't help.

The [Traefik docs](https://doc.traefik.io/traefik/user-guides/grpc/) suggest two approaches for gRPC:
- with plain HTTP, using h2c for the backend: I doubt this would work since grpc-java doesn't support h2c.
- with HTTPS: their example uses HTTPS for the backend too, which wouldn't work for us since the Stargate pod exposes HTTP ports only. I'm not sure if there's a way to terminate TLS in the proxy like in the NGINX example.

jsanda:
Is there a port on which Stargate listens for HTTPS traffic? How is Stargate configured to handle TLS?

It would be nice to use HTTPS since that is what folks would/should use for production. If you are able to figure something out, it could make for a nice blog post how to configure ingress for Stargate :)

olim7t:
Rebased on top of #689 in order to implement the E2E test.
I will clean up the history once the other PR is merged.

jsanda (approved):
Nice work!

olim7t:
This was discussed on other channels, but documenting here for traceability:
- I ended up switching ingress to NGINX in #689, which allowed me to use the solution described above.
- > Is there a port on which Stargate listens for HTTPS traffic? How is Stargate configured to handle TLS?
    
    Stargate does not support HTTPS out of the box. I think the simplest approach would be to terminate TLS at the ingress level and then proxy to Stargate's plaintext ports as in the gRPC example.

PR #680, opened 2022-09-14
Update "Running e2e tests" section in developer docs

olim7t:
**What this PR does**:
Update developer doc: `make kind e2e-test` does not exist anymore, but there are newer targets that do all the configuration automatically.

n/a

**Checklist**
- ~Changes manually tested~
- ~Automated Tests added/updated~
- ~CHANGELOG.md updated (not required for documentation PRs)~

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -99,65 +99,63 @@ Integration tests use the `envtest` package from controller-runtime.  See this [
 **Note:** If you want to run integration tests from your IDE you need to set the `KUBEBUILDER_ASSETS` env var. It should point to `<project-root>/testbin/bin`.
 
 ## Running e2e tests
-End-to-end tests require kind clusters that are built with the `scripts/setup-kind-multicluster.sh` script. 
+End-to-end tests run against local kind clusters. There are two possible setups:
+* "multi-cluster" tests require two kind clusters;
+* "single-cluster" tests only need one kind cluster.
 
-**Note:** There are plans to add the ability to run the tests against other clusters. This is being tracked in [#112](https://github.com/k8ssandra/k8ssandra-operator/issues/112).
+The name of the test will generally indicate its type. You can also find a complete list in the GitHub action
+definitions (look for the `matrix` property in
+[kind_multicluster_e2e_tests.yaml](../../.github/workflows/kind_multicluster_e2e_tests.yaml) or
+[kind_e2e_tests.yaml](../../.github/workflows/kind_e2e_tests.yaml)).
 
-### Automated procedure
+### Resource Requirements
+Multi-cluster tests will be more resource intensive than other tests. The Docker VM used to develop and run these tests
+on a MacBook Pro is configured with 6 CPUs and 10 GB of memory. Your mileage may vary on other operating systems/setups.
 
-The makefile has a target which will create the kind clusters (deleting them first if they already exist), build the docker image and load it into both clusters before running the e2e tests.
-Just run the following:
+### Multi-cluster tests
 
-```
-make kind-e2e-test
-```
+The makefile has targets to create and configure the kind clusters (note that this automates the steps detailed at the
+beginning of this document):
 
-If you want to run a single test, set the `E2E_TEST` variable as follows:
+```shell
+# - Create kind clusters (deleting any existing ones first)
+# - Install Cert Manager and Traefik (used for ingress in e2e tests)
+make multi-create
 
-```
-make E2E_TEST=TestOperator/SingleDatacenterCluster kind-e2e-test
-```
-
-### Manual procedure 
-
-#### Create the kind clusters
-The multi-cluster tests require two clusters.
-
-```
-./scripts/setup-kind-multicluster.sh --clusters 2
-```
-
-#### Build operator image
-
-Before running tests, build the operator image:
-
-```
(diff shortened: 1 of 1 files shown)


docs/developer/README.md
```diff
@@ -99,65 +99,72 @@ Integration tests use the `envtest` package from controller-runtime.  See this [
 **Note:** If you want to run integration tests from your IDE you need to set the `KUBEBUILDER_ASSETS` env var. It should point to `<project-root>/testbin/bin`.
 
 ## Running e2e tests
-End-to-end tests require kind clusters that are built with the `scripts/setup-kind-multicluster.sh` script. 
+End-to-end tests run against local kind clusters. There are two possible setups:
+* "multi-cluster" tests require two kind clusters;
+* "single-cluster" tests only need one kind cluster.
 
-**Note:** There are plans to add the ability to run the tests against other clusters. This is being tracked in [#112](https://github.com/k8ssandra/k8ssandra-operator/issues/112).
```
olim7t, line 104: That issue is closed, as well as a few follow-ups (#446, #463, #515?) but I haven't found a good summary on how to run against other clusters.

docs/developer/README.md
```diff
@@ -99,65 +99,72 @@ Integration tests use the `envtest` package from controller-runtime.  See this [
...
-If you want to run a single test, set the `E2E_TEST` variable as follows:
+```shell
+# - Create kind clusters (deleting any existing ones first)
+# - Install Cert Manager and Traefik (used for ingress in e2e tests)
+make multi-create
 
+# - Build the project
+# - Generate the CRDs, then load them into the Kind cluster
+# - Build the operator image, then load it into the Kind cluster
+make multi-prepare
```
adejanovski, line 128: Running `multi-up` is enough to create the kind clusters and load the image.

docs/developer/README.md
```diff
@@ -99,65 +99,72 @@ Integration tests use the `envtest` package from controller-runtime.  See this [
...
-```
-
-### Manual procedure 
 
-#### Create the kind clusters
-The multi-cluster tests require two clusters.
+You can then run the entire test suite:
 
+```shell
+make e2e-test
```
adejanovski, line 134: I seem to remember that we don't clean up the cluster in a way that allows running multiple e2e tests back to back 🤔 
They would probably take something like 10 hours anyway and fry any laptop 😅 
Maybe we should make it clear that this target shouldn't be used without specifying a test. Maybe we could even modify the makefile to make that mandatory. wdyt?

docs/developer/README.md
adejanovski: then, if you want to rebuild and reload the operator once the cluster is up, you can invoke `multi-reload`.

docs/developer/README.md
```diff
@@ -99,65 +99,72 @@ Integration tests use the `envtest` package from controller-runtime.  See this [
...
-```
+### Single-cluster tests
 
-#### Run the tests
+Same principle, but the targets have different names:
 
-`make e2e-test` runs the tests under `test/e2e`.
+```shell
+make single-create
+make single-prepare
```
adejanovski, line 156: You should replace these by `make single-up`.
It's actually an anomaly that we're not doing it in the makefile and still rely on multiple targets.

adejanovski (approved):
Thanks for updating the docs @olim7t !
I've added a few comments around which makefile targets to use in order to simplify the commands.

docs/developer/README.md
adutra: I created `single-create` and `single-prepare`. 

I actually don't think that `single-up` is very developer friendly as it destroys and re-creates the cluster each time. If, as a developer, you only need to "refresh" the cluster by re-building the operator image and re-deploying it, that's overkill. That's where you can use just `single-prepare` it's much faster.

adutra: ... and `single-reload` is a bit overkill as well: granted, it doesn't re-create the whole cluster; but it deploys the operator images in the default `k8ssandra-operator` namespace, which is unnecessary when running e2e tests.

docs/developer/README.md
jsanda: By default we call the [cleanup](https://github.com/k8ssandra/k8ssandra-operator/blob/main/test/e2e/suite_test.go#L636) function which deletes the K8ssandraCluster, so it should allow for running tests back to back, in theory. I don't know how practical it is though because of the time it would take.

olim7t: Right, maybe we shouldn't mention this at all. Even the github actions run the single-test variant for every test, not the whole suite at once.

docs/developer/README.md
adejanovski: kk, we can discard my comment if you prefer to use those two commands. I think the overhead of `-reload` is low and `-up` is convenient to (re)start from scratch, but it's not a hill to die on 👍

docs/developer/README.md
adejanovski: 👍

docs/developer/README.md
adejanovski: @adutra, are there docs somewhere that @olim7t could refer to here?

olim7t: OK, this is ready to merge then.

PR #685, opened 2022-09-15
Support for multi DC with DSE

### LINKED ISSUE #695: K8SSAND-1797 ⁃ Support multi DC for DSE
DSE accepts the jvm option which sets the system ks replication to span over multiple DCs before they exist, but it will fail applying that replication when the first DC starts.
We need to avoid setting it for DSE clusters and instead rely on the alter+rebuild operations that are triggered during expansions to new DCs.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1797) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1797
┆priority: Medium

### LINKED ISSUE #696: K8SSAND-1798 ⁃ DCs get upgraded concurrently instead of sequentially
A race condition between cass-operator and k8ssandra-operator makes the latter believe a DC is ready although it has a pending upgrade.
This results in DC upgrades happening in all DCs concurrently instead of a single DC at a time.

We need to make sure DCs have finished their upgrades before moving on to the next, by comparing `meta.Generation` (which gets incremented when the spec is updated) with `.status.ObservedGeneration` (which gets updated when the DC completed the upgrade process).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1798) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1798
┆priority: Medium

adejanovski:
**What this PR does**:
Fix issues preventing multi DC setups with DSE by altering the system keyspaces when a new DC is added, even if the cluster isn't fully initialized yet. Previously we relied on a java option to pass the projected replication for theses keyspaces including non existing ones. This works with Cassandra but fails with DSE.
This PR performs "classic" DC expansion even on cluster initialization, using alter and rebuild.

This PR also fixes an issue with upgrades happening concurrently on all DCs, instead of sequentially. This is done by requeuing the reconcile as long as `meta.Generation`  is different than `status.ObservedGeneration`.

It also changes the format of the time in the operator logs from a timestamp to a human readable date and time.

Fixes #695
Fixes #696

**Checklist**
- [x] Changes manually tested


test/e2e/suite_test.go
```diff
@@ -941,6 +949,48 @@ func createMultiDatacenterCluster(t *testing.T, ctx context.Context, namespace s
...
+	dc1Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
+	checkDatacenterReady(t, ctx, dc1Key, f)
+	assertCassandraDatacenterK8cStatusReady(ctx, t, f, kcKey, dc1Key.Name, dc2Key.Name, dc3Key.Name)
+
+	t.Log("retrieve database credentials")
+	username, password, err := f.RetrieveDatabaseCredentials(ctx, f.DataPlaneContexts[0], namespace, k8ssandra.SanitizedName())
+	require.NoError(err, "failed to retrieve database credentials")
+
+	t.Log("check that nodes in dc1 see nodes in dc2 and dc3")
+	pod := DcPrefix(t, f, dc1Key) + "-rack1-sts-0"
```
jsanda, line 981: The rack name is wrong since you haven't declared any racks in your spec.

test/e2e/suite_test.go
```diff
@@ -941,6 +949,48 @@ func createMultiDatacenterCluster(t *testing.T, ctx context.Context, namespace s
...
+	count := 3
+	checkNodeToolStatus(t, f, f.DataPlaneContexts[0], namespace, pod, count, 0, "-u", username, "-pw", password)
+
+	assert.NoError(t, err, "timed out waiting for nodetool status check against "+pod)
+
+	t.Log("check nodes in dc2 see nodes in dc1 and dc3")
+	pod = DcPrefix(t, f, dc2Key) + "-rack1-sts-0"
```
jsanda, line 988: The rack name is wrong since you haven't declared any racks in your spec.

test/e2e/suite_test.go
```diff
@@ -1109,6 +1159,142 @@ func addDcToCluster(t *testing.T, ctx context.Context, namespace string, f *fram
...
+	assert.NoError(err, "timed out waiting for nodetool status check against "+pod)
+
+	t.Log("check nodes in dc2 see nodes in dc1")
+	pod = DcPrefix(t, f, dc2Key) + "-default-sts-0"
+	checkNodeToolStatus(t, f, f.DataPlaneContexts[1], namespace, pod, count, 0, "-u", username, "-pw", password)
+
+	assert.NoError(err, "timed out waiting for nodetool status check against "+pod)
+
+	keyspaces := []string{"system_auth", stargate.AuthKeyspace, "reaper_db"}
+	for _, ks := range keyspaces {
```
jsanda, line 1258: You can check that the expected CassandraTask is successfully executed. It would be good to also add/update integration tests.

test/testdata/fixtures/multi-dc-dse/k8ssandra.yaml
```diff
@@ -0,0 +1,42 @@
...
+        resources:
+          requests:
+            storage: 5Gi
+    config:
+      jvmOptions:
+        heapSize: 512Mi
+        heapNewGenSize: 256Mi
+        gc: CMS
+    mgmtAPIHeap: 64Mi
+    datacenters:
```
jsanda, line 23: I ran into a problem with having multiple DCs in the same namespace. Cleanup won't happen properly. One of the CassandraDatacenters does not get deleted. The problem stems from the fact all 3 DCs share the same superuser secret which gets deleted some where along the way. I think it might be getting deleted by the replicated secret controller, but I need to verify. 

The secret is deleted before the DC which results in this validation error reported by cass-operator:

`1.6632922466632967e+09	ERROR	controllers.CassandraDatacenter	CassandraDatacenter resource is invalid	{"cassandradatacenter": "multi-dc-dse-8rf4pf/dc3", "requestNamespace": "multi-dc-dse-8rf4pf", "requestName": "dc3", "loopID": "b0473739-0c15-4650-9332-6004e74f1786", "error": "Could not load superuser secret for CassandraCluster: multi-dc-dse-8rf4pf/test-superuser"}`

I tried provided my own superuser secret. Interestingly, it too was deleted. That feels like a bug to me. I don't think the operator should be deleting secrets that it doesn't create. This should be addressed separately.

To avoid the cleanup problem, I would deploy the DCs in separate namespaces.

jsanda: It is in fact the secret controller. The K8ssandraCluster controller adds the `replicatedresource.k8ssandra.io/orphan: "true"` annotation to my superuser secret. The secrets controller however, only looks for that annotation on the ReplicatedSecret. An easy work around would be to add the annotation to the ReplicatedSecret. I tested and verified that this works.

adejanovski: Using multiple namespaces would probably be a big change due to how our e2e test framework is built.
I've collocated them in a single cluster/namespace because we had a bug in the image, I'll split them again across multiple clusters now.

test/e2e/suite_test.go
adejanovski: right 👌

test/e2e/suite_test.go
adejanovski: 👌

test/e2e/suite_test.go
adejanovski: yes, definitely worth checking the CassandraTask ran successfully.

controllers/k8ssandra/datacenters.go
```diff
@@ -475,3 +459,13 @@ func dcUpgradePriority(dc api.CassandraDatacenterTemplate) int {
 
 	return 2
 }
+
+func (r *K8ssandraClusterReconciler) removeRebuildDcAnnotation(kc *api.K8ssandraCluster, ctx context.Context) result.ReconcileResult {
```
adejanovski, line 463: This refactoring is probably unnecessary, since it was done when I thought I'd need to remove the annotation for another feature. It was not the case in the end.
It's a valuable refactoring though to reduce the cognitive complexity of the `reconcileDcRebuild()` function.

main.go
```diff
@@ -93,6 +94,7 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	opts := zap.Options{
 		Development: true,
+		TimeEncoder: zapcore.ISO8601TimeEncoder,
```
adejanovski, line 97: Changes the timestamp to an actual date time in the operator logs. Vastly improves readability :)

controllers/k8ssandra/datacenters.go
```diff
@@ -204,6 +208,12 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 				}
 			}
 
+			// DC is in the process of being upgraded but hasn't completed yet. Let's wait for it to go through.
+			if actualDc.GetGeneration() != actualDc.Status.ObservedGeneration {
+				logger.Info("CassandraDatacenter is being upgraded. Requeuing the reconcile.", "Generation", actualDc.GetGeneration(), "ObservedGeneration", actualDc.Status.ObservedGeneration)
```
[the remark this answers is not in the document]
jsanda: I just deployed a K8ssandraCluster with a single node. I see the log message 4 times. More important than the log message, we are changing behavior for scenarios other than upgrades. Maybe that is fine, but it isn't clear.

controllers/k8ssandra/datacenters.go
```diff
@@ -273,23 +281,14 @@ func (r *K8ssandraClusterReconciler) setStatusForDatacenter(kc *api.K8ssandraClu
...
-				Reaper:    kdcStatus.Reaper.DeepCopy(),
-			}
-		}
-	}
-}
-
 func datacenterAddedToExistingCluster(kc *api.K8ssandraCluster, dcName string) bool {
 	_, found := kc.Status.Datacenters[dcName]
-	return kc.Status.GetConditionStatus(api.CassandraInitialized) == corev1.ConditionTrue && !found
+	if kc.Spec.Cassandra.ServerType == api.DistributionDse {
```
jsanda, line 286: Do we have a different logic for DSE here since we cannot apply system replication settings like we do for Cassandra?

test/framework/framework.go
```diff
@@ -286,9 +288,28 @@ func (f *Framework) SetDatacenterStatusReady(ctx context.Context, key ClusterKey
...
 
+// UpdateDatacenterGeneration fetches the CassandraDatacenter specified by key and persists
+// a status update to make the ObservedGeneration match the Generation.
+// It will wait for .meta.Generation to be different from .status.ObservedGeneration.
+// This is done to simulate the behavior of the CassandraDatacenter controller.
+func (f *Framework) UpdateDatacenterGeneration(ctx context.Context, t *testing.T, key ClusterKey) bool {
+	dc := &cassdcapi.CassandraDatacenter{}
+	err := f.Get(ctx, key, dc)
+	require.NoError(t, err, "failed to get CassandraDatacenter %s/%s", key.Namespace, key.Name)
+	if dc.Generation == dc.Status.ObservedGeneration {
```
jsanda, line 303: This `if` block is unnecessary. The CassandraDatacenter controller isn't running, so nothing (other than tests) should be updating `.Status.ObservedGeneration`. I first added a log statement at line 304 and ran the tests. As expected the log statement was never executed. Then I commented out the block and the tests still passed. Lastly, there isn't any need to wrap the method call in an `Eventually` block.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
adejanovski: The term "upgrade" in this logging isn't about a change in the image we use for Cassandra, but rather the upgrade of the cassdc object itself, which is going from one generation to the other.
Maybe "updated" would be less misleading? I reckon I was not sure which term would be more appropriate here.

controllers/k8ssandra/datacenters.go
adejanovski: Yes. For Cassandra, we'll only perform alter+rebuild in the case where we're not initializing the cluster, even if there are multiple DCs being created.
For DSE we need to perform these operations for all DCs except the first one, even if the cluster isn't fully initialized.

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
jsanda: Updated would be better 🙂

test/framework/framework.go
adejanovski: This check has been useful to ensure we're performing the update only if it is necessary. It prevented me from putting generation updates where they didn't belong. That allows making sure we're not complexifying our tests with additional code lines that aren't required.

controllers/k8ssandra/add_dc_test.go
```diff
@@ -90,12 +90,16 @@ func addDcSetupForSingleDc(ctx context.Context, t *testing.T, f *framework.Frame
 	err := f.Get(ctx, dc1Key, dc)
 	require.NoError(err)
 
-	err = f.SetDatacenterStatusReady(ctx, dc1Key)
-	require.NoError(err, "failed to set dc1 status ready")
+	f.SetDatacenterStatusReady(ctx, dc1Key)
```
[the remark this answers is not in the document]
adejanovski: 👍

controllers/k8ssandra/datacenters.go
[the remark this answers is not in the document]
adejanovski: sounds good

controllers/k8ssandra/schemas.go
```diff
@@ -144,7 +144,13 @@ func (r *K8ssandraClusterReconciler) updateReplicationOfSystemKeyspaces(
 
 	logger.Info("Preparing to update replication for system keyspaces", "replication", replication)
 
-	for _, ks := range api.SystemKeyspaces {
+	systemKeyspaces := append([]string{}, api.SystemKeyspaces...)
```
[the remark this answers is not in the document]
adejanovski: 👍

test/e2e/suite_test.go
```diff
@@ -943,6 +947,45 @@ func createMultiDatacenterCluster(t *testing.T, ctx context.Context, namespace s
 	assert.NoError(t, err, "timed out waiting for nodetool status check against "+pod)
 }
 
+// createMultiDatacenterDseCluster creates a K8ssandraCluster with two CassandraDatacenters,
+// one running locally and the other running in a remote cluster.
+// Each CassandraDatacenter is configured with a different DSE workload, with the second DC being
+// processed first due to the priority of its workload.
+func createMultiDatacenterDseCluster(t *testing.T, ctx context.Context, namespace string, f *framework.E2eFramework) {
```
adutra, line 954: Don't you think we should move this test to dse_test.go? This file is already very large :-)

controllers/k8ssandra/stop_dc_test.go
```diff
@@ -287,7 +289,13 @@ func addAndStopDc(t *testing.T, f *framework.Framework, ctx context.Context, kc
 
 	t.Log("check that dc3 was rebuilt")
 	verifyRebuildTaskCreated(ctx, t, f, dc3Key, dc1Key)
-	rebuildTaskKey := framework.NewClusterKey(f.DataPlaneContexts[2], kc.Namespace, "dc3-rebuild")
+	rebuildTaskKey := framework.ClusterKey{
```
adutra, line 292: I'm not sure this change is needed.

controllers/k8ssandra/stop_dc_test.go
```diff
@@ -397,6 +405,17 @@ func addAndStopDc(t *testing.T, f *framework.Framework, ctx context.Context, kc
 	t.Log("check that reaper reaper2 is still present")
 	require.Eventually(t, f.ReaperExists(ctx, reaper2Key), timeout, interval, "failed to verify reaper reaper2 created")
 
+	t.Log("check that dc3 was rebuilt")
+	verifyRebuildTaskCreated(ctx, t, f, dc3Key, dc1Key)
+	rebuildTaskKey = framework.ClusterKey{
```
adutra, line 410: Consider using `framework.NewClusterKey` for brevity.

controllers/k8ssandra/remove_dc_test.go
```diff
@@ -142,6 +145,16 @@ func deleteDcWithStargateAndReaper(ctx context.Context, t *testing.T, f *framewo
 		},
 	}
 
+	require.Eventually(func() bool {
```
adutra, line 148: Do we need Eventually blocks here and below? No other controller is running in this test than k8ssandra controller, so the update of the dc status should never fail, right?

controllers/k8ssandra/reaper_test.go
```diff
@@ -225,6 +225,14 @@ func createMultiDcClusterWithReaper(t *testing.T, ctx context.Context, f *framew
 	err = f.Client.Patch(ctx, kc, patch)
 	require.NoError(err, "failed to update K8ssandraCluster")
 
+	require.Eventually(func() bool {
```
adutra, line 228: Same here, I'm wondering if we need those Eventually blocks.

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -150,6 +150,10 @@ func createMultiDcClusterWithMedusa(t *testing.T, ctx context.Context, f *framew
 	})
 	require.NoError(err, "failed to update dc1 status to ready")
 
+	require.Eventually(func() bool {
```
adutra, line 153: And here.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -261,15 +261,19 @@ func createSingleDcCluster(t *testing.T, ctx context.Context, f *framework.Frame
 	if err := f.Patch(ctx, kc, kcPatch, kcKey); err != nil {
 		assert.Fail(t, "got error patching for telemetry", "error", err)
 	}
-	if err = f.SetDatacenterStatusReady(ctx, framework.ClusterKey{
+
+	dc1Key := framework.ClusterKey{
```
adutra, line 265: Consider using `framework.NewClusterKey`.

controllers/k8ssandra/add_dc_test.go
```diff
@@ -534,6 +550,12 @@ func schemaDisagreementOnStargate(ctx context.Context, t *testing.T, f *framewor
 		},
 	}
 
+	dc1Key := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: kc.Namespace, Name: "dc1"}, K8sContext: f.DataPlaneContexts[0]}
```
adutra, line 553: Consider using `framework.NewClusterKey`.

controllers/k8ssandra/add_dc_test.go
```diff
@@ -411,6 +423,8 @@ func withStargateAndReaper(ctx context.Context, t *testing.T, f *framework.Frame
 
 	addStargateAndReaperToCluster(ctx, t, f, kc)
 
+	dc1Key := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: kc.Namespace, Name: "dc1"}, K8sContext: f.DataPlaneContexts[0]}
```
adutra, line 426: Same here.

adutra (approved):

apis/k8ssandra/v1alpha1/constants.go
```diff
@@ -59,4 +59,5 @@ const (
 
 var (
 	SystemKeyspaces = []string{"system_traces", "system_distributed", "system_auth"}
+	DseKeyspaces    = []string{"dse_leases", "dse_perf", "dse_security"}
```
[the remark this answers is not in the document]
adejanovski: I've thought about this but there are such keyspaces that have exotic strategies that we cannot change such as `LocalStrategy` or `EverywhereStrategy`.

test/e2e/suite_test.go
adejanovski: oh yeah, totally right.
I'll move it there.

controllers/k8ssandra/add_dc_test.go
adejanovski: Cool trick, thanks!

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: 👍

controllers/k8ssandra/remove_dc_test.go
adejanovski: I've tried removing them, and the test fails. There's a delay between the patching/creation of the kc objects and their processing by the controller, where we can end up with the test code going too fast. It's also why I added a check on whether or not the generations are different before updating the observed generation.

controllers/k8ssandra/stop_dc_test.go
adejanovski: no it's not. Not sure why I did that 🤔

PR #689, opened 2022-09-17
Switch to NGINX for ingress in e2e tests

olim7t:
**What this PR does**:
Replace Traefik with NGINX for ingress in e2e tests. This simplifies ingress rules, and will allow adding gRPC ingress in #679.

N/A

**Checklist**
- [x] Changes manually tested


test/framework/ingress.go
```diff
@@ -23,238 +24,165 @@ func (s HostAndPort) Port() string {
...
-	require.NoError(t, err)
-	err = generateStargateIngressKustomization(k8sContext, namespace, stargateServiceName, stargateRestHostAndPort.Host())
-	require.NoError(t, err)
-	err = f.kustomizeAndApply(dest, namespace, k8sContext)
-	require.NoError(t, err)
+	f.deployIngress(t, k8sContext, namespace, "stargate-ingress.yaml", "stargate", stargateTemplate,
+		&ingressKustomization{stargateServiceName, stargateRestHostAndPort.Host()})
+
+	f.deployIngress(t, k8sContext, "ingress-nginx", "stargate-cql-ingress.yaml", "stargate-cql", stargateCqlTemplate,
+		&cqlIngressKustomization{namespace, stargateServiceName, stargateRestHostAndPort.Host()})
```
olim7t, line 43: I used the opportunity to eliminate a bit of duplication here, all the function do essentially the same thing.

test/framework/ingress.go
```diff
@@ -23,238 +24,165 @@ func (s HostAndPort) Port() string {
...
+      value: "{{ .Host }}"
+    - op: replace
+      path: /spec/rules/0/http/paths/0/backend/service/name
+      value: "{{ .ServiceName }}"
+    - op: replace
+      path: /spec/rules/0/http/paths/1/backend/service/name
+      value: "{{ .ServiceName }}"
+    - op: replace
+      path: /spec/rules/0/http/paths/2/backend/service/name
+      value: "{{ .ServiceName }}"
```
olim7t, line 105: I haven't looked in detail at what `patchesJson6902`, but with this version it's not needed, all the operations can be under `patch`.

test/framework/ingress.go
```diff
@@ -23,238 +24,165 @@ func (s HostAndPort) Port() string {
...
-	Host        string
-}
-
-func generateStargateIngressKustomization(k8sContext, namespace, serviceName, host string) error {
-	tmpl := `apiVersion: kustomize.config.k8s.io/v1beta1
+const stargateTemplate = `apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - stargate-ingress.yaml
-namespace: {{ .Namespace }}
```
olim7t, line 89: I removed the namespace from the YAML templates.
- it's one less thing to handle in the Kustomization
- it avoids the possible confusion for CQL, where there are two different ones: ConfigMap namespace and Service namespace.
- the one eventually passed to `kubectl apply` is the only one that matters anyway (the command would fail if it didn't match).

olim7t (commented):
I see an occasional error in the logs:
```
7:49AM ERR CQL client conn [L:127.0.0.1:51269 <-> R:127.0.0.1:30942]: handshake failed error="CQL client conn [L:127.0.0.1:51269 <-> R:127.0.0.1:30942]: failed to retrieve incoming frame: CQL client conn [L:127.0.0.1:51269 <-> R:127.0.0.1:30942]: [in-flight handler]: handler closed"
```
But it doesn't seem to affect test outcome.

test/framework/ingress.go
burmanm: Kustomize does not care about ``--namespace``  given to kubectl, only the templates and what you have there matters. The value from kubectl does not swim to the template rendering (and without namespace in templates, the namespace transformers might not work correctly)

olim7t: What I was aiming for is no namespace transformation at all. In [this code snippet]( https://github.com/k8ssandra/k8ssandra-operator/blob/main/test/framework/e2e_framework.go#L273-L279):
```go
	buf, err := kustomize.BuildDir(kdir)
        [...]
	options := kubectl.Options{Context: context, ServerSide: true, Namespace: namespace}
	if err := kubectl.Apply(options, buf); err != nil {
		return err
	}
```
`kustomize.BuildDir` takes a template that has no namespace (e.g. `stargate-ingress.yaml`), and transforms it into a YAML that has no namespace either. `kubectl.Apply` will create the resource in the correct namespace anyway, so do we need it in the generated YAML file?

test/framework/ingress.go
jsanda: `patch` is newer and actually still applies the Json6902 style patch. For reference:

* https://github.com/kubernetes-sigs/kustomize/blob/master/examples/jsonpatch.md
https://stackoverflow.com/questions/63604579/what-is-the-difference-between-patches-vs-patchesjson6902-in-kustomize#63605421

jsanda (approved):
Nice work!

PR #694, opened 2022-09-21
Upgrade to Go 1.18 and to operator-sdk 1.23 (fixes #681)

### LINKED ISSUE #681: K8SSAND-1779 ⁃ Upgrade to Go 1.18 and to operator-sdk 1.23
**What is missing?**
We are still on Go 1.17.  Our Kubernetes dependencies are on 1.23.4. Based on [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/Makefile#L347) in the Makefile it looks like we are on operator-sdk v1.18.0. Going through the [upgrade guide](https://sdk.operatorframework.io/docs/upgrading-sdk-version/) up through the latest release, which is v1.23.0, will move us to Go 1.18, update Kubernetes dependencies to 1.24, and update tooling dependencies like controller-gen.

For reference these changes have already been made in cass-operator in https://github.com/k8ssandra/cass-operator/pull/399.

@burmanm A couple questions for you:

* In https://github.com/k8ssandra/cass-operator/issues/398 you mentioned upgrading to Kustomize 4.5.7 for performance improvement and the ability to use newer namespace transformer to remove some hack logic. Can you provide any references for additional context?
* https://github.com/k8ssandra/cass-operator/issues/397 removed the dependency on k8s.io/kubernetes. You mentioned that it was recommended by Kubernetes devs. Can you provide any references for additional context?

I believe operator-sdk v1.23.0 moves to Kustomize v4.5.5. Let's go ahead and move to v4.5.7 to be consistent with cass-operator.

**Environment**

* K8ssandra Operator version:

    `v1.2.0`

olim7t:
**What this PR does**:
Update to Go 1.18, Kubernetes 1.24, Operator SDK 1.23, controller-gen 0.9.2, Kustomize 4.5.7, controller-runtime 0.12.2.

Fixes #681

**Checklist**
- [x] Changes manually tested


go.mod
```diff
@@ -1,22 +1,22 @@
...
 
-go 1.17
+go 1.18
 
 require (
 	github.com/Jeffail/gabs v1.4.0
 	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/apache/tinkerpop/gremlin-go v0.0.0-20220530191148-29272fa563ec
 	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/cespare/xxhash/v2 v2.1.2 // indirect
```
olim7t, line 9: I moved all indirect dependencies to the second require block below.

go.mod
```diff
@@ -26,101 +26,105 @@ require (
...
-	k8s.io/api v0.23.4
-	k8s.io/apimachinery v0.23.4
-	k8s.io/client-go v0.23.4
-	k8s.io/kubernetes v1.23.4
-	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704
-	sigs.k8s.io/controller-runtime v0.11.1
-)
-
-require (
-	github.com/apache/tinkerpop/gremlin-go v0.0.0-20220530191148-29272fa563ec
```
olim7t, line 38: Not sure why there was a dedicated require block for this, I moved it to the first one above.

go.mod
```diff
@@ -26,101 +26,105 @@ require (
...
+	k8s.io/apimachinery v0.24.2
+	k8s.io/client-go v0.24.2
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
+	sigs.k8s.io/controller-runtime v0.12.2
 	sigs.k8s.io/yaml v1.3.0
 )
 
 require (
-	cloud.google.com/go v0.93.3 // indirect
+	cloud.google.com/go v0.81.0 // indirect
```
olim7t, line 38: My assumption is that indirect dependencies are managed by `go mod`, not edited by us manually.
So after updating direct dependencies, I deleted all the indirect ones and re-ran `go mod tidy` to restart from a clean slate.
But this results in a few lower versions like this one, I'm not sure why.

I can revisit this if this wasn't the correct approach.

olim7t (commented):
I've kept separate commits so far for clarity, but I will squash everything together when merging.

CHANGELOG/CHANGELOG-1.2.md
```diff
@@ -13,6 +13,11 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
+## V1.2.2
+
+* [ENHANCEMENT] [#681](https://github.com/k8ssandra/k8ssandra-operator/issues/681) Update to Go 1.18, Kubernetes 1.24,
+  Operator SDK 1.23, controller-gen 0.9.2, Kustomize 4.5.7, controller-runtime 0.12.2
```
burmanm, line 19: Nit, but lets keep these one line (instead of extra forced line changes). Helps rendering.

docs/developer/README.md
```diff
@@ -1,5 +1,5 @@
 # Requirements
-* Go >= 1.17
+* Go >= 1.18
 * kubectl >= 1.17
```
burmanm, line 3: These other requirements are wrong since we just updated our tools. We don't know if things work with kubectl 1.17 (well, they won't) since we're on 1.24. And same for kustomize 4.0.5, which I don't think works.

Makefile
```diff
@@ -322,17 +353,21 @@ create-clientconfig:
 	done
 
 
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
-controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
+.PHONY: kustomize
```
burmanm, line 357: .PHONY should be in every Makefile target or none I guess. Now the behavior changes per target. I prefer every target, since we're not dealing with files (other than the $(LOCALBIN) target)

go.mod
burmanm: While this may look prettier, sadly it's pretty futile. Next time someone runs ``go mod tidy`` , it will mess things up. Or ``go get -d``.

go.mod
burmanm: Go does not do single block, it can do one per dependency run also.

go.mod
burmanm: The ``go mod`` commands are not "stable" in a sense that they would generate same output each time. I'm assuming they use map[] underneath for dependency graphs and resolve, in which case the iteration order is random and as such can generate different output.

Makefile
```diff
@@ -38,6 +38,17 @@ IMAGE_TAG_BASE ?= k8ssandra/k8ssandra-operator
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
+# BUNDLE_GEN_FLAGS are the flags passed to the operator-sdk generate bundle command
+BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
```
burmanm, line 42: This code isn't used anywhere.

Makefile
```diff
@@ -38,6 +38,17 @@ IMAGE_TAG_BASE ?= k8ssandra/k8ssandra-operator
...
+
+# USE_IMAGE_DIGESTS defines if images are resolved via tags or digests
+# You can enable this value if you would like to use SHA Based Digests
+# To enable set flag to true
+USE_IMAGE_DIGESTS ?= false
+ifeq ($(USE_IMAGE_DIGESTS), true)
+	BUNDLE_GEN_FLAGS += --use-image-digests
```
burmanm, line 49: Same here, this code isn't used anywhere.

burmanm (commented):
kuttl wasn't updated, the first Go 1.18 version is 0.12.1, but 0.13.0 is maybe better.

docs/developer/README.md
olim7t: Right, I didn't pay attention to this (I have more recent versions installed anyway).

I will set:
- kubectl to 1.23 (based on the [version skew policy docs](https://kubernetes.io/releases/version-skew-policy/#kubectl))
- Kustomize to 4.5.7. Maybe older versions work, but I don't think trying them is worth the time. Also it's the current version installed by package managers (brew, etc.)
- Kind to 0.15.0, for the same reasons.

I'll also update `default_kind_node_version` in `setup-kind-multicluster.sh`.

Makefile
olim7t: Right, I need to reference the new variable when calling `generate bundle` in the `bundle` target.
Btw this also need to be done in cass-operator.

go.mod
olim7t: Yeah, maybe we can just clean things up manually from time to time...

go.mod
olim7t: So do you think the discrepancy matters here? It looks like if the version is higher than what `tidy` would suggest, it's not modified. I can go back to the versions from `main` and re-run it.

olim7t: If I do that I get those differences:
```diff
-       cloud.google.com/go v0.81.0 // indirect
+       cloud.google.com/go v0.93.3 // indirect

-       github.com/Azure/go-autorest/autorest v0.11.18 // indirect
-       github.com/Azure/go-autorest/autorest/adal v0.9.13 // indirect
+       github.com/Azure/go-autorest/autorest v0.11.20 // indirect
+       github.com/Azure/go-autorest/autorest/adal v0.9.15 // indirect

-       github.com/form3tech-oss/jwt-go v3.2.3+incompatible // indirect

+       github.com/golang-jwt/jwt/v4 v4.0.0 // indirect

+       github.com/golang/snappy v0.0.4 // indirect

-       go.uber.org/atomic v1.7.0 // indirect
+       go.uber.org/atomic v1.9.0 // indirect
```

Makefile
olim7t: Yeah this was already inconsistent before it seems.
I can do it here, or on a separate PR if we don't want to mix the changes.

burmanm:
Why did you revert kind image? Now our envtest is 1.24.2, but we run manually on 1.22? That could lead to interesting cases.

go.mod
burmanm: I don't think it matters which one it picks as long as it compiles afterwards. Clearly there's two dependencies which have different version of certain module, so picking either one could break the other.

Makefile
burmanm: Separate PR is fine.

Makefile
burmanm: Yep, though it's not high severity (k8ssandra-operator does not use bundles and cass-operator uses separate script when digests are used)

olim7t:
All the e2e tests are failing with the 1.24 image. I'm going to debug locally.

config/rbac/service_account_token.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: k8ssandra-operator-token
+  annotations:
+    kubernetes.io/service-account.name: "k8ssandra-operator"
+type: kubernetes.io/service-account-token
```
olim7t, line 7: This solves the issue with Kind 1.24+ images. From the [Kubernetes 1.24 upgrade notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes):
> - The `LegacyServiceAccountTokenNoAutoGeneration` feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets).

I used the second option which is functionally equivalent to what we had before. But maybe we should investigate how `TokenRequest` could be used.

burmanm: Excellent. Yes, we should indeed move to a newer API at some point (I didn't check the version ranges - but those are always problematic in Kubernetes). If you wish to create tickets, that would be nice.

Makefile
olim7t: #697

olim7t:
Current test failures:
- `Run multi-cluster e2e tests / CheckStargateApisWithMultiDcCluster`: looks like an unrelated timeout, and it passes locally.
- `Run e2e tests with isolated control plane / CreateMultiReaper`: logs have expired.

I'm going to re-run them.

jsanda (approved):

PR #700, opened 2022-09-27
Remove dc

### LINKED ISSUE #699: K8SSAND-1806 ⁃ User keyspaces shouldn't be altered automatically during a DC decommission
Removing a DC from the K8ssandraCluster spec currently alters all keyspaces (system and user) to remove replicas from that DC, and then deletes the CassandraDatacenter object.
This can create bad surprises in multi tenant clusters where a team could delete a DC without asking all tenants if they're ok with it.
We want to add safety and prevent DC decommissions if any user keyspace still has replicas on it, informing the operators of the nature of the issue.
Since network calls are involved and due to the nature of multi cluster deployments, we cannot rely on webhooks for this and have to accept the patches. Instead, we'll create events to notify ops of the problem, and potentially enrich the K8ssandraCluster object status with the reconcile error message.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1806) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1806
┆priority: Medium

adejanovski:
**What this PR does**:
Prevents DC decommission if user keyspaces are still replicated to it.
This PR also adds event creation at the end of a reconcile. The event will contain the error message if the reconcile failed.
Additionally, the error message will be stored in a new `.spec.error` field which is displayed when running `kubectl get k8c` commands, thanks to the kubebuilder printcolumn annotation.

Fixes #699 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -85,6 +85,9 @@ type K8ssandraClusterStatus struct {
 	//
 	// TODO Figure out how to inline this field
 	Datacenters map[string]K8ssandraStatus `json:"datacenters,omitempty"`
+
+	// +kubebuilder:default=None
+	Error string `json:"error,omitempty"`
 }
 
 type K8ssandraClusterConditionType string
@@ -123,6 +126,7 @@ type K8ssandraStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=k8ssandraclusters,shortName=k8c;k8cs
+// +kubebuilder:printcolumn:name="Error",type=string,JSONPath=".status.error",description="Latest reconcile error"
 
 // K8ssandraCluster is the Schema for the k8ssandraclusters API. The K8ssandraCluster CRD name is also the name of the
 // Cassandra cluster (which corresponds to cluster_name in cassandra.yaml).
--- a/controllers/k8ssandra/datacenters.go
+++ b/controllers/k8ssandra/datacenters.go
@@ -461,7 +461,7 @@ func dcUpgradePriority(dc api.CassandraDatacenterTemplate) int {
 }
 
 func (r *K8ssandraClusterReconciler) removeRebuildDcAnnotation(kc *api.K8ssandraCluster, ctx context.Context) result.ReconcileResult {
-	patch := client.MergeFromWithOptions(kc.DeepCopy())
+	patch := client.MergeFrom(kc.DeepCopy())
 	delete(kc.Annotations, api.RebuildDcAnnotation)
 	if err := r.Client.Patch(ctx, kc, patch); err != nil {
 		err = fmt.Errorf("failed to remove %s annotation: %v", api.RebuildDcAnnotation, err)
--- a/controllers/k8ssandra/k8ssandracluster_controller.go
+++ b/controllers/k8ssandra/k8ssandracluster_controller.go
@@ -18,6 +18,7 @@ package k8ssandra
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
@@ -33,6 +34,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/kubernetes/pkg/apis/core"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -55,6 +58,7 @@ type K8ssandraClusterReconciler struct {
 	Scheme        *runtime.Scheme
 	ClientCache   *clientcache.ClientCache
 	ManagementApi cassandra.ManagementApiFactory
+	Recorder      record.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=k8ssandra.io,namespace="k8ssandra",resources=k8ssandraclusters;clientconfigs,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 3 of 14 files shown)


adejanovski:
Only consider the last commit. The others are inherited from other PRs which are pending merge.

olim7t (approved):
Approving with the disclaimer that I'm still very new to the project.

controllers/k8ssandra/datacenters.go
```diff
@@ -204,6 +208,12 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
 				}
 			}
 
+			// DC is in the process of being upgraded but hasn't completed yet. Let's wait for it to go through.
+			if actualDc.GetGeneration() != actualDc.Status.ObservedGeneration {
+				logger.Info("CassandraDatacenter is being updated. Requeuing the reconcile.", "Generation", actualDc.GetGeneration(), "ObservedGeneration", actualDc.Status.ObservedGeneration)
```
burmanm, line 213: Do we need this as logger.Info, or logger.V(1).Info (debug) ?

controllers/k8ssandra/datacenters.go
```diff
@@ -466,3 +459,13 @@ func dcUpgradePriority(dc api.CassandraDatacenterTemplate) int {
 
 	return 2
 }
+
+func (r *K8ssandraClusterReconciler) removeRebuildDcAnnotation(kc *api.K8ssandraCluster, ctx context.Context) result.ReconcileResult {
+	patch := client.MergeFromWithOptions(kc.DeepCopy())
```
burmanm, line 464: MergeFromWithOptions, but no options? Is this meant to be optimistic lock patch?

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -86,6 +91,13 @@ func (r *K8ssandraClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	patch := client.MergeFromWithOptions(kc.DeepCopy())
 	result, err := r.reconcile(ctx, kc, logger)
 	if kc.GetDeletionTimestamp() == nil {
+		if err != nil {
+			kc.Status.Error = err.Error()
+			r.Recorder.Event(kc, core.EventTypeWarning, "Reconcile Error", err.Error())
+		} else {
+			kc.Status.Error = "None"
+			r.Recorder.Event(kc, core.EventTypeNormal, "Updated", fmt.Sprintf("Updated K8ssandraCluster %s/%s", kc.Namespace, kc.Name))
```
burmanm, line 99: This feels odd event. We don't really update K8ssandraCluster, other than the Status. But I don't see that as an event that we should notify the user about.

controllers/k8ssandra/remove_dc_test.go
```diff
@@ -48,6 +49,63 @@ func deleteDcWithUserKeyspaces(ctx context.Context, t *testing.T, f *framework.F
 	mockMgmtApi.On(testutils.ListKeyspaces, "").Return(userKeyspaces, nil)
 	mockMgmtApi.On(testutils.GetSchemaVersions).Return(map[string][]string{"fake": {"test"}}, nil)
 
+	for _, ks := range userKeyspaces {
+		mockMgmtApi.On(testutils.GetKeyspaceReplication, ks).Return(replicationStr, nil)
+		//mockMgmtApi.On(testutils.AlterKeyspace, ks, updatedReplication).Return(nil)
```
burmanm, line 54: Dead comment.

controllers/k8ssandra/datacenters.go
adejanovski: I don't think so, I'll change this to `MergeFrom()`.

controllers/k8ssandra/k8ssandracluster_controller.go
adejanovski: I was wondering if that was a good idea. I'll remove it.

controllers/k8ssandra/remove_dc_test.go
adejanovski: good catch 👍

controllers/k8ssandra/schemas.go
```diff
@@ -284,14 +256,8 @@ func (r *K8ssandraClusterReconciler) updateUserKeyspacesReplicationForDecommissi
 		if err != nil {
 			return result.Error(fmt.Errorf("failed to get replication for keyspace (%s): %v", ks, err))
 		}
-		if _, updateNeeded := replication[decommDc]; updateNeeded {
-			delete(replication, decommDc)
-			if err = mgmtApi.AlterKeyspace(ks, replication); err != nil {
```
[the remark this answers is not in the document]
adejanovski: This isn't specific to DSE, so I'd rather avoid different behaviors unless we have no other choice.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -86,6 +90,12 @@ func (r *K8ssandraClusterReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	patch := client.MergeFromWithOptions(kc.DeepCopy())
 	result, err := r.reconcile(ctx, kc, logger)
 	if kc.GetDeletionTimestamp() == nil {
+		if err != nil {
+			kc.Status.Error = err.Error()
+			r.Recorder.Event(kc, core.EventTypeWarning, "Reconcile Error", err.Error())
```
jsanda, line 95: It would be nice to see specific, concrete error types reported vs `Reconcile Error`. If I saw for the first time while using an operator I might think it is an unexpected error or bug in the controller. 

Note that we already have a `K8ssandraError` type. See [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/errors/errors.go#L20).

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -85,6 +85,9 @@ type K8ssandraClusterStatus struct {
 	//
 	// TODO Figure out how to inline this field
 	Datacenters map[string]K8ssandraStatus `json:"datacenters,omitempty"`
+
+	// +kubebuilder:default=None
+	Error string `json:"error,omitempty"`
```
jsanda, line 90: It would be nice to add some docs for this.

controllers/k8ssandra/k8ssandracluster_controller.go
adejanovski: I'm not sure I understand what you want. Would you have an example?
Seems like golang errors require a switch to check what type of error we're dealing with and we could get many different types here.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adejanovski: Sure thing 👍

controllers/k8ssandra/schemas.go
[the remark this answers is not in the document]
burmanm: But now we're removing a feature that's better than the new one, which feels odd. The automation is the right thing to do here, even if we're backing off to please some users who don't trust computers.

Issue #617, opened 2022-07-11
Add a valid status condition for K8ssandraCluster


jsanda:
**What is missing?**
We have a lot of OpenAPI validation checks for the K8ssandraCluster CRD. For checks that have to be applied dynamically at runtime we have a validating webhook. There are also some validation checks performed during the reconciliation loop. 

Several things are needed:

* There needs to be a `Valid` status condition that is set to false when there are validation errors
    * The condition should include specific details about what the errors are
* The validation checks performed in the webhook should also be performed during reconciliation (and vice versa)
*  Validation checks should be performed as early as possible during reconciliation   

**Why do we need it?**
First, let's address why we need validation checks during reconciliation if we have a validating webhook. It's because the operator can be deployed without the webhook.

We need a `Valid` condition to better surface validation errors. When there is a validation error during reconciliation, the user won't find out until he looks at the operator logs. The status condition will let him know right away and can be easily monitored.

Validation checks should be performed as early as possible during reconciliation. We should fail fast and avoid doing unnecessary work if the desired state is invalid. That's not the case today. Furthermore, we have validation checks scattered in different places. Ideally, those checks should happen in a single place.

**Environment**

* K8ssandra Operator version:

    `v1.1.1`
    
    

**Anything else we need to know?**:

adejanovski:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/K8ssandra-Grooming-623b197f1fcd4800110540ef/issues/k8ssandra/k8ssandra-operator/617?planning-poker) with ZenHub @jsanda @Miles-Garnsey

PR #709, opened 2022-10-04
Disallow rack name change of a live cluster

### LINKED ISSUE #667: K8SSAND-1763 ⁃ Operator should not allow rack name change of a live cluster
**What is missing?**

Changing the rack name of a live cluster (and thus changing the NetworkTopology) is unsupported.
The operator should reflect that and explicitely block this operation.
Instead, it tries to carry out this change and new Cassandra replicas display such errors:

```
INFO  [main] 2022-09-03 17:22:43,469 NetworkTopologyStrategy.java:88 - Configured datacenter replicas are dc1:rf(3)
ERROR [main] 2022-09-03 17:22:43,472 CassandraDaemon.java:911 - Fatal configuration error
org.apache.cassandra.exceptions.ConfigurationException: Token allocation failed: the number of racks 2 in datacenter dc1 is lower than its replication factor 3.
```

**Why do we need it?**

Protect from accidental changes
Make the operator safer to use

**Environment**

* K8ssandra Operator version: 1.2.0

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1763) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1763
┆priority: Medium

olim7t:
**What this PR does**:
Add checks in the webhook and controller that no racks are deleted, and the rack names do not change (according to the declaration order).

Fixes #667

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
@@ -18,7 +18,6 @@ package v1alpha1
 
 import (
 	"fmt"
-
 	"github.com/k8ssandra/k8ssandra-operator/pkg/clientcache"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -133,6 +132,10 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) error {
 		return ErrClusterName
 	}
 
+	if err := validateNoRackRenamed(oldCluster.Spec.Cassandra, r.Spec.Cassandra); err != nil {
+		return err
+	}
+
 	// Some of these could be extracted in the cass-operator to reusable methods, do not copy code here.
 	// Also, reusing methods from cass-operator allows to follow updates to features if they change in cass-operator,
 	// such as allowing rack modifications or expanding PVCs.
@@ -142,6 +145,56 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) error {
 	// TODO Racks can only be added and only at the end of the list - no other operation is allowed to racks
 
 	return nil
+}
+
+func validateNoRackRenamed(oldCassandra, newCassandra *CassandraClusterTemplate) error {
+	oldDcs := collectRackNamesPerDc(oldCassandra)
+	newDcs := collectRackNamesPerDc(newCassandra)
+
+	for dc, oldRacks := range oldDcs {
+		if newRacks, ok := newDcs[dc]; ok {
+			if len(newRacks) < len(oldRacks) {
+				return fmt.Errorf("number of racks can't be lowered (DC %s, current=%d, desired=%d)",
+					dc, len(oldRacks), len(newRacks))
+			}
+			for i, oldRack := range oldRacks {
+				newRack := newRacks[i]
+				if newRack != oldRack {
+					return fmt.Errorf("racks can't be renamed (DC %s, index %d, current=%s, desired=%s)",
+						dc, i, oldRack, newRack)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func collectRackNamesPerDc(cassandra *CassandraClusterTemplate) map[string][]string {
+	if len(cassandra.Datacenters) == 0 {
+		// Unlikely but we allow it
+		return map[string][]string{}
+	}
+
+	// The top-level object can define default racks, to be inherited by DCs that don't declare them.
+	// If they aren't declared anywhere, there is a single rack named "default".
+	defaultNames := collectRackNames(cassandra.DatacenterOptions, []string{"default"})
+
+	m := make(map[string][]string, len(cassandra.Datacenters))
+	for _, dc := range cassandra.Datacenters {
+		m[dc.Meta.Name] = collectRackNames(dc.DatacenterOptions, defaultNames)
+	}
+	return m
+}
+
+func collectRackNames(dc DatacenterOptions, ifNoRacks []string) []string {
+	if len(dc.Racks) == 0 {
+		return ifNoRacks
+	}
+	names := make([]string, len(dc.Racks))
+	for i, rack := range dc.Racks {
+		names[i] = rack.Name
+	}
+	return names
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook_test.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook_test.go
@@ -143,6 +143,7 @@ func TestWebhook(t *testing.T) {
 	t.Run("ReaperKeyspaceValidation", testReaperKeyspaceValidation)
 	t.Run("StorageConfigValidation", testStorageConfigValidation)
 	t.Run("NumTokensValidation", testNumTokens)
+	t.Run("RackNameValidation", testRackNameValidation)
 }
 
 func testContextValidation(t *testing.T) {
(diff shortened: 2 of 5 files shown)


apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -146,6 +146,10 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) error {
 		return ErrClusterName
 	}
 
+	if err := validateNoRackRenamed(oldCluster.Spec.Cassandra, r.Spec.Cassandra); err != nil {
+		return err
+	}
```
olim7t, line 151: I think this covers this from a few lines below:
```
// TODO Racks can only be added and only at the end of the list - no other operation is allowed to racks
```
Unless we want to control more fields in the rack configuration.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -105,8 +105,9 @@ func TestK8ssandraCluster(t *testing.T) {
 	t.Run("CreateMultiDcClusterWithMedusa", testEnv.ControllerTest(ctx, createMultiDcClusterWithMedusa))
 	t.Run("CreateSingleDcClusterNoAuth", testEnv.ControllerTest(ctx, createSingleDcClusterNoAuth))
 	t.Run("CreateSingleDcClusterAuth", testEnv.ControllerTest(ctx, createSingleDcClusterAuth))
-	// If webhooks are installed, this testcase is handled by the webhook test
+	// If webhooks are installed, these test cases are handled by the webhook test
 	// t.Run("ChangeNumTokensValue", testEnv.ControllerTest(ctx, changeNumTokensValue))
+	// t.Run("ChangeRackName", testEnv.ControllerTest(ctx, changeRackName))
```
olim7t, line 110: Note: I'm still having trouble running this locally (I commented out the webhook code to reach it, but it's giving me an error).

olim7t: It's fixed now.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -144,6 +147,56 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) error {
 	return nil
 }
 
+func validateNoRackRenamed(oldCassandra, newCassandra *CassandraClusterTemplate) error {
```
Miles-Garnsey, line 150: I would probably give a custom type the error you're returning, but this is a suggestion not a requirement.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook_test.go
```diff
@@ -319,6 +320,114 @@ func testNumTokens(t *testing.T) {
...
+			cluster.Spec.Cassandra.Datacenters[0].Racks = append(
+				cluster.Spec.Cassandra.Datacenters[0].Racks, v1beta1.Rack{Name: rackName})
+		}
+		return cluster
+	}
+
+	var oldCluster, newCluster *K8ssandraCluster
+
+	// DC-level declaration
+	oldCluster = createCluster(nil, []string{"rack1", "rack2"})
```
Miles-Garnsey, line 345: These tests might be a good candidate for a table driven style of testing https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook_test.go
```diff
@@ -319,6 +320,114 @@ func testNumTokens(t *testing.T) {
...
+	newCluster = createCluster([]string{"rack1"}, nil)
+	required.NoError(newCluster.ValidateUpdate(oldCluster))
+
+	oldCluster = createCluster([]string{"rack1"}, nil)
+	newCluster = createCluster(nil, []string{"rack1", "rack2"})
+	required.NoError(newCluster.ValidateUpdate(oldCluster))
+
+	oldCluster = createCluster(nil, []string{"rack1"})
+	newCluster = createCluster([]string{"rack1", "rack2"}, nil)
+	required.NoError(newCluster.ValidateUpdate(oldCluster))
```
Miles-Garnsey, line 428: From these tests, it isn't quite clear to me what should happen when you have multiple DCs. 

I'd almost like to suggest that you take a step back and stop providing the DC and cluster level racks. Instead perhaps you could provide the whole K8ssandraCluster struct to make sure that the coalescing logic is all covered?

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -144,6 +147,56 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) error {
...
+
+func collectRackNamesPerDc(cassandra *CassandraClusterTemplate) map[string][]string {
+	if len(cassandra.Datacenters) == 0 {
+		// Unlikely but we allow it
+		return map[string][]string{}
+	}
+
+	// The top-level object can define default racks, to be inherited by DCs that don't declare them.
+	// If they aren't declared anywhere, there is a single rack named "default".
+	defaultNames := collectRackNames(cassandra.DatacenterOptions, []string{"default"})
```
Miles-Garnsey, line 180: This logic made me think a bit... We do have a [coalesce](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L272) function which historically has been used to merge the cluster and DC level configs into a `DatacenterConfig`.

It seems like you're doing some similar logic here, and I wonder if this logic shouldn't be operating on the DatacenterConfig type and doing all validations against a coalesced data structure. At present it seems to be trying to coalesce the cluster and DC level settings, and it might be nicer to re-use the existing logic which does that.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -100,8 +100,9 @@ func TestK8ssandraCluster(t *testing.T) {
 	t.Run("CreateMultiDcClusterWithMedusa", testEnv.ControllerTest(ctx, createMultiDcClusterWithMedusa))
 	t.Run("CreateSingleDcClusterNoAuth", testEnv.ControllerTest(ctx, createSingleDcClusterNoAuth))
 	t.Run("CreateSingleDcClusterAuth", testEnv.ControllerTest(ctx, createSingleDcClusterAuth))
-	// If webhooks are installed, this testcase is handled by the webhook test
+	// If webhooks are installed, these test cases are handled by the webhook test
```
Miles-Garnsey, line 103: What happens when the webhook is not installed? Do we have matching logic to prevent racks from being modified in the reconciliation loop too? Historically, we've often put validation in both places to provide more protection.

Miles-Garnsey: Ah, it looks like the reconciliation logic is also protected down [here](https://github.com/k8ssandra/k8ssandra-operator/pull/709/files#diff-7e422b847100c3ee477ee61b97c8ad4da71063c1fc698d4e049081a98c89399cR535)?

Could we consolidate this logic into a single validation function (I'd suggest under `pkg/cassandra/racks/validation.go`) and call it from both the webhook and the reconciliation loop?

Miles-Garnsey (changes requested):
This looks great in many respects, but I'm not fully across what has been agreed RE design on this PR. My big questions would be:

1. I've read the comment that this logic already exists in cass-operator. I'd have thought that rack configs should be passed down through the CassDC which should do the checks against its own CRs rather than checking at the k8ssandra-operator layer. I think I'm missing something though.
2.  There are some functions here which also seem to appear in cass-operator (e.g. [here](https://github.com/k8ssandra/cass-operator/blob/master/apis/cassandra/v1beta1/cassandradatacenter_types.go#L284))
3. Is there more existing logic in k8ssandra-operator we can leverage? Where can we consolidate the validation logic?
4. What structs should the validation logic be collaborating with? I'd try to focus on verifying that K8ssandraClusters are processed correctly. I'd also look at the DatacenterConfig as the struct to validate.
5. (Related to 4) Are the tests sitting at the right level, and (given there are a lot) could they be structured more clearly?

pkg/cassandra/datacenter.go
```diff
@@ -532,6 +532,20 @@ func ValidateConfig(desiredDc, actualDc *cassdcapi.CassandraDatacenter) error {
 		return fmt.Errorf("tried to change num_tokens in an existing datacenter")
 	}
 
+	desiredRacks := desiredDc.GetRacks()
+	actualRacks := actualDc.GetRacks()
+	if len(desiredRacks) < len(actualRacks) {
```
Miles-Garnsey, line 537: It is frustrating that we can't fold this into [this](https://github.com/k8ssandra/k8ssandra-operator/pull/709/files#diff-7e422b847100c3ee477ee61b97c8ad4da71063c1fc698d4e049081a98c89399cR541) logic. Obviously, we'll get a panic due to the index being out of bounds if len(desiredRacks) < len(actualRacks). I wonder whether it would be clearer to `recover` from the panic instead of doing the bounds check up front.

Miles-Garnsey: That suggestion is 100% non-blocking by the way, just a stylistic thing.

burmanm: No, never do panic + recover. That's worse than coding Java with always ``} catch(NullPointerException e) {``  instead of doing null checks. Go's panics are not exceptions and are not meant to be used this way.

controllers/k8ssandra/k8ssandracluster_controller_test.go
burmanm: > What happens when the webhook is not installed? Do we have matching logic to prevent racks from being modified in the reconciliation loop too?

Do we need that here also? cass-operator reconciliation logic would block that. Does it really matter which one fails?

Miles-Garnsey: That's a fair question. I don't see why we shouldn't do the check here to minimise the number of logs the user needs to follow. If they're looking for an error in the k8ssandra-operator logs it might be confusing if it turns up in the cass-operator logs.

pkg/cassandra/datacenter.go
Miles-Garnsey: Sure, it was just a thought bubble really. Do you have any docs/resources on why this is a bad idea?

controllers/k8ssandra/k8ssandracluster_controller_test.go
burmanm: We can, but I'm not sure if we're close to that "only check single logs" in reality. But that's far beyond the scope of this ticket. 

The entire webhook validation should probably be in a single method that could be called by the webhook or start of the reconcile.

olim7t:
> If they're looking for an error in the k8ssandra-operator logs it might be confusing if it turns up in the cass-operator logs.

It even goes a bit further than that: with the `main` branch, the error is far from obvious. Cass-operator does indeed reject the change, and the `CassandraDatacenter` is not modified. But `kubectl apply` appears to succeed, the only feedback is in the `K8ssandraCluster.Error` field (if it even occurs to the user to check it):
```
admission webhook "vcassandradatacenter.kb.io" denied the request: CassandraDatacenter write rejected, attempted to change rack name from 'rack1' to 'rack12'
```
What's worse, the rack name in the `K8ssandraCluster` spec _is_ updated, so it now shows an erroneous value that doesn't match the actual `CassandraDatacenter`.

What this PR gives us is clear feedback as soon as you try to apply the changes:
```shell
$ k apply -f k8c_with_renamed_rack.yaml
Error from server (racks can't be renamed (DC dc1, index 0, current=rack1, desired=rack12))...
```
And the `K8ssandraCluster` is not modified.


> The entire webhook validation should probably be in a single method that could be called by the webhook or start of the reconcile.

Agreed. One thing that occurred to me is that this PR handles one particular check, but they are probably others that we'd similarly want to duplicate.

The hard part is that we are dealing with different separate hierarchies depending on the level. For example in this PR the webhook deals exclusively with k8ssandra-operator structures, while the controller code handles a mix of k8ssandra-operator and cass-operator (this is why I had to implement the check twice). And obviously the logic in cass-operator, if we can reuse it, will be 100% cass-operator structures.

controllers/k8ssandra/k8ssandracluster_controller_test.go
Miles-Garnsey: So we agree on single-method. But I also think that method should get a call in the reconciliation loop too. I'm not dogmatic about it, but it seems like a small piece of logic to provide a bit of a usability uplift - albeit in rare situations.

Miles-Garnsey:
> (this is why I had to implement the check twice).

Unfortunately that's the nature of this codebase. The best way forward IMO is to re-use cass-operator logic where possible within k8ssandra-operator, and re-use a `validation` package in both the webhook and reconciliation within k8ssandra-operator. Some level of duplication is to be expected given we're operating across a number of systems, and each needs to do its own validation.

I do also suggest that you consider operating on the `DatacenterConfig` so that you're maximally re-using the logic we have for merging the cluster and DC levels. Hopefully that'll at least help reduce some of the acrobatics :)

pkg/cassandra/datacenter.go
burmanm: panic() has different behavior than normal function returns and thus if the caller is not aware (such as when using goroutines), the execution could be impacted in ways not expected by the user. "The convention in the Go libraries is that even when a package uses panic internally, its external API still presents explicit error return values." (https://go.dev/blog/defer-panic-and-recover)

The problem with panic is that it immediately starts to unwind the goroutine stack. If we look from the perf perspective, then I don't think Go's compiled code even uses traps (like Java for exceptions) to catch these situations, so one could take memory / perf impact also. But that's just IIRC and could have changed during years.

Mostly I'd say because it's a thing that Go itself doesn't use in APIs and as such it's not a idiom (with all the caveats) that usually the caller expects. So, unidiomatic code.

controllers/k8ssandra/k8ssandracluster_controller_test.go
burmanm: I wasn't disagreeing with the reconciliation loop too if implementation is done correctly (but lets not repeat code so that the reconciliation could have different validation rules after time passes). Just saying that we have an issue where error location is already a bit unclear in many situations. cass-operator might decide to do something that k8ssandra-operator doesn't know about, so for a user that's something they need to learn to check (at least at this point).

olim7t:
The more I look at this, the more duplicating the checks sounds like a bad idea.

As Michael said:
> cass-operator might decide to do something that k8ssandra-operator doesn't know about

In other words, we'd have to keep everything in sync over time, which would certainly be tedious and error-prone.
The [rack checks in cass-operator](https://github.com/k8ssandra/cass-operator/blob/master/apis/cassandra/v1beta1/cassandradatacenter_webhook.go#L154) are already more subtle than what I implemented here.

The `K8ssandraClusterStatus.Error` field was added in #700, but reading the ticket it sounds like it was intended for a specific use case. I wonder if we should be more selective in the errors we surface in there (as opposed to failing the reconcile directly).

jsanda:
@olim7t take a look at #617. I think it may be along the lines of what we want.

olim7t:
> The `K8ssandraClusterStatus.Error` field was added in #700, but reading the ticket it sounds like it was intended for a specific use case. I wonder if we should be more selective in the errors we surface in there (as opposed to failing the reconcile directly).

Scratch that, cass-operator errors will never be surfaced directly, regardless of that field.

olim7t:
Closing this as an ad-hoc check is too simplistic, and we're debating duplicating the whole cass-operator validation logic. I'll add a comment on the original ticket.

PR #713, opened 2022-10-06
Add token computation function

### LINKED ISSUE #728: K8SSAND-1845 ⁃ Add token generation utility
**What is missing?**

We need a token generation utility in the operator. 

**Why do we need it?**

This utility could be used to generate initial tokens for clusters opting for single-token nodes, and thus avoid unbalanced token distributions.

The utility should handle multi-dc.

Cassandra OSS has such a utility: https://github.com/apache/cassandra/blob/trunk/tools/bin/generatetokens. However it doesn't handle the multi-dc case.

DSE also has a similar utility. I suggest that we use it instead since it handles multi-dc (and is licensed under a compatible license).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1845) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1845
┆priority: Medium

olim7t:
**What this PR does**:
Add a token computation function for the cases where num_token is low we won't rely on automatic token allocation.

This is adapted from an existing script in the DSE codebase; I've focused on producing identical results, even though some parts of the algorithm are perplexing (see inline comments).

Fixes #728 

**Checklist**
- [x] Changes manually tested


pkg/utils/tokens.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+var RandomPartitioner = Partitioner{
+	RingOffset: new(big.Int),               // 0
+	RingRange:  new(big.Int).Lsh(one, 127), // 1<<127
+}
+
+// ComputeTokens takes a list of numbers indicating the number of nodes in each datacenter, and returns a recommended
+// list of tokens for the given partitioner (one token per node, grouped by datacenter).
+func ComputeTokens(dcCounts []int, partitioner Partitioner) [][]string {
```
olim7t, line 27: Returning strings because I think that will be the most convenient in the calling code. If we prefer raw numerical values it's an easy change.

pkg/utils/tokens.go
```diff
@@ -0,0 +1,89 @@
...
+		if dcCount > maxDcCount {
+			maxDcCount = dcCount
+		}
+	}
+
+	// Ensure that we preserve the original order, i.e. the offset does not push us past the next token of the previous
+	// DC. For example, if there are 3 DCs with 4 nodes each, and we picture the ring as a 360° circle, the first two
+	// tokens of DC1 would be at 0° and 90°. So the next two starting points must be between 0 and 90.
+	// 360 / (3*4*2) gets us there.
+	divisor := len(dcCounts) * maxDcCount * 2
```
olim7t, line 73: This one I kind of understand.
Even though if the DC sizes are not equal, the tokens at the end will end up out of order anyway, but that's unavoidable if we divide evenly.

pkg/utils/tokens.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Copied verbatim from the DSE code that this is adapted from. The intent is not clear (it looks like this will
+	// almost always kick in, so the previous computation is irrelevant). But we keep it as-is to be able to compare
+	// results.
+	const minDivisor = 235
+	if divisor < minDivisor {
+		divisor = minDivisor
+	}
```
olim7t, line 81: This is really puzzling. As I mentioned in the comment, the divisor will be 235 most of the time, unless there is a really high number of nodes.

The original DSE code mentions something about keeping the offset large enough so that it is visible in a graphical representation, but for that we'd need a _max_ divisor, not a min.

pkg/utils/tokens.go
```diff
@@ -0,0 +1,89 @@
...
+
+	result := new(big.Int).Neg(partitioner.RingRange)
+	return result.Div(result, bigInt(divisor))
```
olim7t, line 84: Not sure why we negate here, but again keeping it as-is.

CHANGELOG/CHANGELOG-1.3.md
```diff
@@ -15,6 +15,7 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+* [FEATURE] Add token computation function
```
adutra, line 18: We are missing the issue link.

pkg/utils/tokens.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func bigInt(i int) *big.Int {
+	return new(big.Int).SetInt64(int64(i))
```
adutra, line 88: FYI, you have the constructor function `big.NewInt()`.

pkg/utils/tokens_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestComputeTokens(t *testing.T) {
+	required := require.New(t)
+
+	// All the expected data was generated with the original DSE script
+
+	required.Equal(
```
adutra, line 13: We might want to add tests for 2 very common cases: `[]int{3}` and `[]int{3,3}`. 

We could also introduce fuzz testing; in this case, the test could verify e.g. that the resulting ring is evenly distributed. Just a suggestion.

pkg/utils/tokens.go
adutra: This is indeed puzzling. Do we know the authors of the original script?

adutra (commented):
Just a superficial review, I haven't looked into the algorithm in detail, but it does look puzzling.

adutra:
@olim7t do you mind adding at least the issue URL in the changelog so that we can merge this? The rest can wait imo.

CHANGELOG/CHANGELOG-1.3.md
olim7t: There's no public ticket for the issue as far as I know.

olim7t:
Rebased and addressed some of the feedback.

olim7t:
For what it's worth, there's also a token generation utility in open-source Cassandra. It plugs directly into the Cassandra internals (see [GenerateTokens.java](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/tools/GenerateTokens.java)), so extracting the logic would not be as straightforward.

A few differences in the expected input:
- it only handles one datacenter, there's no explanation on how you would extrapolate to multiple DCs.
- the replication factor must be provided, and it does influence the results (I've checked). I don't really understand why.

On a single DC it generates different results than the DSE script.

CHANGELOG/CHANGELOG-1.3.md
adutra: Indeed, we need to create it. I went ahead and did it: https://github.com/k8ssandra/k8ssandra-operator/issues/728.

adutra (approved):

adutra:
@olim7t all is good I fixed the changelog and merged this. Thanks!

PR #716, opened 2022-10-11
Don't restart whole Cassandra 4 DC when Stargate is added or removed

### LINKED ISSUE #714: K8SSAND-1820 ⁃ Removing Stargate triggers a DC restart
**What happened?**

When Stargate is removed from a K8ssandraCluster object suing Cassandra 4.0, this triggers a DC restart.

I expect no DC restart.

1. Create a cluster with 1 DC and Stargate
1. Delete Stargate from the CRD
1. Observe the DC being restarted

**Environment**

* K8ssandra Operator version:

1.2

Due to this code in the k8ssandra controller:

```
if dcConfig.ServerVersion.Major() != 3 && kc.HasStargates() {
	// if we're not running Cassandra 3.11 and have Stargate pods, we need to allow alter RF during range movements
	cassandra.AllowAlterRfDuringRangeMovement(dcConfig)
}
```

I suggest that we always call `cassandra.AllowAlterRfDuringRangeMovement(dcConfig)` for Cassandra 4.0, regardless of whether we have Stargate nodes.

See also https://github.com/stargate/stargate/issues/1274.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1820) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1820
┆priority: Medium

olim7t:
**What this PR does**:
Always set `AllowAlterRfDuringRangeMovement` for Cassandra 4 (not just when there are Stargate nodes).
Otherwise adding/removing Stargate node triggers a configuration change and a restart of the DC.

Fixes #714

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/k8ssandracluster_controller_test.go
+++ b/controllers/k8ssandra/k8ssandracluster_controller_test.go
@@ -95,6 +95,7 @@ func TestK8ssandraCluster(t *testing.T) {
 	t.Run("ApplyClusterTemplateConfigs", testEnv.ControllerTest(ctx, applyClusterTemplateConfigs))
 	t.Run("ApplyDatacenterTemplateConfigs", testEnv.ControllerTest(ctx, applyDatacenterTemplateConfigs))
 	t.Run("ApplyClusterTemplateAndDatacenterTemplateConfigs", testEnv.ControllerTest(ctx, applyClusterTemplateAndDatacenterTemplateConfigs))
+	t.Run("CreateSingleDcCassandra4ClusterWithStargate", testEnv.ControllerTest(ctx, createSingleDcCassandra4ClusterWithStargate))
 	t.Run("CreateMultiDcClusterWithStargate", testEnv.ControllerTest(ctx, createMultiDcClusterWithStargate))
 	t.Run("CreateMultiDcClusterWithReaper", testEnv.ControllerTest(ctx, createMultiDcClusterWithReaper))
 	t.Run("CreateMultiDcClusterWithMedusa", testEnv.ControllerTest(ctx, createMultiDcClusterWithMedusa))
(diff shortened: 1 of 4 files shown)


test/e2e/suite_test.go
```diff
@@ -740,6 +741,11 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace
 		}
 	}, polling.k8ssandraClusterStatus.timeout, polling.k8ssandraClusterStatus.interval)
 
+	t.Log("check that Cassandra DC was not restarted")
+	err = f.Client.Get(ctx, kcKey, k8ssandra)
+	require.NoError(err, "failed to get K8ssandraCluster in namespace %s", namespace)
+	require.Equal(dcGeneration, k8ssandra.Status.Datacenters["dc1"].Cassandra.ObservedGeneration)
+
```
olim7t, line 748: This is a bit futile given that the test uses Cassandra 3, I don't see any that uses 4. But I've verified manually that the test passes (and fails before the fix) with 4.0.6.

adejanovski: We've had more flakiness in the tests with 4.0, possibly due to it requiring more resources 🤷 that's why all the tests moved to 3.11.
You could turn a light test (a single dc one) to use 4.0 and see how it goes maybe? The simplest one shouldn't deploy Stargate, which could be a good fit for your test.

adejanovski: You can work around this by just checking if the flag is added even if Stargate isn't there. That would verify your change is working.

adejanovski (approved):
If you don't have time to make one of the e2e tests use 4.0, let's move forward and get this merged.

test/e2e/suite_test.go
olim7t: In that case it should probably not be an e2e test. I'm trying to write an integration test right now, but I'm less familiar with that part of the codebase.

jsanda: +1 to an integration test

adejanovski: works for me 👍

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -970,6 +971,154 @@ func setReplicationStatusDone(ctx context.Context, t *testing.T, f *framework.Fr
...
+	patch := client.MergeFromWithOptions(kc.DeepCopy(), client.MergeFromWithOptimisticLock{})
+	kc.Spec.Cassandra.Datacenters[0].Stargate = nil
+	err = f.Client.Patch(ctx, kc, patch)
+	require.NoError(err, "failed to update K8ssandraCluster")
+
+	t.Log("check that stargate sg1 is deleted")
+	require.Eventually(func() bool {
+		err = f.Get(ctx, sg1Key, &stargateapi.Stargate{})
+		return errors.IsNotFound(err)
+	}, timeout, interval)
```
olim7t, line 1099: Is this enough to guarantee that I will observe the latest versions of the `K8ssandraCluster` and `cassdcapi.CassandraDatacenter` in the lines that follow, or should I wait on another condition?

jsanda: The check is fine. Keep in mind that the controllers using a caching client by default. They perform reads from a cache which means that they can be stale. Tests however are configure client a client that reads directly from the API server which means they they will always observe the latest version with respect to what is in etcd. Polling as you have done here with the `Eventually` call is needed when we need to wait for changes to be made by the controller, i.e., desired state --> actual state.

jsanda:
@olim7t The envtest looks good :)

PR #717, opened 2022-10-14
Use Cassandra internals for JMX authentication

olim7t:
**What this PR does**:
- configure Cassandra to delegate JMX authentication to the underlying CQL roles
- remove the now-obsolete JMX init container
- stop creating a JMX secret for Reaper; use its CQL secret instead.

Fixes k8ssandra/k8ssandra#323

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/reaper/v1alpha1/reaper_types.go
+++ b/apis/reaper/v1alpha1/reaper_types.go
@@ -49,8 +49,8 @@ type ReaperTemplate struct {
 	// +optional
 	CassandraUserSecretRef corev1.LocalObjectReference `json:"cassandraUserSecretRef,omitempty"`
 
-	// +kubebuilder:deprecatedversion:warning="JMX security is now based on CQL roles. Reaper will use
-	// cassandraUsefSecretRef for authentication, JmxUserSecretRef is now ignored."
+	// Defines the username and password that Reaper will use to authenticate JMX connections to Cassandra clusters.
+	// +kubebuilder:deprecatedversion:warning="JMX security is now based on CQL roles. Reaper will use CassandraUsefSecretRef for authentication, JmxUserSecretRef is now ignored."
 	// +optional
 	JmxUserSecretRef corev1.LocalObjectReference `json:"jmxUserSecretRef,omitempty"`
 
--- a/apis/reaper/v1alpha1/reaper_types.go
+++ b/apis/reaper/v1alpha1/reaper_types.go
@@ -49,11 +49,8 @@ type ReaperTemplate struct {
 	// +optional
 	CassandraUserSecretRef corev1.LocalObjectReference `json:"cassandraUserSecretRef,omitempty"`
 
-	// Defines the username and password that Reaper will use to authenticate JMX connections to Cassandra clusters.
-	// These credentials will be automatically passed to each Cassandra node in the datacenter, as well as to the Reaper
-	// instance, so that the latter can authenticate against the former. If JMX authentication is not required, leave
-	// this field empty. The secret must be in the same namespace as Reaper itself and must contain two keys: "username"
-	// and "password".
+	// Deprecated: JMX security is now based on CQL roles. Reaper will use cassandraUsefSecretRef for authentication,
+	// this field is ignored.
 	// +optional
 	JmxUserSecretRef corev1.LocalObjectReference `json:"jmxUserSecretRef,omitempty"`
 
--- a/apis/reaper/v1alpha1/reaper_types.go
+++ b/apis/reaper/v1alpha1/reaper_types.go
@@ -49,9 +49,8 @@ type ReaperTemplate struct {
 	// +optional
 	CassandraUserSecretRef corev1.LocalObjectReference `json:"cassandraUserSecretRef,omitempty"`
 
-	// Defines the username and password that Reaper will use to authenticate JMX connections to Cassandra clusters.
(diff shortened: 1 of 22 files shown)


pkg/cassandra/auth.go
```diff
@@ -34,48 +33,12 @@ func ApplyAuth(dcConfig *DatacenterConfig, authEnabled bool) error {
...
-				MountPath: "/config",
-			}}
-			c.Args = []string{
-				"/bin/sh",
-				"-c",
-				"echo \"$SUPERUSER_JMX_USERNAME $SUPERUSER_JMX_PASSWORD\" >> /config/jmxremote.password",
-			}
-		})
+		addOptionIfMissing(dcConfig, "-Dcassandra.jmx.remote.login.config=CassandraLogin")
+		addOptionIfMissing(dcConfig, "-Djava.security.auth.login.config=$CASSANDRA_HOME/conf/cassandra-jaas.config")
```
adutra, line 40: What are the contents of cassandra-jaas.config? This file doesn't seem supported by config-builder, so I wonder if it's being generated at all? In cassandra-env.sh, it is said that this file is merely a "sample JAAS configuration".

olim7t: It does exist in the cassandra container, with the following contents:
```
// Delegates authentication to Cassandra's configured IAuthenticator
CassandraLogin {
  org.apache.cassandra.auth.CassandraLoginModule REQUIRED;
};
```
It points the JAAS engine to the implementation of `CassandraLogin` referenced in the previous option.

CHANGELOG/CHANGELOG-1.3.md
```diff
@@ -13,6 +13,10 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
+## unreleased
+
+* [ENHANCEMENT] [#323](https://github.com/k8ssandra/k8ssandra/issues/323) Use Cassandra internals for JMX authentication
```
olim7t, line 18: Should we indicate in any way that the issue is in a different project?

adutra: IMHO we are fine. That happened before, and is usually a symptom that some issue was logged a long time ago in the k8ssandra repo, but was fixed in the k8ssandra-operator repo since we don't bring new features anymore to the k8ssandra repo. I expect these situations to happen less and less frequently though.

adutra (approved):

pkg/cassandra/datacenter.go
```diff
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+// TODO remove
 var DefaultJmxInitImage = images.Image{
```
olim7t, line 23: Hrmm, I forgot to address this, will do.

adejanovski (commented):
I have a few minor comments, around the kubebuilder deprecation annotation and a remaining reference to the jmx-credentials init container.
Did you get to test this change by upgrading from a cluster with Reaper using the old JMX credential system?
I'm a little nervous as to how much of a breaking change that would be, especially if a specific jmx user/password is used that is different from the superuser 🤔

olim7t:
> Did you get to test this change by upgrading from a cluster with Reaper using the old JMX credential system?
> I'm a little nervous as to how much of a breaking change that would be, especially if a specific jmx user/password is used that is different from the superuser 🤔

I tested the following:
- run `make single-up` from the `main` branch
- create a cluster with Reaper and a dedicated JMX secret
- checkout this branch and run `make single-reload`

The Cassandra statefulset gets updated, the new pods don't have the jmx-init container.
A new Reaper replicaset gets created. The environment variables now reference the CQL secret (generated in my case) instead of the JMX secret:
```
REAPER_JMX_AUTH_PASSWORD : secretKeyRef(demo-reaper.password)
```

The only unexpected outcome is that the old Reaper replicaset isn't deleted, just downscaled to 0. I'll look into it.

olim7t:
> The only unexpected outcome is that the old Reaper replicaset isn't deleted, just downscaled to 0.

nm, that's actually normal behavior given that `revisionHistoryLimit` defaults to 10 on the Reaper Deployment.

apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -50,10 +50,7 @@ type ReaperTemplate struct {
 	CassandraUserSecretRef corev1.LocalObjectReference `json:"cassandraUserSecretRef,omitempty"`
 
 	// Defines the username and password that Reaper will use to authenticate JMX connections to Cassandra clusters.
-	// These credentials will be automatically passed to each Cassandra node in the datacenter, as well as to the Reaper
-	// instance, so that the latter can authenticate against the former. If JMX authentication is not required, leave
-	// this field empty. The secret must be in the same namespace as Reaper itself and must contain two keys: "username"
-	// and "password".
+	// +kubebuilder:deprecatedversion:warning="JMX security is now based on CQL roles. Reaper will use CassandraUsefSecretRef for authentication, JmxUserSecretRef is now ignored."
```
adutra, line 53: I think this annotation is valid on types or packages only.

adejanovski:
@olim7t, now that 1.4.0 was released, you'll need to rebase on top of main and move the changelog entry to the `CHANGELOG-1.5.md` file.
After that you can go ahead and merge ✅

adejanovski (approved):
We're good for a merge 👍

adejanovski:
@olim7t, could you rebase one last time and merge the PR? Thanks!

PR #725, opened 2022-10-17
Ability to provide per-node configuration

### LINKED ISSUE #724: K8SSAND-1843 ⁃ Ability to provide per-node configuration
**What is missing?**

Ability to provide configuration overrides on a per-node basis.

**Why do we need it?**

So that we can leverage individual pod configuration tweaks. Examples include `initial_token` assignments and pod-specific TLS certificates.

The proposal would be to store specific pe-node config in a config map:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-dc1-per-node-config
data:
  test-dc1-default-sts-0_cassandra.yaml: |
      initial_token: 0,100,200
```

Each entry is of the form: `<pod name>_<yaml_file>`. Each entry should contain overrides for the target yaml file that will be merged into it.

A new init container needs to be created to update the config files with pod specific overrides, using the mounted config map.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1843) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1843
┆priority: Medium

adutra:
**What this PR does**:

Ability to provide per-node configuration.

Fixes #724 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/per_node_config.go
+++ b/controllers/k8ssandra/per_node_config.go
@@ -9,6 +9,7 @@ import (
 	"github.com/k8ssandra/k8ssandra-operator/pkg/result"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -35,7 +36,6 @@ func (r *K8ssandraClusterReconciler) reconcilePerNodeConfiguration(
 		return result.Error(err)
 	}
 
-	// TODO do we need to own the cm and add a hash annotation?
 	logger.Info("Found per-node ConfigMap, mounting")
 
 	// if the config-builder init container isn't found, declare a placeholder now to guarantee order of execution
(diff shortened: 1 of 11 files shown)


controllers/k8ssandra/per_node_config.go
```diff
@@ -0,0 +1,109 @@
...
+		{
+			Name:      perNodeConfigVolumeName,
+			MountPath: "/per-node-config",
+		},
+	},
+	Command: []string{
+		"sh",
+		"-c",
+		"for src in /per-node-config/${POD_NAME}_*; do " +
+			"dest=/config/`echo $src | cut -d \"_\" -f2`; " +
```
adutra, line 82: Since we are tokenizing by underscore, the file name itself should not contain underscores. I wonder if we should opt for a more robust delimiter? Maybe `|`?

controllers/k8ssandra/per_node_config.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// add per-node config volume to pod spec
+	cassandra.AddVolumesToPodTemplateSpec(dcConfig, newPerNodeConfigVolume(perNodeConfigKey.Name))
+
+	return result.Continue()
+}
+
+const perNodeConfigVolumeName = "per-node-config"
+
+var perNodeConfigInitContainer = &corev1.Container{
```
adejanovski, line 55: Would it be possible to set default resources requests/limits?
There are environments where pods containing (init-)containers without those will get rejected straight away.

adejanovski (changes requested):
Brilliant execution!
I just have a tiny request about adding default resources limits/requests on the newly introduced init-container, but aside from this it all looks great to me 👍

adejanovski (approved):

PR #731, opened 2022-10-19
Add the ability to set resources limits on the server-config-init container

### LINKED ISSUE #723: Ability to override resources for default init-containers
**What is missing?**
Ability to override resource limits and requests for CassandraDatacenter default initContainers.

**Why do we need it?**
Under certain conditions, cassandra datacenter pods do not start because the server-config-init container is configured with too little memory and crashes with a status of 137, i.e. OOMKilled.

**Environment**

* K8ssandra Operator version:

    `v1.3.0`
    
    
* K8s version:

   `v1.23.7`

Relevant snippet of K8ssandraDatacenter resource definition:
```
  cassandra:
    serverVersion: "4.0.1"
    superuserSecretRef:
      name: [redacted]
    datacenters:
      - metadata:
          name: dc1
        size: 5
#        initContainers:
#          - name: server-config-init # defaults cannot be overridden ?
#            resources:
#              requests:
#                cpu: 1
#                memory: 1G
#              limits:
#                cpu: 1
#                memory: 1G
        storageConfig:
          cassandraDataVolumeClaimSpec:
            storageClassName: local-path
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 500Gi
        resources:
          requests:
            memory: 8G
          limits:
            memory: 24G
        telemetry:
          prometheus:
            enabled: true
```
In our environment, the server-config-init container is automatically configured with a memory limit of 256Mi and memory request of 256Mi. This does not seem to be enough for the config application to start correctly. We have found that 512Mi is sufficient in our case and would like to set this from the get-go instead of patching the statefulset post-deployment. Node resources are sufficient in every case.

adejanovski:
**What this PR does**:
cass-operator doesn't merge the server-config-init init container definition with what can be provided through the K8ssandraCluster `.spec.cassandra.initContainers`. Thus, changing its resources limits and requests has to be done through the cassdc `.spec.configBuilderResources` field.
This PR exposes it in the `DatacenterOptions` struct so that it can be set in a K8ssandraCluster object. 

Fixes #723 

**Checklist**
- [x] Changes manually tested


apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -383,6 +383,11 @@ type DatacenterOptions struct {
 
 	// +optional
 	DseWorkloads *cassdcapi.DseWorkloads `json:"dseWorkloads,omitempty"`
+
+	// Kubernetes resource requests and limits for the server-config-init init-container that is created by cass-operator to write the Cassandra config files.
+	// The default memory limit of 256Mi is sufficient for most use cases, but there are reports of OOM Kills which require to make this configurable.
+	// +optional
+	ConfigBuilderResources *corev1.ResourceRequirements `json:"configBuilderResources,omitempty"`
```
adutra, line 390: Instead of exposing another field in an already lengthy spec, shouldn't we instead detect that a provided init container has the `"server-config-init"` name, and if so, extract any resource requirements it may contain, and put them in the right place in the CassandraDatacenter spec?

Alternatively: I also wonder if this isn't a bug in cass-operator. In general it merges any provided containers with the default ones, but here indeed it's doing something different and ignoring the resource requirements completely.

adejanovski: > Instead of exposing another field in an already lengthy spec, shouldn't we instead detect that a provided init container has the "server-config-init" name, and if so, extract any resource requirements it may contain, and put them in the right place in the CassandraDatacenter spec?

Very good suggestion, let me try that

pkg/cassandra/datacenter.go
```diff
@@ -410,6 +415,10 @@ func AddInitContainersToPodTemplateSpec(dcConfig *DatacenterConfig, initContaine
 		return errors.New("PodTemplateSpec was nil, cannot add init containers")
 	} else {
 		dcConfig.PodTemplateSpec.Spec.InitContainers = append(dcConfig.PodTemplateSpec.Spec.InitContainers, initContainers...)
+		position, found := FindInitContainer(dcConfig.PodTemplateSpec, "server-config-init")
```
adutra, line 418: In the `reconciliation` package of cass-operator, there is an exported constant for this container's name: `reconciliation.ServerConfigContainerName`. It would be slightly better to use it instead of hard-coding the name.

adutra (approved):

PR #733, opened 2022-10-20
Enable client-side CQL encryption in Stargate if it is configured on the cluster (fixes #722)

### LINKED ISSUE #722: K8SSAND-1841 ⁃ Stargate client-to-node encryption allows unencrypted connections
It seems like enabling client to node encryption in k8ssandra-operator, and despite the encryption settings being passed to Stargate, it is still possible to use cqlsh without any encryption setting to connect to the Stargate service. The Cassandra service refuses similar connection attempts in this case.

This can be easily reproduced with the encryption with Stargate manifest we use for e2e test: `test/testdata/fixtures/single-dc-encryption-stargate/k8ssandra.yaml`

It requires to install first the following manifests to get the encryption stores:

* `test/testdata/fixtures/server-encryption-secret.yaml`
* `test/testdata/fixtures/client-encryption-secret.yaml`

Then ssh into one of the Cassandra pods and try both:

`cqlsh --username test-superuser --password <superuser password> test-dc1-service`
and
`cqlsh --username test-superuser --password <superuser password> test-dc1-stargate-service`

The former will fail but the latter will succeed.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1841) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1841
┆priority: Medium

olim7t:
**What this PR does**:
Generate a config file to enable client encryption for Stargate's CQL module.
Adapt e2e tests (thankfully NGINX ingress forwards TCP port 9042 the same whether it uses SSL or not, so the only thing to change is client code).

Fixes #722

**Checklist**
- [x] Changes manually tested


apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -43,7 +43,7 @@ type StargateTemplate struct {
 	// ContainerImage is the image characteristics to use for Stargate containers. Leave nil
 	// to use a default image.
 	// +optional
-	// +kubebuilder:default={repository:"stargateio", tag:"v1.0.45"}
+	// +kubebuilder:default={repository:"stargateio", tag:"v1.0.66"}
```
olim7t, line 46: `-Dstargate.cql.config_path` was introduced in Stargate 1.0.64. I went with the latest.

controllers/stargate/stargate_controller.go
```diff
@@ -398,42 +400,39 @@ func (r *StargateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 func (r *StargateReconciler) reconcileStargateConfigMap(
 	ctx context.Context,
 	stargateObject *api.Stargate,
-	desiredConfig map[string]interface{},
-	userConfigMapContent string,
+	dcConfig map[string]interface{},
+	userCassandraYaml, userCqlYaml string,
```
olim7t, line 404: A bit of naming/refactoring in this function, because it was getting really confusing with two files.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/apache/tinkerpop/gremlin-go v0.0.0-20220530191148-29272fa563ec
 	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/datastax/go-cassandra-native-protocol v0.0.0-20210829124742-a80a54434112
+	github.com/datastax/go-cassandra-native-protocol v0.0.0-20220706104457-5e8aad05cf90
```
olim7t, line 10: Pull in TLS support.

test/framework/cql.go
```diff
@@ -0,0 +1,83 @@
...
+		credentials = &client.AuthCredentials{Username: username, Password: password}
+	}
+	cqlClient := client.NewCqlClient(string(cqlEndpoint), credentials)
+
+	if ssl {
+		secret, err := f.getClientEncryptionSecret(ctx, k8sContext, namespace)
+		if err != nil {
+			return nil, err
+		}
+		rootCas, err := extractCaCertificates(secret)
```
olim7t, line 41: Alternatively we could extract the certificate manually once, commit it and just load that file.
Pros:
- this avoids the extra dependency to jceks (keystore management).

Cons:
- we'd have to keep the cert file and config map in sync, but the certs are valid until 2049...
- this is test code and it works so why change it

controllers/stargate/stargate_controller.go
```diff
@@ -133,15 +133,17 @@ func (r *StargateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	if stargate.Spec.CassandraConfigMapRef != nil {
 		userConfigMap := &corev1.ConfigMap{}
 		configMapKey := types.NamespacedName{Namespace: req.Namespace, Name: stargate.Spec.CassandraConfigMapRef.Name}
 		err := r.Get(ctx, configMapKey, userConfigMap)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		userConfigMapContent = userConfigMap.Data["cassandra.yaml"]
+		userStargateCassandraYaml = userConfigMap.Data["cassandra.yaml"]
+		userStargateCqlYaml = userConfigMap.Data[stargateutil.CqlConfigName]
```
adejanovski, line 146: I'm trying to think of a case where we'd get a user input that should go into that Stargate cql config file.
Which settings other than `client_encryption_options` have to be passed through this file?

pkg/stargate/config.go
```diff
@@ -37,12 +41,13 @@ func CreateStargateConfigMap(namespace, configYaml string, dc cassdcapi.Cassandr
...
 		Data: map[string]string{
-			"cassandra.yaml": configYaml,
+			"cassandra.yaml": cassandraYaml,
+			CqlConfigName:    stargateCqlYaml,
 		},
 	}
 }
 
-func MergeConfigMaps(userConfigMap string, generatedConfigMap string) string {
+func MergeYamlString(userConfigMap string, generatedConfigMap string) string {
```
adejanovski, line 50: yep, definitely more accurate 👍

test/framework/cql.go
adejanovski: Happy with what we have here, no need to change.

adejanovski (approved):
Awesome work @olim7t 👏 
Just one tiny question around the use case of externally provided cql yaml options to Stargate, which is non blocking for me. If there are specific settings other than `client_encryption_options` which should go there instead of cassandra.yaml, we'll need to document them.

olim7t:
> If there are specific settings other than `client_encryption_options` which should go there instead of cassandra.yaml, we'll need to document them.

Right, there might be more than the encryption options.
The full list supported by Stargate CQL is:
```
rpc_address
rpc_interface
rpc_interface_prefer_ipv6
rpc_keepalive *
native_transport_port
native_transport_port_ssl
native_transport_max_frame_size_in_mb *
native_transport_max_concurrent_connections *
native_transport_max_concurrent_connections_per_ip *
native_transport_flush_in_batches_legacy *
native_transport_allow_older_protocols *
native_transport_max_concurrent_requests_in_bytes_per_ip *
native_transport_max_concurrent_requests_in_bytes *
native_transport_idle_timeout_in_ms *
```
For the ones I marked with `*`: I think it make sense to copy to `stargate_cql.yaml`, I will add them to the list.

RPC address / interface: unlikely to be set globally on the `CassandraDatacenter`, if they are set to non-default values it's unlikely that it would be something global for all nodes.

Ports: the service hard-codes port 9042, and it doesn't allow both SSL and non-SSL either. So I don't think customizing those is a use case either.

PR #768, opened 2022-11-16
Fix removal of DC that is local to the control plane

### LINKED ISSUE #642: K8SSAND-1704 ⁃ Decommissioning datacenter does not work unless k8sContext is set
**What happened?**
Someone on Discord reported a problem with deleting a datacenter (see [here](https://discord.com/channels/836217371453685760/836217371453685763/1002205193879953439)). The replication of keyspaces is updated, removing the DC as expected. k8ssandra-operator reports that the deletion happens, but the CassandraDatacenter is neither decommissioned nor deleted.

For reference a decommissioning a DC is done simply by removing it from the K8ssandraCluster spec. On a subsequent reconciliation, k8ssandra-operator detects that the DC is removed by comparing the DCs in the spec to those in the status, and then starts the decommissioning process.

Removing a DC from the K8ssandraCluster should result in the following:

* The DC should be removed from the replication strategy of all keyspaces
* Every C* node in the DC should be decommissioned
* The CassandraDatacenter should be deleted

```yaml
kind: K8ssandraCluster
metadata:
  name: test
spec:
  cassandra:
    serverVersion: "4.0.3"
    storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
    config:
      jvmOptions:
        heapSize: 1Gi
    # Only needed when running a local k8s distro like kind
    networking:
      hostNetwork: true
    datacenters:
      - metadata:
          name: dc1
          namespace: dev1
        size: 1
      - metadata:
          name: dc2
          namespace: dev2
        size: 1
```

This example requires a cluster-scoped deployment, but the problem will happen with namespace-scoped deployments as well. I was too lazy to set up a multi-cluster environment :) 

**Environment**

* K8ssandra Operator version:

    `v1.2.0`
    
    `kubectl version`
The first thing that the `reconcileDatacenters` method does is to call the `checkDcDeletion` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/132f65c1a507ccfea605e21da2208107a93a9a4f/controllers/k8ssandra/datacenters.go#L34).

After replication settings are updated the `deleteDc` method is called [here](https://github.com/k8ssandra/k8ssandra-operator/blob/132f65c1a507ccfea605e21da2208107a93a9a4f/controllers/k8ssandra/cleanup.go#L147). This method first attempts to delete any Stargate and Reaper instances that are deployed with the DC. Then `findDcForDeletion` is called [here](https://github.com/k8ssandra/k8ssandra-operator/blob/132f65c1a507ccfea605e21da2208107a93a9a4f/controllers/k8ssandra/cleanup.go#L178).

The problem is in the `if` block:

```go
if remoteClient == nil {
	for _, remoteClient := range r.ClientCache.GetRemoteClients() {
		err := remoteClient.List(ctx, dcList, options)
		if err != nil {
			return nil, nil, fmt.Errorf("failed to CassandraDatacenter (%s) for DC (%s) deletion: %v", dcName, dcName, err)
		}
		for _, dc := range dcList.Items {
			if dc.Name == dcName {
				return &dc, remoteClient, nil
			}
		}
	}
```

We iterate over the remote clients from the ClientCache in search of the DC. We have to do this because we no longer know in which cluster the CassandraDatacenter was deployed. If the DC is in the local, control plane cluster and`k8sContext` is not set, then the DC won't be found. This is because `r.ClientCache.GetRemoteClients` does not include a client for the local, control plane cluster.

An easy fix for this is to also iterate over the local, control plane client, maybe introduce a method like `ClientCache.GetAllClients()`.

Lastly, the same iteration logic is used in the methods do delete Stargate and Reaper which means they won't be deleted either. The `findStargateForDeletion` and `findReaperForDeletion` methods need to be updated as well.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1704) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1704
┆priority: Medium

olim7t:
**What this PR does**:
Fix the DC deletion code so that it works even when the DC is "local", i.e. in the control plane itself.

Fixes #642

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/cleanup.go
+++ b/controllers/k8ssandra/cleanup.go
@@ -222,7 +222,7 @@ func (r *K8ssandraClusterReconciler) findStargateForDeletion(
 	stargateName := kcKey.Name + "-" + dcName + "-stargate"
 
 	if remoteClient == nil {
-		for _, remoteClient := range r.ClientCache.GetRemoteClients() {
+		for _, remoteClient := range r.ClientCache.GetAllClients() {
 			err := remoteClient.List(ctx, stargateList, options)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
@@ -261,7 +261,7 @@ func (r *K8ssandraClusterReconciler) findReaperForDeletion(
 	reaperName := kcKey.Name + "-" + dcName + "-reaper"
 
 	if remoteClient == nil {
-		for _, remoteClient := range r.ClientCache.GetRemoteClients() {
+		for _, remoteClient := range r.ClientCache.GetAllClients() {
 			err := remoteClient.List(ctx, reaperList, options)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to find Reaper (%s) for DC (%s) deletion: %v", reaperName, dcName, err)
@@ -298,7 +298,7 @@ func (r *K8ssandraClusterReconciler) findDcForDeletion(
 	dcList := &cassdcapi.CassandraDatacenterList{}
 
 	if remoteClient == nil {
-		for _, remoteClient := range r.ClientCache.GetRemoteClients() {
+		for _, remoteClient := range r.ClientCache.GetAllClients() {
 			err := remoteClient.List(ctx, dcList, options)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to CassandraDatacenter (%s) for DC (%s) deletion: %v", dcName, dcName, err)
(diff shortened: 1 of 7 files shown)


controllers/k8ssandra/cleanup.go
```diff
@@ -310,16 +310,14 @@ func (r *K8ssandraClusterReconciler) findDcForDeletion(
...
 	} else {
-		for _, remoteClient := range r.ClientCache.GetRemoteClients() {
-			err := remoteClient.List(ctx, dcList, options)
-			if err != nil {
-				return nil, nil, fmt.Errorf("failed to find CassandraDatacenter (%s) for deletion: %v", dcName, err)
-			}
+		err := remoteClient.List(ctx, dcList, options)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to find CassandraDatacenter (%s) for deletion: %v", dcName, err)
+		}
```
olim7t, line 316: Not directly related to the fix, but I think this was a typo. See `findStargateForDeletion` and `findReaperForDeletion`.

Makefile
```diff
@@ -343,7 +343,7 @@ install-helm:
 	./bin/get_helm.sh
 
 create-clientconfig:
-	for ((i = 0; i < $(NUM_CLUSTERS); ++i)); do \
+	for ((i = 1; i < $(NUM_CLUSTERS); ++i)); do \
 		scripts/create-clientconfig.sh \
 		  --namespace $(NS) \
 		  --src-kubeconfig "$(KIND_KUBECONFIG)" \
```
olim7t, line 349: Changed here but also in `e2e_framework.go`, it looks like the two are redundant.

adejanovski (approved):
Great stuff!

PR #772, opened 2022-11-18
Retry failed e2e tests

### LINKED ISSUE #762: K8SSAND-1886 ⁃ Stabilize Stargate e2e tests
Stargate e2e tests started getting flakier lately, both in single and multi DC modes.
We'd need to investigate those flakes and stabilize the tests to secure our CI runs.

olim7t:
Experimenting with an automatic retry when e2e tests fail.

Fixes #762 (kind of)


adejanovski (commented):
While this could give the impression that we fixed the flakes, it may be hiding some potential race conditions that were introduced in the tests recently (and possibly bugs).
Those tests were fairly stable not so long ago and we should probably investigate a little deeper to see what's causing the actual failures.

adejanovski:
Replaced by #782

PR #780, opened 2022-11-24
Use a merge library to merge cluster- and dc-level specs

### LINKED ISSUE #525: K8SSAND-1460 ⁃ Cluster- and dc-level templates should be deep-merged
**What happened?**

Consider the following `K8ssandraCluster`:

```yaml
kind: K8ssandraCluster
metadata:
  name: test
spec:
  # Cluster level
  stargate:
    size: 1
    heapSize: 512Mi
  cassandra:
    datacenters:
      - metadata:
          name: dc1
        size: 1
        # DC level
        stargate:
          size: 1
          racks:
            # Rack level
            - name: rack1
```

If I deploy this CRD, the observed heap size for Stargate nodes is `256Mi` (the default).

I would expect the heap size of all Stargate nodes to be `512Mi`.

This is because the `Coalesce` methods do not merge fields, they simply pick the first non-nil template. In the above example, the Rack template wins over the DC template and the cluster template, so it gets applied, and its heap size is 256Mi (the default).

Note that many e2e are wrongly conceived because of that. They try to set the heap size cluster-wide, yet it is not applied because there is a DC or Rack template somewhere.

Use the above CRD.

**Environment**

* K8ssandra Operator version: 1.0.0

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1460) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1460
┆priority: Medium

### LINKED ISSUE #715: K8SSAND-1821 ⁃ Ability to merge cluster- and dc- level configs
**What is missing?**

Currently cluster-level configs are coalesced with dc-level ones in an all-or-nothing fashion: if the dc-level config is present, then the cluster-level one is ignored.

**Why do we need it?**

With #710, YAML config files are now schemaless in the CRD and represented internally as `map[string]interface{}`. It should be fairly simple to use `utils.MergeMapNested` to merge cluster-level and dc-level configs.

Note: JVM options are still schemaful so we can't apply this fix to the `JvmOptions` struct. We might be able to write custom code to handle it, or resort to merging libraries as suggested in #526.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1821) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1821
┆priority: Medium

adutra:
**What this PR does**:

This PR uses Goalesce to do all the cluster/DC spec merges in the operator code.

Fixes #525 
Fixes #715 

**Checklist**
- [x] Changes manually tested


apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -411,6 +411,27 @@ type DatacenterOptions struct {
 	ManagementApiAuth *cassdcapi.ManagementApiAuthConfig `json:"managementApiAuth,omitempty"`
 }
 
+// NetworkingConfig is a copy of cass-operator's NetworkingConfig struct. It is copied here to
+// change the HostNetwork field type from bool to *bool, which makes merging 2 values of this struct
+// more intuitive.
+type NetworkingConfig struct {
```
adutra, line 417: This was a bit disappointing, but there is no way we can merge a `bool` field and expect trilean semantic, so I had to copy the struct here and change to `*bool`.

apis/stargate/v1alpha1/stargate_types.go
```diff
@@ -151,16 +152,20 @@ type StargateDatacenterTemplate struct {
 	Racks []StargateRackTemplate `json:"racks,omitempty"`
 }
 
-// Coalesce compares this StargateDatacenterTemplate with the given StargateClusterTemplate and returns the first
-// non-nil StargateDatacenterTemplate it finds.
-// TODO revisit the merging strategy and/or find a better name for this method
-func (in *StargateDatacenterTemplate) Coalesce(clusterTemplate *StargateClusterTemplate) *StargateDatacenterTemplate {
+// MergeWith merges the given StargateClusterTemplate with this StargateDatacenterTemplate and
+// returns the result.
+func (in *StargateDatacenterTemplate) MergeWith(clusterTemplate *StargateClusterTemplate) *StargateDatacenterTemplate {
```
adutra, line 157: I had to standardize method names and the general contract. I went for `MergeWith` and the contract is that the receiver is the DC level object, and the passed object is the cluster level one.

apis/telemetry/v1alpha1/telemetry_methods.go
```diff
@@ -1,49 +1,12 @@
...
-		}
-	}
-	if b.CommonLabels != nil {
-		for k, v := range b.CommonLabels {
-			out.CommonLabels[k] = v
-		}
-	}
-	out.Enabled = b.Enabled
-	return &out
+func (in *TelemetrySpec) IsPrometheusEnabled() bool {
```
adutra, line 10: This isn't  really related to merges but it simplifies the code a lot.

apis/telemetry/v1alpha1/telemetry_types.go
```diff
@@ -1,14 +1,15 @@
...
 
 type TelemetrySpec struct {
 	Prometheus *PrometheusTelemetrySpec `json:"prometheus,omitempty"`
 	Mcac       *McacTelemetrySpec       `json:"mcac,omitempty"`
 }
+
 type PrometheusTelemetrySpec struct {
 	// Enable the creation of Prometheus serviceMonitors for this resource (Cassandra or Stargate).
-	Enabled bool `json:"enabled,omitempty"` // A bool flag required here to disambiguate when e.g. the cluster should have telemetry turned on but one DC should have it explicitly turned off.
+	Enabled *bool `json:"enabled,omitempty"`
```
adutra, line 12: Change required for trilean logic. This has no impact on the CRD definition.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Note that this has no impact on the CRD definition.

docs/arch-overview.md
```diff
@@ -36,15 +36,15 @@ cassandra:
       size: 3  
 ``` 
 
-There are two elements in the array. K8ssnandra Operator will create CassandraDatacenters named `dc1` and `dc2`. Each one will have a size of 3.
+There are two elements in the array. K8ssandra Operator will create CassandraDatacenters named `dc1` and `dc2`. Each one will have a size of 3.
```
adutra, line 39: I didn't change the contents of this file, just fixed the typos.

pkg/goalesce/utils.go
```diff
@@ -0,0 +1,66 @@
+package goalesceutils
+
+import (
+	"github.com/adutra/goalesce"
```
adejanovski, line 4: ahem... can we fork the goalesce project into the k8ssandra org and reference that version instead?

adutra: As we discussed offline, this is a project that I started outside K8ssandra, and I intend to maintain it. Of course you could fork it if you want to, but at this point, I don't see any valid reason to do so.

adejanovski: sounds good 👍

adejanovski (approved):

Issue #532, opened 2022-04-26
K8SSAND-1476 ⁃ Revisit default values for images.Image struct


adutra:
`Image` currently has a rather complex set of rules to determine default values for its fields. It turns out that the defaults for almost all fields can be guessed by Kubernetes, e.g. if no tag and no pull policy is provided, then the former defaults to `latest` and the latter to `Always`. Also, the registry and the repository are also automatically defaulted to `docker.io` and `library` respectively.

This logic is consequently not necessary at our level. We should be able to remove the logic altogether, but we might need to check how definitive image names are computed and that they remain valid.

On a side note, the current struct doesn't accommodate very nicely tags that are actually image hashes. See [here](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) for the format syntax. We may want to improve that too.

## Update

After discussing it, we'd like to revisit how images are referenced and allow customizing registries, pull secrets, ... by using cass-operator's ImageConfig.
This would require to:

- Replace all occurrences of images.Image in our CRDs and use a string instead (this means we'll have to think about upgrades and possibly introduce a conversion webhook)
- See if we can rely on the same ImageConfig entry that is used by cass-operator
- Evolve the format of the ImageConfig CRD to allow setting a dynamic list of images, each with their own pull secrets/registries/...

```
apiVersion: config.k8ssandra.io/v1beta1
kind: ImageConfig
metadata:
  name: image-config
images:
  system-logger:
     image: "k8ssandra/system-logger:latest"
     imageRegistry: ...
     imagePullSecret:
       name: ....
  config-builder:
     image: "datastax/cass-config-builder:1.0.4-ubi7"
  medusa:
     image: "....."
     imageRegistry: ...
     imagePullSecret:
       name: ....
  # Default settings unless overridden at the image level
  imageRegistry: "localhost:5000"
  imagePullPolicy: Always
  imagePullSecret:
    name: my-secret-pull-registry
defaults:
  # Note, postfix is ignored if repository is not set
  cassandra:
    repository: "k8ssandra/cass-management-api"
  dse:
    repository: "datastax/dse-mgmtapi-6_8"
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-272) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-272

jsanda:
>Also, the registry and the repository are also automatically defaulted to docker.io 

[...]

PR #786, opened 2022-12-01
Expose perNodeConfigInitContainerImage through CRD

### LINKED ISSUE #781: Expose perNodeConfigInitContainer.Image through CRD
**What is missing?**
Now image for per-node configuration hardcoded in https://github.com/k8ssandra/k8ssandra-operator/blob/d58b87de760da5eb10c8579898d0da303f3a7914/pkg/nodeconfig/mount.go#L29

Why dont we expose image config  like JmxInitContainerImage with default kubebuilder annotation?
https://github.com/k8ssandra/k8ssandra-operator/blob/06a89fc154384bb77661c1e655e376c0353ab72b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go#L343

It will be user get the ability to override own registry,repo,tag

**Why do we need it?**
Many installation restricts public access, for example, we have admission controller that cheks all images are pulled only from private registry

**Environment**

* K8ssandra Operator version: 1.4.0

dnugmanov:
**What this PR does**:
Expose PerNodeInitContainerImage through CRD

Fixes #781 

**Checklist**
- [x] Changes manually tested
- [x] Documentation added/updated - CRD description


apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -409,6 +409,13 @@ type DatacenterOptions struct {
 	// ManagementApiAuth defines the authentication settings for the management API in the Cassandra pods.
 	// +optional
 	ManagementApiAuth *cassdcapi.ManagementApiAuthConfig `json:"managementApiAuth,omitempty"`
+
+	// The image to use in each Cassandra pod for the (short-lived) init container that merges global and perNodeConfig
+	// configuration. This is only useful when PerNodeConfigMapRef is set.
+	// The default is "mikefarah/yq:4".
+	// +optional
+	// +kubebuilder:default={repository:"mikefarah",name:"yq",tag:"4"}
+	PerNodeInitContainerImage *images.Image `json:"perNodeInitContainerImage,omitempty"`
```
adutra, line 418: Hi @studypyth thanks for contributing to the K8ssandra project!

Two remarks on that line:

1. I suppose you declared this field here because, contrary to `PerNodeConfigMapRef`, you would like it to be declarable at both cluster and DC level? If that's why, I think that's fair, and a good idea.
2. However you used `images.Image` here, probably taking `JmxInitContainerImage` as a model. But we want to get rid of `images.Image` because it's too clunky, see https://github.com/k8ssandra/k8ssandra-operator/issues/532 for details. Given that, would you mind replacing `images.Image` with just a string?

adutra: Also, I'm nitpicking but I'd prefer that you name your field `perNodeConfigInitContainerImage` – because otherwise the name is ambiguous and would mean "the image to use for per-node init-containers".

go.sum
```diff
@@ -343,7 +343,6 @@ github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.3/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
-github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
```
adutra, line 346: These changes are unrelated and probably due to a `go mod tidy`? It's OK to leave them though.

pkg/cassandra/datacenter.go
```diff
@@ -378,6 +379,12 @@ func Coalesce(clusterName string, clusterTemplate *api.CassandraClusterTemplate,
 		dcConfig.ManagementApiAuth = clusterTemplate.DatacenterOptions.ManagementApiAuth
 	}
 
+	if dcTemplate.DatacenterOptions.PerNodeInitContainerImage != nil {
```
adutra, line 382: Beware that when #780 is going to be merged, there will be a conflict here. I can help you solve the conflict if need be.

adutra (changes requested):

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
dnugmanov: Hi, @adutra, thank you for review.
1. Yes, exactly
2. I think that image.Image is the target type  according to medusa, stargate, jmxInitContainer. I have changed type to string, pls check.

pkg/cassandra/datacenter.go
dnugmanov: Thank you for your comments, I have rebased commit on current main branch

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
adutra: Thanks for the change. We need to unify the way we handle images in K8ssandra. The team will discuss this topic shortly and hopefully we'll arrive to a good, common way of doing that. In the meantime, string is perfect 👍

adutra (approved):

dnugmanov:
@adutra, could you pls merge PR? i don't have enough rights

adutra:
@dnugmanov thanks again! Before I merge, one question: the change from `images.Image` -> `string` will make it impossible for you to use pull secrets to pull the per-node config image from a private registry. Is that OK for your use case? This will be possible later, when we will revisit the mechanism to specify images throughout K8ssandra; but in the meantime, you won't be able to do that.

dnugmanov:
> @dnugmanov thanks again! Before I merge, one question: the change from `images.Image` -> `string` will make it impossible for you to use pull secrets to pull the per-node config image from a private registry. Is that OK for your use case? This will be possible later, when we will revisit the mechanism to specify images throughout K8ssandra; but in the meantime, you won't be able to do that.

I am not using `pullSecretRef` field from `images.Image`, i configured `ImageConfig` to setup correct secretRef
```
    apiVersion: config.k8ssandra.io/v1beta1
    kind: ImageConfig
    imagePullSecret:
      name: my-secret-pull-registry
```

PR #788, opened 2022-12-05
configurable labels & annotations

### LINKED ISSUE #501: K8SSAND-1430 ⁃ Allow configuring annotations and labels on services, statefulsets, deployments and pods
**What is missing?**

Annotations and labels cannot be configured on services, statefulsets, deployments and pods created by the operator.

**Why do we need it?**

The annotations and labels are useful in many ways, for example: exposing a service using [external-dns](https://github.com/kubernetes-sigs/external-dns), tagging logs from a pod so they can indexed and searched appropriately.

**Environment**

* K8ssandra Operator version: v1.0.1

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1430) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1430
┆priority: Medium

sseidman:
**What this PR does**:
Adds ResourceMeta and MetaTags to each component that allows a user to set annotations for the following:
- Placing labels & annotations on pods;
- Placing labels & annotations on services;
- Placing labels on other types of objects, including custom resources.

example yaml:
```
  cassandra:
    metadata:
      name: demo
      commonLabels:
      labels:
      pods:
        labels:
        annotations:
      services:
        dcService:
          annotations:
          labels:
  reaper:
    metadata:
      commonLabels:
      labels:
      annotations:
      pods:
        labels:
        annotations:
      service:
        labels:
        annotations:
  stargate:
    metadata:
      commonLabels:
      labels:
      annotations:
      pods:
        labels:
        annotations:
      service:
        labels:
        annotations:
```

Fixes #501 

**Checklist**
- [x] Changes manually tested


pkg/meta/meta.go
```diff
@@ -0,0 +1,55 @@
+package meta
+
+import "github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+
+// +kubebuilder:object:generate=true
+type MetaTags struct {
```
adutra, line 6: Nit: `Tags` would be enough since the package is named `meta` already.

pkg/cassandra/datacenter.go
```diff
@@ -192,6 +193,21 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
 		dc.Spec.ManagementApiAuth = *template.ManagementApiAuth
 	}
 
+	if m := template.Meta.Metadata; m != nil {
```
adutra, line 196: This should be done by the merging library we are using:

```go
dcConfig.Meta = goalesceutils.MergeCRs(clusterTemplate.Meta, dcTemplate.Meta)
```

pkg/cassandra/datacenter.go
```diff
@@ -309,6 +325,13 @@ func Coalesce(clusterName string, clusterTemplate *api.CassandraClusterTemplate,
 	dcConfig.PodTemplateSpec.Spec.SecurityContext = mergedOptions.PodSecurityContext
 	dcConfig.PerNodeInitContainerImage = mergedOptions.PerNodeConfigInitContainerImage
 
+	m := api.MergeCassandraDatacenterMeta(clusterTemplate.Meta.Metadata, dcTemplate.Meta.Metadata)
```
adutra, line 328: Same here, you don't need a specific function to merge these objects.

pkg/meta/meta.go
```diff
@@ -0,0 +1,55 @@
...
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
+
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// merge MetaTags into a destination MetaTags. If properties are shared between the two configs,
+// the destination MetaTags property will be used
+func MergeMetaTags(m1 *MetaTags, m2 *MetaTags) *MetaTags {
```
adutra, line 16: This function should go away when we switch to using Goalesce to merge the meta tags.

pkg/meta/meta.go
```diff
@@ -0,0 +1,55 @@
...
+	// created by the CRD
+	// +optional
+	CommonLabels map[string]string `json:"commonLabels,omitempty"`
+
+	// labels/annotations for the pod components
+	// +optional
+	Pods *MetaTags `json:"pods,omitempty"`
+
+	// labels/annotations for the service component
+	Service *MetaTags `json:"service,omitempty"`
```
adutra, line 54: I wonder if we shouldn't call it `services` in the plural, for similarity with the `pods` field? But I understand you probably used the singular because we expect only one service to be created. I'm hesitating. Wdyt?

pkg/meta/meta.go
```diff
@@ -0,0 +1,55 @@
...
+		Annotations: utils.MergeMap(a1, a2),
+	}
+}
+
+// Struct to hold labels and annotations for a resource
+// +kubebuilder:object:generate=true
+type ResourceMeta struct {
+	// labels/annotations for the top-level CRD component
+	// +optional
+	Resource *MetaTags `json:"resource,omitempty"`
```
adutra, line 42: Nit. I'm not sure the pointer type buys us a lot here, since the `MetaTags` type only contains map fields, that are nil by default anyways. But we do need to check for the nilness of these fields now. Maybe `Resource MetaTags` is enough?

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -457,26 +461,84 @@ type EmbeddedObjectMeta struct {
 	Name string `json:"name"`
 
 	// +optional
-	Labels map[string]string `json:"labels,omitempty"`
+	Metadata *CassandraDatacenterMeta `json:"dcMetadata,omitempty"`
+}
 
+type CassandraDatacenterMeta struct {
```
adutra, line 467: I would move this type to the `meta` package as well. But wouldn't fight for it :-)

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -457,26 +461,84 @@ type EmbeddedObjectMeta struct {
...
+		Pods:          meta.MergeMetaTags(m1.Pods, m2.Pods),
+		ServiceConfig: mergeServiceConfig(m1.ServiceConfig, m2.ServiceConfig),
+	}
 }
 
-func (in *EmbeddedObjectMeta) SetLabels(labels map[string]string) {
-	in.Labels = labels
+// merge a ServiceConfig into a destination ServiceConfig. If properties are shared between the two configs,
+// the destination ServiceConfig property will be used
+func mergeServiceConfig(s1 *cassdcapi.ServiceConfig, s2 *cassdcapi.ServiceConfig) *cassdcapi.ServiceConfig {
```
adutra, line 515: Again, not required. Use `goalesceutils`.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -225,6 +227,8 @@ type K8ssandraClusterList struct {
 type CassandraClusterTemplate struct {
 	DatacenterOptions `json:",inline"`
 
+	Meta EmbeddedObjectMeta `json:"metadata,omitempty"`
```
adutra, line 230: This is exposing `name ` and `namespace` as well at cluster-level, which imho doesn't make sense.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -457,26 +461,84 @@ type EmbeddedObjectMeta struct {
 	Name string `json:"name"`
 
 	// +optional
-	Labels map[string]string `json:"labels,omitempty"`
+	Metadata *CassandraDatacenterMeta `json:"dcMetadata,omitempty"`
```
adutra, line 464: It would be nice to not have both `metadata` and `dcMetadata` here. With a bit of refactoring, I think this is achievable:

```go
type EmbeddedObjectMeta struct {
	Namespace string `json:"namespace,omitempty"`
	Name string `json:"name"`
	CassandraDatacenterMeta `json:",inline"`
}
```

The beauty of it is that you can then expose only `CassandraDatacenterMeta` at cluster-level, and the whole `EmbeddedObjectMeta` at DC level. This solves the issue of not exposing `name` and `namespace` at cluster level:

```go
type CassandraClusterTemplate struct {
	DatacenterOptions `json:",inline"`
	Meta CassandraDatacenterMeta `json:"metadata,omitempty"`
	// ...
}

type CassandraDatacenterTemplate struct {
	DatacenterOptions `json:",inline"`
	Meta EmbeddedObjectMeta `json:"metadata,omitempty"`
        // ...
}
```

Merging the 2 fields can be done this way:

```go
	cassandraDatacenterMeta := goalesceutils.MergeCRs(clusterTemplate.Meta, dcTemplate.Meta.CassandraDatacenterMeta)
	dcConfig.Meta.CassandraDatacenterMeta = cassandraDatacenterMeta
```

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -457,26 +461,84 @@ type EmbeddedObjectMeta struct {
...
+	CommonLabels map[string]string `json:"commonLabels,omitempty"`
+
+	// labels/annotations for the pod components
+	// +optional
+	Pods *meta.MetaTags `json:"pods,omitempty"`
 
-func (in *EmbeddedObjectMeta) SetAnnotations(annotations map[string]string) {
-	in.Annotations = annotations
+	// labels/annotations for all of the CassandraDatacenter service components
+	ServiceConfig *cassdcapi.ServiceConfig `json:"services,omitempty"`
```
adutra, line 482: So, using `ServiceConfig` directly is easy, but it exposes fields like `additionalLabels` and `additionalAnnotations`. It would be nicer to reuse `MetaTags`. I think we can achieve that with something like this:

```go

// CassandraDatacenterServicesMeta is very similar to cassdcapi.ServiceConfig and is passed to
// cass-operator in the AdditionalServiceConfig field of the CassandraDatacenter spec.
// +kubebuilder:object:generate=true
type CassandraDatacenterServicesMeta struct {
	DatacenterService     MetaTags `json:"dcService,omitempty"`
	SeedService           MetaTags `json:"seedService,omitempty"`
	AllPodsService        MetaTags `json:"allPodsService,omitempty"`
	AdditionalSeedService MetaTags `json:"additionalSeedService,omitempty"`
	NodePortService       MetaTags `json:"nodePortService,omitempty"`
}

func (in CassandraDatacenterServicesMeta) ToCassAdditionalServiceConfig() cassdcapi.ServiceConfig {
	return cassdcapi.ServiceConfig{
		DatacenterService: cassdcapi.ServiceConfigAdditions{
			Annotations: in.DatacenterService.Annotations,
			Labels:      in.DatacenterService.Labels,
		},
		SeedService: cassdcapi.ServiceConfigAdditions{
			Annotations: in.SeedService.Annotations,
			Labels:      in.SeedService.Labels,
		},
		AdditionalSeedService: cassdcapi.ServiceConfigAdditions{
			Annotations: in.AdditionalSeedService.Annotations,
			Labels:      in.AdditionalSeedService.Labels,
		},
		AllPodsService: cassdcapi.ServiceConfigAdditions{
			Annotations: in.AllPodsService.Annotations,
			Labels:      in.AllPodsService.Labels,
		},
		NodePortService: cassdcapi.ServiceConfigAdditions{
			Annotations: in.NodePortService.Annotations,
			Labels:      in.NodePortService.Labels,
		},
	}
}
```

Wdyt?

adutra (changes requested):
Thanks a lot for putting this up! I'm leaving a first round of review for now, will continue later today.

pkg/meta/meta.go
sseidman: I think `service` makes sense since we only ever expect 1 for now. If there are multiple services, it would likely break out into a separate struct similar to the CassandraDatacenter meta. I also don't mind switching to `services` if needed

adutra: True, I think we can keep `service`.

pkg/reaper/deployment_test.go
```diff
@@ -33,8 +35,20 @@ func TestNewDeployment(t *testing.T) {
 			Name: "truststore-secret",
 		}},
 	}
+	reaper.Spec.ResourceMeta = &meta.ResourceMeta{
```
adutra, line 38: Could you add a test case that verifies that a label present in both `CommonLabels` and `Pods.Labels`  gets correctly merged? Same for `CommonLabels` + `Service.Labels`.

pkg/stargate/deployments_test.go
```diff
@@ -43,7 +44,21 @@ var (
 		Spec: api.StargateSpec{
 			DatacenterRef: corev1.LocalObjectReference{Name: dc.Name},
 			StargateDatacenterTemplate: api.StargateDatacenterTemplate{
-				StargateClusterTemplate: api.StargateClusterTemplate{Size: 1},
+				StargateClusterTemplate: api.StargateClusterTemplate{
```
adutra, line 47: Same here: we could add some overlapping `CommonLabels` to spice up the tests.

pkg/stargate/service.go
```diff
@@ -53,3 +50,44 @@ func NewService(stargate *api.Stargate, dc *cassdcapi.CassandraDatacenter) *core
 	annotations.AddHashAnnotation(service)
 	return service
 }
+
+func createServiceLabels(stargate *api.Stargate) map[string]string {
```
adutra, line 54: Same here, the function is unused.

pkg/stargate/service.go
```diff
@@ -53,3 +50,44 @@ func NewService(stargate *api.Stargate, dc *cassdcapi.CassandraDatacenter) *core
...
+		coreapi.ComponentLabel: coreapi.ComponentLabelValueStargate,
+		coreapi.CreatedByLabel: coreapi.CreatedByLabelValueStargateController,
+		api.StargateLabel:      stargate.Name,
+	}
+
+	var annotations map[string]string
+	if meta := stargate.Spec.ResourceMeta; meta != nil {
+		labels = utils.MergeMap(labels, meta.CommonLabels)
+
+		labels = utils.MergeMap(labels, meta.Service.Labels)
```
adutra, line 83:
```suggestion
		labels = utils.MergeMap(labels, meta.CommonLabels, meta.Service.Labels)
```

pkg/stargate/stargate.go
```diff
@@ -45,19 +46,27 @@ func NewStargate(
 		cassandraEncryption.ServerEncryptionStores = kc.Spec.Cassandra.ServerEncryptionStores
 	}
 
+	labels := map[string]string{
```
adutra, line 49: Could we extract this to a new `createResourceMeta` function like you did in some other places?

adutra (changes requested):
Impressive work, thanks! We are close to an approval, just a few more things to fix :-)

pkg/meta/meta.go
```diff
@@ -0,0 +1,88 @@
...
+	// labels/annotations for the service component
+	Service Tags `json:"service,omitempty"`
+}
+
+// Struct to hold labels and annotations for a CassandraDatacenter
+// +kubebuilder:object:generate=true
+type CassandraDatacenterMeta struct {
+	// labels/annotations for the CassandraDatacenter component
+	// +optional
+	Resource Tags `json:"resource,omitempty"`
```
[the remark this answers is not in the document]
sseidman: I might be missing something, but I think something goes wrong with the serialization when I try and `inline` here and in `ResourceMeta`. If I apply this yaml:
```
kind: K8ssandraCluster
metadata:
  name: demo
spec:
  cassandra:
    metadata:
      commonLabels:
        commonlabel: "everywhere"
      annotations:
        test: "test"
      labels:
        test: "other"
        some-tag: "some-val"
  reaper:
    metadata:
      commonLabels:
        reaper.common: "every"
      labels:
        deployment-label: "some-value"
      annotations:
        deployment-annotation: "some-ann"
```
neither the `Reaper` or `CassandraDatacenter` objects get the annotations/labels

pkg/cassandra/datacenter.go
```diff
@@ -192,6 +193,16 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
 
+	m := template.Meta.Metadata
+	dc.ObjectMeta.Labels = utils.MergeMap(dc.ObjectMeta.Labels, m.Resource.Labels)
+	dc.ObjectMeta.Annotations = utils.MergeMap(dc.ObjectMeta.Annotations, m.Resource.Annotations)
+
+	if m.CommonLabels != nil {
+		dc.Spec.AdditionalLabels = m.CommonLabels
+	}
+
+	dc.Spec.AdditionalServiceConfig = m.ServiceConfig.ToCassAdditionalServiceConfig()
```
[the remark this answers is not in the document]
sseidman: these labels/annotations are necessary for the CassandraDatacenter object that is being created; that is, taking the merged dcTemplate/clusterTemplate and adding to these:
```
	dc := &cassdcapi.CassandraDatacenter{
		ObjectMeta: metav1.ObjectMeta{
			Namespace:   namespace,
			Name:        template.Meta.Name,
			Annotations: map[string]string{},
			Labels: map[string]string{
				api.NameLabel:                      api.NameLabelValue,
				api.PartOfLabel:                    api.PartOfLabelValue,
				api.ComponentLabel:                 api.ComponentLabelValueCassandra,
				api.CreatedByLabel:                 api.CreatedByLabelValueK8ssandraClusterController,
				api.K8ssandraClusterNameLabel:      klusterKey.Name,
				api.K8ssandraClusterNamespaceLabel: klusterKey.Namespace,
			}, ...

```

pkg/reaper/deployment_test.go
sseidman: with [this ](https://github.com/k8ssandra/k8ssandra-operator/pull/788#discussion_r1053154792) change, we'll delegate adding the common labels to cass-operator. Added some `commonLabel` fields in existing tests to show that.

pkg/cassandra/datacenter.go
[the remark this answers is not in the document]
adutra: Ah right, sorry. Please disregard my comment.

pkg/reaper/deployment_test.go
adutra: In `CassandraDatacenter` resources, yes, but not for `Reaper` and `Stargate` resources: for these, we do the overlay. The current test design is fine, as it shows that deployment will have `override -> commonLevel` but pods will have `override -> podLevel`.

pkg/reaper/deployment_test.go
```diff
@@ -515,3 +542,33 @@ func TestCustomResources(t *testing.T) {
...
+	deployment := NewDeployment(reaper, newTestDatacenter(), nil, nil)
+
+	deploymentLabels := map[string]string{
+		k8ssandraapi.NameLabel:      k8ssandraapi.NameLabelValue,
+		k8ssandraapi.PartOfLabel:    k8ssandraapi.PartOfLabelValue,
+		k8ssandraapi.ComponentLabel: k8ssandraapi.ComponentLabelValueReaper,
+		k8ssandraapi.ManagedByLabel: k8ssandraapi.NameLabelValue,
+		reaperapi.ReaperLabel:       reaper.Name,
+		k8ssandraapi.CreatedByLabel: k8ssandraapi.CreatedByLabelValueReaperController,
+		"common":                    "everywhere", "override": "commonLevel",
```
adutra, line 557:
```suggestion
		"common":                    "everywhere",
		"override":                  "commonLevel",
```

pkg/reaper/service.go
```diff
@@ -40,7 +49,7 @@ func NewService(key types.NamespacedName, reaper *api.Reaper) *corev1.Service {
 					StrVal: "admin",
 				},
 			}},
-			Selector: labels,
+			Selector: getConstantLabels(reaper),
```
adutra, line 52: I think this is changing the semantics slightly to limit the selector labels to just the constant ones, right? I was a bit wary initially but I saw that you added more assertions in `TestNewService` and the results seem OK, so 👍 for the change.

pkg/meta/meta.go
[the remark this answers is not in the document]
adutra: Ah right, the joys of JSON unmarshiling... you need to embed the field for this to work. See https://go.dev/play/p/S9U0R9cwwAW.

```suggestion
	Tags `json:",inline`
```

adutra (changes requested):
Thanks for the latest changes. We are close to the approval, just need fixing the embedded/inline fields issue.

adutra (approved):
Thank you @sseidman !

PR #794, opened 2022-12-13
Add support for deploying the Vector Agent

### LINKED ISSUE #775: Add the ability to inject and configure a Vector agent sidecar in the Cassandra pods
The K8ssandraCluster API should be modified to include an option to deploy Vector as ~~a daemonset~~ an agent.

The ~~daemonset~~ agent should be configured to scrape the cassandra/reaper/stargate endpoints depending on which spec it's in.

The ~~K8ssandraCluster~~ telemetry spec should be augmented with a `.vector` spec which allows to enable and configure Vector:

```
type TelemetrySpec struct {
	Vector *VectorSpec `json:"vector,omitempty"`
	Mcac       *McacTelemetrySpec       `json:"mcac,omitempty"`
}

type PrometheusTelemetrySpec struct {
	// Enable the creation of Prometheus serviceMonitors for this resource (Cassandra or Stargate).
	Enabled *bool `json:"enabled,omitempty"`
	// CommonLabels are applied to all serviceMonitors created.
	// +optional
	CommonLabels map[string]string `json:"commonLabels,omitempty"`
}

type VectorSpec struct {
	Enabled *bool `json:"enabled,omitempty"`
        Config corev1.LocalObjectReference `json:"config,omitempty"`
}
```

The config part should only be composed of the sink, as the source will be set by the operator.
The sink can be a Prometheus instance or a local file (in an emptyDir) for testing purposes.

adejanovski:
**What this PR does**:
Upgrades the telemetry spec, adding support for deploying Vector as an agent to scrape Cassandra metrics.
The Vector source configuration is set up to scrape the metrics endpoint of the Cassandra container, and the sinks can be customized through a configmap to forward the metrics anywhere Vector supports.

A little refactoring was made to fix what appears to be a bug in per node config cleanups. The configmaps get the "Part Of" labels but the cleanup selector was using the "Managed By" ones. I made it so that it's using Managed By ones in both cases.
Since the selector deals with all config maps, I made the code generic and it will effectively cleanup any config map that is managed by K8ssandra upon deletion of the K8ssandraCluster object.

Fixes #775 

**Checklist**
- [x] Changes manually tested


emerkle826 (approved):
1 Typo, and maybe a slight rearrangement in the docs. Otherwise tested and looks good. :+1:

docs/vector/README.md
```diff
@@ -0,0 +1,138 @@
...
+metadata:
+  name: test
+spec:
+  cassandra:
+    serverVersion: 4.0.6
+    telemetry:
+      vector:
+        enabled: true
+        config:
+          name: vector-custom-conf
```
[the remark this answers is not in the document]
adejanovski: very good point, let me re-arrange.

docs/vector/README.md
```diff
@@ -0,0 +1,138 @@
...
+metadata:
+  name: vector-custom-conf
+data:
+  vector.toml: |
+    [sinks.void]
+    type = "blackhole"
+    inputs = [ "cassandra_metrics" ]
+```
+
+The above configuration will discard all the metrics scraped from Cassandra and replace the defaul console sink.
```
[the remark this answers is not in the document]
adejanovski: 👍

apis/telemetry/v1alpha1/telemetry_types.go
```diff
@@ -15,6 +20,32 @@ type PrometheusTelemetrySpec struct {
...
+
+	// Resources is the resource requirements for the Vector agent.
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Image is the name of the Vector image to use. If not set, the default image will be used.
+	// +optional
+	// kube:default="timberio/vector:0.26.0-alpine"
+	Image string `json:"image,omitempty"`
+
+	// ScrapeInterval is the interval at which the Vector agent will scrape the metrics endpoint.
```
adutra, line 43: Would be nice to specify which time unit to use (I imagine seconds?).

adutra: Also, in meta/v1 there is a Duration type:

```go
// Duration is a wrapper around time.Duration which supports correct
// marshaling to YAML and JSON. In particular, it marshals into strings, which
// can be used as map keys in json.
type Duration struct {
	time.Duration `protobuf:"varint,1,opt,name=duration,casttype=time.Duration"`
}
```

controllers/k8ssandra/vector.go
```diff
@@ -0,0 +1,71 @@
...
+) result.ReconcileResult {
+	if kc.Spec.Cassandra.Telemetry.IsVectorEnabled() {
+		kcKey := utils.GetKey(kc)
+		namespace := utils.FirstNonEmptyString(dcConfig.Meta.Namespace, kc.Namespace)
+		configMapKey := client.ObjectKey{
+			Namespace: namespace,
+			Name:      telemetry.VectorAgentConfigMapName(kc.SanitizedName()),
+		}
+		// Create the vector toml config content
+		toml, err := telemetry.CreateCassandraVectorToml(ctx, kc.Spec.Cassandra.Telemetry, dcConfig, remoteClient, namespace)
```
adutra, line 34: The parameter `dcConfig` is unused.

pkg/telemetry/vector.go
```diff
@@ -0,0 +1,181 @@
...
+  codec = "json"`
+
+	if telemetrySpec.Vector.Config != nil {
+		// Read the Vector provided config map content and use it as the Vector sink config
+		vectorConfigMap := &corev1.ConfigMap{}
+		err := remoteClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: telemetrySpec.Vector.Config.Name}, vectorConfigMap)
+		if err != nil {
+			return "", err
+		} else {
+			sinks = vectorConfigMap.Data["vector.toml"]
```
adutra, line 118: You might want to test whether the key exists, otherwise you would be assigning an empty string to `sinks`.

controllers/k8ssandra/vector.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := remoteClient.Get(ctx, configMapKey, actualVectorConfigMap); err != nil {
+			if errors.IsNotFound(err) {
+				if err := remoteClient.Create(ctx, desiredVectorConfigMap); err != nil {
+					dcLogger.Error(err, "Failed to create Vector Agent ConfigMap")
+					return result.Error(err)
+				}
+			}
+		}
+
+		actualVectorConfigMap = actualVectorConfigMap.DeepCopy()
```
adutra, line 55: 1. I think there is a flaw here: if the cm doesn't exist and is created line 48, `actualVectorConfigMap` will be empty here, so the check line 57 will fail and an unnecessary update will be done line 61.
2. Also, you might want to handle some corner cases such as `apierrors.IsAlreadyExists`.
3. You also might want to handle the deletion case, that is, if vector was enabled, then disabled, the configmap certainly exists but we need to delete it.

I suggest that you take a look at the `reconcileDefaultPerNodeConfiguration` method since the situation is similar.

docs/vector/README.md
```diff
@@ -0,0 +1,161 @@
...
+            cpu: 100m
+            memory: 128Mi
+          limits:
+            cpu: 1000m
+            memory: 1Gi
+...
+...
+```
+
+The following content will be added automatically to the vector.toml file, setting up Cassandra metrics scraping:
```
adutra, line 33: Suggestion: place this paragraph after the following one on line 47. It is more straightforward to first explain how to pass a configmap containing the desired sinks, then explain how the operator will use the provided sinks and which sources it will create. Wdyt?

docs/vector/README.md
```diff
@@ -0,0 +1,161 @@
+# Using Vector with k8ssandra-operator
+
+Since k8ssandra-operator v1.5.0, it is possible to scrape Cassandra metrics with the Vector agent and customize its configuration to add transformers and sinks.
+
+More information about Vector can be found in the [official documentation](https://vector.dev/docs/).
+
+## Enabling Vector agent
+
+To enable Vector agent, you need to add a `.spec.cassandra.telemetry` section in the `K8ssandraCluster` manifest:
```
adutra, line 9: Note: as for other pieces in the `telemetry` section, you can configure Vector cluster-wide and override things at DC level. Both levels will be merged in the `Coalesce` function.

docs/vector/README.md
```diff
@@ -0,0 +1,161 @@
...
+target = "stdout"
+
+  [sinks.console.encoding]
+  codec = "json"`
+```
+
+
+## Custom Vector configuration
+
+To customize the Vector configuration, you need to create a `ConfigMap` containing a `vector.toml` file. The `ConfigMap` must be in the same namespace as the `K8ssandraCluster` resource:
```
adutra, line 62: From what I see in the `reconcileVector` method, it needs to be in the same namespace as the datacenter, which in turn defaults to the `K8ssandraCluster` namespace. But these can be different in a multi-namespace deployment.

docs/vector/README.md
```diff
@@ -0,0 +1,161 @@
...
+          requests:
+            cpu: 100m
+            memory: 128Mi
+          limits:
+            cpu: 1000m
+            memory: 1Gi
+...
+```     
+
+In case of multi DC deployment, the configmap needs to exist in each cluster and namespace where a Cassandra Datacenter is deployed.
```
adutra, line 103: Did you mean "multi-cluster deployment"?

pkg/telemetry/vector.go
```diff
@@ -0,0 +1,181 @@
...
+	DefaultVectorMemoryRequest = "128Mi"
+	DefaultVectorCpuLimit      = "2"
+	DefaultVectorMemoryLimit   = "2Gi"
+	DefaultScrapeInterval      = 30
+)
+
+// InjectCassandraVectorAgent adds the Vector agent container to the Cassandra pods.
+// If the Vector agent is already present, it is not added again.
+func InjectCassandraVectorAgent(telemetrySpec *telemetry.TelemetrySpec, dcConfig *cassandra.DatacenterConfig, k8cName string, logger logr.Logger) error {
+	if telemetrySpec != nil {
```
adutra, line 32: Tip: you don't need to test `telemetrySpec != nil` before calling `telemetrySpec.IsVectorEnabled()`.

pkg/telemetry/vector.go
```diff
@@ -0,0 +1,181 @@
...
+		if telemetrySpec.IsVectorEnabled() {
+			logger.Info("Injecting Vector agent into Cassandra pods")
+			vectorImage := defaultVectorImage
+			if telemetrySpec.Vector != nil && telemetrySpec.Vector.Image != "" {
```
adutra, line 36: `telemetrySpec.Vector != nil` is always true here.

emerkle826 (approved):
LGTM

apis/telemetry/v1alpha1/telemetry_types.go
adejanovski: Sounds good, I'll use the Duration type.

controllers/k8ssandra/vector.go
adejanovski: > I suggest that you take a look at the reconcileDefaultPerNodeConfiguration method since the situation is similar.

Ok, I'll pick a few things you have there.
The workflow is slightly different so I can't use the exact same implementation.

Good catch on the line 57 failure. I've added a requeue after the creation of the config map so that we can get the actual cm on the next reconcile and be able to move forward.

docs/vector/README.md
adejanovski: I'll add a sentence on this 👍

docs/vector/README.md
adejanovski: well, you could have multiple DCs in the same cluster but different namespaces.

pkg/telemetry/vector.go
adejanovski: oh, neat 👍

pkg/telemetry/vector.go
adejanovski: 👍

pkg/telemetry/vector.go
adejanovski: good point 👍

docs/vector/README.md
```diff
@@ -0,0 +1,165 @@
...
+          limits:
+            cpu: 1000m
+            memory: 1Gi
+...
+...
+```
+
+Telemetry settings can be configured at the cluster level and then overridden at the datacenter level.
+
+```yaml
```
adutra, line 35: I think this markup is a leftover.

adejanovski, line 35:
```suggestion
```

PR #798, opened 2022-12-16
Create a mutating webhook for the internal secrets provider

### LINKED ISSUE #605: K8SSAND-1623 ⁃ Create a mutating webhook for the internal secrets provider
Create a webhook for the internal secrets provider which leverages annotations to inject an init container in first position into pods, create an emptyDir memory volume and mount it to all (init-)containers in the pod.
The webhook shall pass the list of secrets to mount along with golang formatting templates to the secrets injection init container.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1623) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1623
┆priority: Medium

sseidman:
**What this PR does**:
Creates a webhook that, as a default, creates volumes and volume mounts on created/updated pods for secrets that are specified via a specific annotation. 

For example, a pod with the annotation `k8ssandra.io/inject-secret-cassandraSuperuser=[{"secretName": "cassandraSuperUser", "path": "/etc/credentials/cassandra"}]` will create a volume for the `cassandraSuperuser` secret and mount the secret to the path `/etc/credentials/cassandra` on each container for the pod. Pods will only be eligible for mutation by the webhook if they have the following match label: `k8ssandra.secrets/mutate: "inject"`.  This will mount secrets in the default Kubernetes style, which will create a single file for each key/value pair in the Secret (e.g. there will be `username` file and `password` file for credential sets).

The following manifest can be used to test the secrets-webhook after setting up a cluster with operators via kind:

```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
data:
  USER_NAME: YWRtaW4=
  PASSWORD: MWYyZDFlMmU2N2Rm

---
kind: K8ssandraCluster
metadata:
  name: demo
spec:
  secretsProvider: external
  cassandra:
    metadata:
      pods:
        labels:
          k8ssandra.secrets/mutate: "inject"
        annotations:
          k8ssandra.io/inject-secret: "[{\"secretName\": \"mysecret\", \"path\": \"/my/secret/path\"}]"
    serverVersion: "4.0.1"
    serverImage: k8ssandra/cass-management-api:4.0.1
    storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
    config:
      jvmOptions:
        heapSize: 512Mi
    datacenters:
      - metadata:
          name: dc1
        size: 3
  medusa:
    storageProperties:
      storageProvider: local
      bucketName: k8ssandra-medusa
      prefix: test
      podStorage:
        storageClassName: standard
        accessModes:
          - ReadWriteOnce
        size: 100Mi
  reaper:
    autoScheduling:
      enabled: false
 ```

Verify that when pods are started, they have the new Volumes and VolumeMounts and that the secrets are written to files in the `/my/secret/path` directory.

```
cassandra@demo-dc1-default-sts-2:/my/secret/path$ pwd
/my/secret/path
cassandra@demo-dc1-default-sts-2:/my/secret/path$ ls
PASSWORD  USER_NAME
```

Fixes #605 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/workflows/kind_multicluster_e2e_tests.yaml
+++ b/.github/workflows/kind_multicluster_e2e_tests.yaml
@@ -63,6 +63,23 @@ jobs:
         with:
           name: k8ssandra-operator
           path: /tmp/k8ssandra-operator.tar
+      - name: Build secrets webhook image
+        id: docker_build_secrets_webhook
+        uses: docker/build-push-action@v2
+        with:
+          file: Dockerfile
+          context: .
+          push: false
+          tags: k8ssandra/k8ssandra-secrets-webhook:latest,k8ssandra/k8ssandra-secrets-webhook:${{ steps.parse_image_tag.outputs.image_tag }}
+          platforms: linux/amd64
+          outputs: type=docker,dest=/tmp/k8ssandra-secrets-webhook.tar
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
+      - name: Upload k8ssandra-secrets-webhook image
+        uses: actions/upload-artifact@v2
+        with:
+          name: k8ssandra-secrets-webhook
+          path: /tmp/k8ssandra-secrets-webhook.tar
   kind_e2e_tests:
     runs-on: ubuntu-latest
     needs: build_image
@@ -129,9 +146,15 @@ jobs:
         with:
           name: k8ssandra-operator
           path: /tmp
+      - name: Download k8ssandra-secrets-webhook image
+        uses: actions/download-artifact@v2
+        with:
+          name: k8ssandra-secrets-webhook
+          path: /tmp
       - name: Load images
         run: |
           docker load --input /tmp/k8ssandra-operator.tar
+          docker load --input /tmp/k8ssandra-secrets-webhook.tar
       - name: Setup kind clusters
         run: make IMG="k8ssandra/k8ssandra-operator:${{ needs.build_image.outputs.image_tag }}" create-kind-multicluster kind-load-image-multi cert-manager-multi nginx-kind-multi
       - name: Run e2e test ( ${{ matrix.e2e_test }} )
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ IMG ?= $(IMAGE_TAG_BASE):latest
 # Create Kubernetes objects with embeddedObjectMeta
 CRD_OPTIONS ?= "crd"
 
+WEBHOOK_IMG_BASE ?= k8ssandra/k8ssandra-secrets-webhook
+WEBHOOK_IMG ?= $(WEBHOOK_IMG_BASE):latest
+
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.25.x
 
@@ -165,8 +168,13 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build:
+docker-build-operator:
 	docker buildx build --load -t ${IMG} .
+
+docker-build-secrets-webhook:
+	docker buildx build -t ${WEBHOOK_IMG} -f cmd/secrets-webhook/Dockerfile .
+
+docker-build: docker-build-operator docker-build-secrets-webhook
 
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
@@ -194,7 +202,7 @@ kind-multi-e2e-test: multi-create multi-prepare e2e-test
 
 single-create: cleanup create-kind-cluster cert-manager nginx-kind
 
-single-prepare: build manifests docker-build kind-load-image
+single-prepare: build manifests docker-build kind-load-image kind-load-image-secrets-webhook
 
 single-up: single-create single-prepare kustomize
 	kubectl config use-context kind-k8ssandra-0
@@ -215,7 +223,7 @@ endif
 
 multi-create: cleanup create-kind-multicluster cert-manager-multi nginx-kind-multi
 
-multi-prepare: build manifests docker-build kind-load-image-multi
+multi-prepare: build manifests docker-build docker-build-secrets-webhook kind-load-image-multi  
 
 multi-up: multi-create multi-prepare kustomize
 ## install the control plane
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

**Comment by Miles-Garnsey** [comment]
Thanks for this contribution @sseidman, we are having some internal discussions right now regarding this one, give us a couple of days to work up some feedback for you :)

**Comment by sseidman** [comment]
Hi @adutra, I think my thought process at the time was to try and make it configurable so that someone could use their own webhook if they wanted to (e..g https://developer.hashicorp.com/vault/docs/platform/k8s/injector). This isn't a requirement for us specifically, but I remember in the early design talks that was often referenced as an example.

I agree, making it separate like this would increase the overall maintenance complexity. I have a draft PR for building the webhook into the operator [here](https://github.com/k8ssandra/k8ssandra-operator/pull/857) that's almost done. I suppose there could be a property that toggles whether the webhook is registered with the operator, which could allow for the ability to introduce a separate webhook if needed.

**Comment by adejanovski** [comment]
Hi @sseidman, 

thanks a lot for reworking your PR. @Miles-Garnsey will be performing a review shortly.

File: .github/workflows/kind_multicluster_e2e_tests.yaml [hunk 1]
```diff
@@ -129,9 +146,15 @@ jobs:
         with:
           name: k8ssandra-operator
           path: /tmp
+      - name: Download k8ssandra-secrets-webhook image
```
Miles-Garnsey, line 149: My main note from my preliminary review is that there is an issue with the kuttl tests, I think that you'll need to add the same logic there as well as to the other test workflow files (especially `.github/workflows/kuttl_tests.yaml`, `.github/workflows/version_tests.yaml`).

The part that needs to be added is around line 100 and 94 respectively, and I've reproduced it for clarity below:
```
     - name: Download k8ssandra-secrets-webhook image
        uses: actions/download-artifact@v2
        with:
          name: k8ssandra-secrets-webhook
          path: /tmp
      - name: Load images
        run: |
          docker load --input /tmp/k8ssandra-operator.tar
          docker load --input /tmp/k8ssandra-secrets-webhook.tar
```

Hopefully that will fix the bulk of the test failures.

Makefile
```diff
@@ -54,6 +54,9 @@ IMG ?= $(IMAGE_TAG_BASE):latest
 # Create Kubernetes objects with embeddedObjectMeta
 CRD_OPTIONS ?= "crd"
 
+WEBHOOK_IMG_BASE ?= k8ssandra/k8ssandra-secrets-webhook
+WEBHOOK_IMG ?= $(WEBHOOK_IMG_BASE):latest
```
Miles-Garnsey, line 58: @adejanovski how will we handle versioning for this new image? Do we version it in lockstep with the existing k8ssandra-operator image? Do we want to publish a snapshot image on every commit?

Makefile
```diff
@@ -165,9 +168,14 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build:
+docker-build-operator:
 	docker buildx build --load -t ${IMG} .
 
+docker-build-secrets-webhook:
```
Miles-Garnsey, line 174: Would it possibly be worth importing the makefile targets from the webhook's makefile into this makefile? You might need to change their names to avoid collisions, but right now you'd need to `cd` into the webhook folder, build the binary, then cd out before building everything else.

cmd/secrets-webhook/mutation/inject_secrets.go
```diff
@@ -0,0 +1,116 @@
...
+		se.Logger.Debug("added volume and volumeMount to podSpec",
+			zap.String("secret", secretName),
+			zap.String("secret path", mountPath),
+		)
+	}
+
+	return mpod, nil
+}
+
+func injectVolume(pod *corev1.Pod, volume corev1.Volume) {
```
Miles-Garnsey, line 80: Would it be reasonable to use some of the functions [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L497) to perform these operations? I think this binary is being built entirely separately from the k8ssandra-operator binary, so I'd hope we won't encounter any circular dependencies.

cmd/secrets-webhook/mutation/inject_secrets.go
```diff
@@ -0,0 +1,116 @@
...
+func hasVolume(volumes []corev1.Volume, volume corev1.Volume) bool {
+	for _, v := range volumes {
+		if v.Name == volume.Name {
+			return true
+		}
+	}
+	return false
+}
+
+func injectVolumeMount(pod *corev1.Pod, volumeMount corev1.VolumeMount) {
```
Miles-Garnsey, line 95: Similar here, can some of the functions [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L497) be re-used?

Miles-Garnsey (changes requested):
Firstly @sseidman thanks for updating the webhook to use Kubebuilder, I think that'll help keep our codebase more consistent. 

I have a few comments on currently failing tests and the new build structures. I haven't yet manually tested but will try to do so today.

My main note is that I'm not seeing quite enough tests here just yet. I think we'd ideally like to see a few unit tests for each logic path for each function. 

An end to end test can also be run out of the parent folder if you want to make use of those frameworks. A kuttl test would be an easy way to confirm that the expected volumes are present - as you can just copy and paste from a live pod's yaml.

I'll provide more feedback as I do some more manual testing and read the code in more depth.

cmd/secrets-webhook/Makefile
```diff
@@ -0,0 +1,18 @@
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+.PHONY: test
+test: ## run unit tests
+	@echo "\n🛠️  Running unit tests..."
+	go test ./...
+
+.PHONY: build
+build: ## build binaries
```
Miles-Garnsey, line 10: Can I suggest running steps including `go fmt`, `test` and `go vet` as separate make targets but as dependancies to this target.

Makefile
```diff
@@ -165,9 +168,14 @@ build: generate fmt vet ## Build manager binary.
...
+	docker buildx build -t ${WEBHOOK_IMG} -f cmd/secrets-webhook/Dockerfile .
+
+docker-build: docker-build-operator docker-build-secrets-webhook
```
Miles-Garnsey, line 177: Running this target doesn't seem to leave me with a `k8ssandra/k8ssandra-secrets-webhook:latest` image in my local docker registry. I think you might need to add `--load` to the `docker-build-secrets-webhook` target to make this work.

sseidman:
Thanks for the review comments @Miles-Garnsey. Just want to clarify before I make changes, are we proceeding with this PR (separate webhook image, non-kubebuilder) or this https://github.com/k8ssandra/k8ssandra-operator/pull/857 (webhook built into operator, kubebuilder)

Miles-Garnsey:
> Thanks for the review comments @Miles-Garnsey. Just want to clarify before I make changes, are we proceeding with this PR (separate webhook image, non-kubebuilder) or this https://github.com/k8ssandra/k8ssandra-operator/pull/857 (webhook built into operator, kubebuilder)

Ah, indeed, I have gotten confused between the two PRs, oops!

I'll carry the comments from this PR over to there (since I think most still apply) and continue my review there. There will be some more feedback as I haven't gotten a manual test to work just yet.

PR #801, opened 2022-12-20
Enable smart token allocation by default for DSE

### LINKED ISSUE #796: New token allocator is disabled with DSE
The new token allocator uses the `allocate_tokens_for_local_replication_factor` setting to perform smart token allocations during bootstrap.

While this setting exists by default in Cassandra 4, it's not the case in DSE 6.8.
We need to set it explicitly to 3 (using the same default as Cassandra) unless the setting is overridden in the manifest.

Currently we end up with balance issues when using a small number of tokens.

adejanovski:
**What this PR does**:
DSE cassandra.yaml file doesn't set a default value for `allocate_tokens_for_local_replication_factor` and thus disables smart allocations by default.
This PR will enable it for DSE, using the same defaults as C* 4.x, while preserving any override from users.

Fixes #796 

**Checklist**
- [x] Changes manually tested


adutra (approved):
LGTM, e2e failures are unrelated afaict.

PR #812, opened 2023-01-17
Allow disabling MCAC

### LINKED ISSUE #806: Add support for the new metrics endpoint replacing MCAC
As we're wrapping up the work on replacing MCAC [in the management API](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/238), we'll need to make the changes in k8ssandra-operator to support it.

A new config map will need to be mounted in order to configure the metrics extraction using [this file format](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/471b7bc87b5b3ece9e5fe798b6cbcd48a281ad8e/management-api-agent-common/src/test/resources/collector.yaml).

The new metrics endpoint listens by default on `http://127.0.0.1:9000/metrics`. The port could be overridden to match the currently used port: `9103`.

Filters will need to be set so that we apply the same filters as [with MCAC](https://github.com/k8ssandra/k8ssandra-operator/blob/7f5687f9b78f6189e7094b1c72a2e1913ae43eb5/pkg/telemetry/cassandra_metrics_filters.go#L16-L29). Metrics names are fairly similar, so it shouldn't be much of a problem. The way filters apply is a little different though.

We'll have to remove the relabelling rules from the service monitors as the metrics will be correctly labeled in the agent.
_We may want to enrich the labels with the K8ssandraCluster name?_

Vector's agent config needs the cassandra_metrics source to point at the new URL: `http://localhost:9000/metrics` (or 9103 if we override the port).

Note that we'll need to keep supporting MCAC for older versions of the mgmt api. We'll have either have to detect if the new endpoint is being used (not sure how), or introduce a new telemetry setting that indicates the new endpoint is being used (which is not dynamic at all and won't provide the best experience).

### LINKED ISSUE #814: Add a CRD field to disable MCAC
Add a new field named McacTelemetrySpec.Enabled which will allow enabling/disabling MCAC. The default is true for now. Setting it to false translates into adding the MGMT_API_DISABLE_MCAC env variable to the cassandra container, with a value of true.

### LINKED ISSUE #820: Update yq `eval-all` functionality
**What happened?**
`yq` test is failing when using the `eval-all` option. Example:
```sh
/bin/yq ea .spec.cassandra.datacenters.[] as $item ireduce (0; . +1) ../testdata/fixtures/multi-dc/k8ssandra.yaml ../testdata/fixtures/single-dc/k8ssandra.yaml

--- FAIL: TestEval (0.10s)
    --- FAIL: TestEval/count_dcs_many_files_eval_all (0.01s)
        yq_test.go:99: 
            	Error Trace:	/home/runner/work/k8ssandra-operator/k8ssandra-operator/test/yq/yq_test.go:99
            	Error:      	Not equal: 
            	            	expected: "3"
            	            	actual  : "3\n3"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1,2 @@
            	            	 3
            	            	+3
            	Test:       	TestEval/count_dcs_many_files_eval_all
```
The test expects all input files to be processed once. So in the example above, it's trying to count the number of datacenters defined in 2 different files. One file has 2 DCs defined, the other has only 1. With the `eval-all` option, the output is expected to be 3. With newer versions of `yq` the output is repeated for the number of input files, so we now get:
```
3
3
```

This test was passing as recently as yesterday (from the time of writing this ticket):
https://github.com/k8ssandra/k8ssandra-operator/actions/runs/3956583819/jobs/6776181660

There seems to be a bug reporting this issue:
https://github.com/mikefarah/yq/issues/1206

See the test failure above.

Miles-Garnsey:
**What this PR does**:

Adds a field to disable MCAC

Fixes #814
Fixes #820


test/kuttl/test-servicemonitors/03-assert.yaml
```diff
@@ -3,6 +3,22 @@ kind: StatefulSet
...
+            - name: MGMT_API_DISABLE_MCAC
+              value: "true"
+            - name: LOCAL_JMX
+            - name: METRIC_FILTERS
+            - name: DS_LICENSE
+            - name: DSE_AUTO_CONF_OFF
+            - name: USE_MGMT_API
+            - name: MGMT_API_EXPLICIT_START
+            - name: DSE_MGMT_EXPLICIT_START
+        - name: server-system-loggerßß
```
adejanovski, line 21:
```suggestion
        - name: server-system-logger
```

Not sure if that was intentional :)

Miles-Garnsey: It wasn't. Thanks...

apis/telemetry/v1alpha1/telemetry_types.go
```diff
@@ -68,4 +68,7 @@ type McacTelemetrySpec struct {
 	// Setting it to an empty list will result in all metrics being extracted.
 	// +optional
 	MetricFilters *[]string `json:"metricFilters,omitempty"`
+
+	// enabled sets whether MCAC (legacy metrics endpoint) is enabled.
```
adutra, line 72: We could add a comment stating that MCAC is considered enabled by default, since it might be surprising given that MCAC is legacy.

controllers/k8ssandra/cassandra_telemetry_reconciler.go
```diff
@@ -72,6 +72,7 @@ func (r *K8ssandraClusterReconciler) reconcileCassandraDCTelemetry(
 			return result.Error(err)
 		}
 	}
+
```
adutra, line 75: Unnecessary change.

pkg/cassandra/datacenter.go
```diff
@@ -209,21 +212,40 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
+}
+
+func setMcacDisabled(dc *cassdcapi.CassandraDatacenter, template *DatacenterConfig) {
+	if !isMcacEnabled(template) {
+		UpdateCassandraContainer(dc.Spec.PodTemplateSpec, func(c *corev1.Container) {
+			c.Env = append(
+				c.Env,
+				corev1.EnvVar{
+					Name:  mcacDisabledEnvVar,
+					Value: fmt.Sprintf("%v", !isMcacEnabled(template)),
```
adutra, line 242:
```suggestion
					Value: "true",
```

pkg/cassandra/datacenter.go
```diff
@@ -209,21 +212,40 @@ func NewDatacenter(klusterKey types.NamespacedName, template *DatacenterConfig)
...
 
+func isMcacEnabled(template *DatacenterConfig) bool {
+	if template.McacEnabled != nil {
+		return *template.McacEnabled
+	}
+	return true
+}
+
+func setMcacDisabled(dc *cassdcapi.CassandraDatacenter, template *DatacenterConfig) {
+	if !isMcacEnabled(template) {
```
adutra, line 236: Nit: I'd rather pull this if check out of this function since it is surprising to test whether X is enabled or not inside a function whose name is `setXDisabled`.

pkg/cassandra/datacenter.go
```diff
@@ -115,6 +116,7 @@ type DatacenterConfig struct {
 	PerNodeConfigMapRef       corev1.LocalObjectReference
 	PerNodeInitContainerImage string
 	ExternalSecrets           bool
+	McacEnabled               *bool
```
adutra, line 119: Is this considered enabled if nil? I think so. We might want to document this.

pkg/cassandra/datacenter.go
```diff
@@ -243,7 +265,7 @@ func UpdateVectorContainer(p *corev1.PodTemplateSpec, f func(c *corev1.Container
 // f. Only the Name field is initialized.
 func UpdateContainer(p *corev1.PodTemplateSpec, name string, f func(c *corev1.Container)) {
 	idx, found := FindContainer(p, name)
-	container := &corev1.Container{}
+	var container *corev1.Container
```
adutra, line 268: Nice 👍

pkg/cassandra/datacenter.go
```diff
@@ -344,6 +366,10 @@ func Coalesce(clusterName string, clusterTemplate *api.CassandraClusterTemplate,
 	// we need to declare at least one container, otherwise the PodTemplateSpec struct will be invalid
 	UpdateCassandraContainer(&dcConfig.PodTemplateSpec, func(c *corev1.Container) {})
 
+	if mergedOptions.Telemetry != nil && mergedOptions.Telemetry.Mcac != nil && mergedOptions.Telemetry.Mcac.Enabled != nil {
```
adutra, line 369: I think we can simplify this: `dcConfig.McacEnabled` could be turned into a `bool` instead of `*bool`, then we could just:

```go
dcConfig.McacEnabled = mergedOptions.Telemetry.IsMcacEnabled()
```

test/yq/yq_test.go
```diff
@@ -1,11 +1,12 @@
 package yq
 
 import (
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"os"
 	"strings"
 	"testing"
+
```
adutra, line 7: Unnecessary change.

apis/telemetry/v1alpha1/telemetry_types.go
adejanovski: ok, fixed 👍

adejanovski (commented):
@adutra, I think I've addressed all your comments.
Let me know what you think.

adutra (approved):

.github/workflows/test_and_build_image.yaml
```diff
@@ -37,6 +37,12 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
+      - name: Install yq
+        run: |
+          sudo rm /bin/yq
+          wget https://github.com/mikefarah/yq/releases/download/v4.28.1/yq_linux_amd64.tar.gz
```
burmanm, line 43: Why are we installing 4.28.1 ?

emerkle826: > Why are we installing 4.28.1 ?

To fix the `yq` issue [here](https://github.com/k8ssandra/k8ssandra-operator/issues/820).

emerkle826: I just tried with version 4.30.6, and it works. So maybe we can update to that version, but 4.30.8 seems to have the issue in #820

emerkle826: So it seems the issue exists with version 4.30.7 and 4.30.8, but not 4.30.6

burmanm: It might make sense to issue a bug report, since that 820 was referring to a bug in 4.9.8, not 4.30. And I don't see any change like that between 4.30.6 and 4.30.8 release notes.

burmanm: Or, alternatively.. it's this we hit actually: https://github.com/mikefarah/yq/issues/1515

In that case, we should probably fix our implementation and not change yq version ;)

PR #813, opened 2023-01-17
New telemetry agent config

### LINKED ISSUE #806: Add support for the new metrics endpoint replacing MCAC

### LINKED ISSUE #815: Add a Cassandra struct in the Telemetry spec for the new metrics endpoint
A new TelemetrySpec.Cassandra struct must be added to configure the new metrics endpoint.
The values there will be used to generate a config map using [this file format](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/471b7bc87b5b3ece9e5fe798b6cbcd48a281ad8e/management-api-agent-common/src/test/resources/collector.yaml).
Filters can reuse Prometheus's [RelabelConfig](https://github.com/prometheus-operator/prometheus-operator/blob/e4e27052f57040f073c6c1e4aedaecaaec77d170/pkg/apis/monitoring/v1/types.go#L1150) struct, reduced to just the supported fields for our filters.

Miles-Garnsey:
**What this PR does**:

Adds configuration functionality for the new telemetry agent.

Fixes #815

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
I had a few questions on this one:

1. It looks like the new metrics [config](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/471b7bc87b5b3ece9e5fe798b6cbcd48a281ad8e/management-api-agent-common/src/test/resources/collector.yaml) doesn’t conform to the Prometheus RelabelConfig API as expected. It uses source_labels instead of sourceLabels. @burmanm has confirmed that we need to update this so it is camelcased.
2. I wrote some code to unwind the PodTemplateSpec configuration when Spec.Cassandra.Telemetry.Cassandra is nil, but I think that (as the new agent is always enabled) this code has no purpose. Is there any case in which we don't run the  new agent or might want to remove it's ConfigMaps?
3. The types in this part of the codebase have become problematic, and so is the original ticket. At some point, folks have added MCAC and Vector fields into the TelemetrySpec. This does not make sense because Mcac only applies when the TelemetrySpec is under the Spec.Cassandra field (but TelemetrySpec is used for Reaper and Stargate too). I'm not sure about Vector, but it probably needs more thought too.

I propose that:
1. I fix the field name on source_labels (changing to sourceLabels).
2. I remove the `RemoveStsVolumes` function, and always add the configmap in reconciliation.
3. We split the types out so that Mcac and Cassandra fields (and Vector?) become embedded structs under a new types CassandraTelemetrySpec.

adejanovski:
> I fix the field name on source_labels (changing to sourceLabels).

Sounds good

> I remove the RemoveStsVolumes function, and always add the configmap in reconciliation.

Sure, I don't see a reason to remove the volume.

> We split the types out so that Mcac and Cassandra fields (and Vector?) become embedded structs under a new types CassandraTelemetrySpec.

Vector applies to Reaper and Stargate as well.
I'm not keen on breaking stuff, knowing that we're probably not going to write a converting webhook.
So the Cassandra struct can move if you feel like it makes more sense, but the MCAC one should stick around, knowing that we'll fully remove it in the near future (We're not gonna keep MCAC for very long, just the time to transition).

Miles-Garnsey:
> I'm not keen on breaking stuff, knowing that we're probably not going to write a converting webhook.

This doesn't break anything, the yaml stays the same. The only breakage will occur if users have been adding the Mcac field to their telemetry specs for Reaper/Stargate, in which case that configuration is actually unsupported/broken anyway.

adejanovski:
> > I'm not keen on breaking stuff, knowing that we're probably not going to write a converting webhook.
> 
> This doesn't break anything, the yaml stays the same. The only breakage will occur if users have been adding the Mcac field to their telemetry specs for Reaper/Stargate, in which case that configuration is actually unsupported/broken anyway.

ok then, if there's no breakage then you can go nuts.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if err := c.RemoteClient.Create(c.Ctx, &cm); err != nil {
+		return err
+	}
+	c.AddStsVolumes(dc)
+
+	return nil
+}
+
+func (c Configurator) AddStsVolumes(dc *cassdcapi.CassandraDatacenter) error {
```
adejanovski, line 56: Could you reuse `AddOrUpdateVolume()` which is in `pkg/cassandra/datacenter.go` ?

Miles-Garnsey:
I don't quite know what's happening here, but for some reason using `sourceLabels` doesn't work. The Prometheus APIs appear to be serialising that field into `sourcelabels` instead. That's despite the fact that the json tag is definitely `sourceLabels` for that field (see [here](https://github.com/prometheus-operator/prometheus-operator/blob/e4e27052f57040f073c6c1e4aedaecaaec77d170/pkg/apis/monitoring/v1/types.go#L1154)).

@burmanm I might need to update your metrics agent again I'm sorry :(

Miles-Garnsey:
Can I draw any reviewer's attention to this block, which is the default location for the configuration file, and the default configuration itself. 

Is this the way we want it set up?

```
var (
	agentConfigLocation = "/config/metric-collector.yaml"
	defaultAgentConfig  = telemetryapi.CassandraAgentSpec{
		Endpoint: telemetryapi.Endpoint{
			Port:    "9001",
			Address: "127.0.0.1",
		},
		Filters: []promapi.RelabelConfig{
			{
				SourceLabels: []string{"__tag1__", "__tag2__"},
				Separator:    ";",
				Regex:        "(.*);(b.*)",
				Action:       "drop",
			},
			{
				SourceLabels: []string{"__tag1__", "__tag2__"},
				Separator:    ",",
				Regex:        "^(a|b|c),.*",
				Action:       "drop",
			},
		},
	}
)
```

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: I had a close look at this and wasn't quite sure that it was appropriate. I'm open minded here, but it seems that the function you're referring to works on the `datacenterConfig` type, whereas we currently work directly on the cassDC and hook into the reconciler at that level. 

If you want me to investigate a change of this nature in more detail, I'm happy to do so, but it is a fairly deep re-write.

Miles-Garnsey: The advantage is that using that function does overwrite any existing `Volume` which might already exist. But also, we don't usually provide guarantees in our controllers that a manually modified object will be self-healed back to the desired state. This is in line with in-built k8s controllers like the STS controller.

Miles-Garnsey:
This is ready to go as far as I can see. The only areas where we may want changes are:

* Using `cassandra.AddOrUpdateVolume` in the reconciliation logic. This does change the locations where we'd hook our reconciliation logic into the reconcilers quite substantially though and involves a bit of rework. If there are benefits or this is the way it is done for the rest of the reconciliation I'm happy to make the change.
* I've modified an interface [here](https://github.com/k8ssandra/k8ssandra-operator/pull/813/files#diff-649b8176156d3db5da1a98c702c5e16b905937907e9b461241bdda24afe49185R15) to enable smoother integration of reconciliation logic further away from the reconcilers themselves (passing Result types around in `pkg` modules seems to lead to better modularity). But the alternative would have been to use a case switch statement on the returned result type, which would resulted in less overall change to the existing Result interfaces. I'm  open to doing it this way instead if preferred.

Miles-Garnsey:
`ClusterScoped/MultiDcMultiCluster` is failing consistently, I can't test locally because `make multi-create` errors out. 

@adejanovski do you want me to just risk merging this and see what happens? Or should I try to figure out some way to debug manually?

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -0,0 +1,148 @@
...
+		yamlData, err = yaml.Marshal(&defaultAgentConfig)
+		if err != nil {
+			return &corev1.ConfigMap{}, err
+		}
+	}
+
+	cm := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: c.Kluster.Namespace,
+			Name:      c.Kluster.Name + "-metrics-agent-config",
```
adejanovski, line 59: You'll have one configmap per DC, so you can't use the K8ssandraCluster namespace.
Also the name should be `<KlusterName>-<DC name>-....`.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (c Configurator) ReconcileTelemetryAgentConfig(dc *cassdcapi.CassandraDatacenter) result.ReconcileResult {
+	//Reconcile the agent's ConfigMap
+	desiredCm, err := c.GetTelemetryAgentConfigMap()
+	if err != nil {
+		return result.Error(err)
+	}
+	cmObjectKey := types.NamespacedName{Name: c.Kluster.Name + "-metrics-agent-config",
+		Namespace: c.Kluster.Namespace}
```
adejanovski, line 73: Same here, the configmap will be created for each datacenter, instead of one per cluster.

The configmap name should be extracted to a function so that we don't repeat it in multiple places.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: Working on a fix, let's see if CI passes.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -0,0 +1,155 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	agentConfigLocation = "/opt/management-api/metric-collector.yaml"
```
adejanovski, line 24:
```suggestion
	agentConfigLocation = "/opt/management-api/configs/metric-collector.yaml"
```

As you can see [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/commit/f631db71b4f4bfefb3791e42b080e6c2df3f7fc5#diff-813a5eb5912d07588a0a72cf84860f552a482d75d77b91546d4125e53df29258R11) the right location is under `$MAAC_PATH/configs/...`.
I've successfully tested a custom filter after switching to that path. Otherwise my config was ignored.

FTR, another way of making it work with the current path is to set the [CONFIG_PATH_PROPERTY](https://github.com/k8ssandra/management-api-for-apache-cassandra/commit/f631db71b4f4bfefb3791e42b080e6c2df3f7fc5#diff-813a5eb5912d07588a0a72cf84860f552a482d75d77b91546d4125e53df29258R15) accordingly, but I think the right default path is all we need.

adejanovski (approved):
@Miles-Garnsey, please merge the PR after correcting the default path for the metrics config file.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: Good catch, thanks. I've pushed another commit. I wasn't sure if it was right (hence why I [asked](https://github.com/k8ssandra/k8ssandra-operator/pull/813#issuecomment-1406123112)) but I obviously misread [this](https://github.com/k8ssandra/k8ssandra-operator/issues/815#issuecomment-1396508708) answer quite badly - sorry!

Miles-Garnsey:
For some reason I cannot fathom, the tests have started failing since I changed the mount path. I'm testing manually using the file `test/testdata/fixtures/single-dc-reaper` which is deployed as part of `CreateSingleReaper`.

I'm seeing a strange effect where the cluster deploys fine if we set the server image to the snapshot containing the new management API changes (`serverImage: k8ssandra/cass-management-api:4.0-ea6f0a0`) but I did see issues with Cassandra becoming ready when the serverImage wasn't set. I'm running that latter test again now to validate.
.

Miles-Garnsey:
Since I made a single line change to move the configmap's mount point under `/opt/management-api/configs`, most of the e2e tests are now failing.

I've attempted to replicate the failures locally, and can report:
* I found that deploying `test/testdata/fixtures/single-dc-reaper` (as used in `CreateSingleReaper`) failed. It started succeeding when I moved to a 4.0.4 serverVersion with explicit image set to `serverImage: k8ssandra/cass-management-api:4.0-ea6f0a0`
* Working through to isolate the problem, I tracked down the below error message. I think there is some problem with the metrics agent on Cassandra 3.x.

```
ERROR [main] 2023-02-02 02:22:58,711 CassandraDaemon.java:803 - Exception encountered during startup
java.lang.NoSuchMethodError: com.fasterxml.jackson.dataformat.yaml.YAMLFactory._createContentReference(Ljava/lang/Object;)Lcom/fasterxml/jackson/core/io/ContentReference;
    at com.fasterxml.jackson.dataformat.yaml.YAMLFactory.createParser(YAMLFactory.java:391) ~[datastax-mgmtapi-agent-3.x-0.1.0-SNAPSHOT.jar:na]
    at com.fasterxml.jackson.dataformat.yaml.YAMLFactory.createParser(YAMLFactory.java:15) ~[datastax-mgmtapi-agent-3.x-0.1.0-SNAPSHOT.jar:na]
    at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2903) ~[jackson-databind-2.9.10.8.jar:2.9.10.8]
    at io.k8ssandra.metrics.config.ConfigReader.readConfig(ConfigReader.java:27) ~[datastax-mgmtapi-agent-3.x-0.1.0-SNAPSHOT.jar:na]
    at io.k8ssandra.metrics.interceptors.MetricsInterceptor.intercept(MetricsInterceptor.java:42) ~[datastax-mgmtapi-agent-3.x-0.1.0-SNAPSHOT.jar:na]
    at org.apache.cassandra.service.CassandraDaemon.start(CassandraDaemon.java) [apache-cassandra-3.11.11.jar:3.11.11]
    at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:648) [apache-cassandra-3.11.11.jar:3.11.11]
    at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:786) [apache-cassandra-3.11.11.jar:3.11.11]

```

Miles-Garnsey:
(From offline) We hit an issue where older versions of Cassandra caused test failures because jackson XML's methods changed between versions, and different Cassandra 3.11.x versions used different Jackson versions. 

To move things forward, I have updated all tests to use Cassandra 3.11.14, which should be compatible with the new metrics agent. 

I will leave it to the rest of the team (ping @burmanm and @adejanovski) to decide whether(a) we want to  continue supporting older Cassandra versions with the new metrics agent, or whether (b) we will make the call that older patch versions of Cassandra cannot benefit from the new agent. In general, the changes in this PR will be compatible with either whatever scenario (since we are just adding a file to a directory in `/opt`) but scenario (b) requires that the metrics agent doesn't try to start on older Cassandra / management-api images.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (c Configurator) ReconcileTelemetryAgentConfig(dc *cassdcapi.CassandraDatacenter) result.ReconcileResult {
+	//Reconcile the agent's ConfigMap
+	desiredCm, err := c.GetTelemetryAgentConfigMap()
+	if err != nil {
+		return result.Error(err)
+	}
+	cmObjectKey := types.NamespacedName{Name: c.Kluster.Name + "-metrics-agent-config",
+		Namespace: c.Kluster.Namespace}
+	labels.SetManagedBy(desiredCm, cmObjectKey)
```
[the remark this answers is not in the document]
Miles-Garnsey: @adejanovski yes the object is definitely created by the operator. Once the tests are passing I'll try removing the "Managed by" labels, and see if the tests keep passing.

PR #818, opened 2023-01-18
Allow configuring Vector agent sidecar in the CRD

### LINKED ISSUE #817: Allow configuring the Vector agent sidecar in the CRD
The initial implementation of the Vector Agent sidecar relies on an externally provided configmap for Vector's configuration.
This isn't always the most convenient way of configuring it and it would be nice to pass that configuration for sources, transforms and sinks directly in the CRD.

adejanovski:
**What this PR does**:
This PR adds fields in the telemetry spec for Vector which allow to pass components that need to be configured in Vector's toml.
This can be more convenient than using a configmap, which is still possible though.

Fixes #817 

**Checklist**
- [x] Changes manually tested


pkg/telemetry/vector.go
```diff
@@ -114,11 +115,14 @@ target = "stdout"
 			return "", err
 		} else {
 			if toml, found := vectorConfigMap.Data["vector.toml"]; found {
-				sinks = toml
+				vectorConfigToml = toml
 			} else {
 				return "", fmt.Errorf("vector.toml not found in config map %s", telemetrySpec.Vector.Config.Name)
 			}
 		}
+	} else if telemetrySpec.Vector.Components != nil {
```
emerkle826, line 123: We discussed this briefly in private, but I think it would be less complex to just allow for the new way of configuring Vector via the CRD/Components. Unless we need to also support the original way of configuring it through a ConfigMap, I'd prefer the Component configuration you have added here.

emerkle826 (approved):
Unless we need it, I vote to replace the ConfigMap configuration with the Component/CRD configuration you've added here.

PR #833, opened 2023-01-28
Don't reference DCs in system ks replication settings before they exist (fixes #829)

### LINKED ISSUE #829: Cannot create MultiDC cluster on cassandra 4.1+
**What happened?**
Due to releasing https://issues.apache.org/jira/browse/CASSANDRA-12988 creating multidc clusters stucked on first cassandraDatacanter. 
1. K8ssandra-operator creates first cassandraDatacanter with `k8ssandra.io/initial-system-replication: '{"xxxx-dca":3,"xxxx-dcb":3}'` and wait cassandraOperatorProgress = "ready"
2. Cass-operator receives 500 in loop from mgmt-api POST /api/v0/ops/auth/role
```
2023-01-26T10:15:40.257Z    INFO    calling Management API create role - POST /api/v0/ops/auth/role    {"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "cassandraDatacenter": {"name":"xxxx-dca","namespace":"default"}, "namespace": "default", "name": "xxxx-dca", "reconcileID": "c01b266e-4461-4122-98f9-a40d28d64bb1", "pod": "xxxx-xxxx-dca-rack1-sts-0"}

2023-01-26T10:15:40.367Z    ERROR    incorrect status code when calling Node Management Endpoint    {"controller": "cassandradatacenter_controller", "controllerGroup": "cassandra.datastax.com", "controllerKind": "CassandraDatacenter", "cassandraDatacenter": {"name":"xxxx-dca","namespace":"default"}, "namespace": "default", "name": "xxxx-dca", "reconcileID": "c01b266e-4461-4122-98f9-a40d28d64bb1", "statusCode": 500, "pod": "10.111.104.61", "error": "incorrect status code of 500 when calling endpoint"}
```
3. Cassandra 4.1 Cannot achieve EACH_QUORUM consistency because k8ssandra-operator didnt create second DC
```
INFO  [epollEventLoopGroup-5-3] 2023-01-26 10:14:59,138 RpcExecutionException.java:39 - Failed to execute method NodeOps.createRole
org.apache.cassandra.exceptions.UnavailableException: Cannot achieve consistency level EACH_QUORUM in DC xxxx-dcb
```

* K8ssandra Operator version: v1.4.1

It can be fixed by changing `auth_write_consistency_level`

olim7t:
**What this PR does**:
Only reference the first DC in the `k8ssandra.io/initial-system-replication` annotation. This avoids referencing non-existing DCs ahead of time (which doesn't work with Cassandra 4.1). We alter system keyspaces after each DC gets added.

Fixes #829

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/schemas.go
+++ b/controllers/k8ssandra/schemas.go
@@ -102,7 +102,10 @@ func (r *K8ssandraClusterReconciler) checkInitialSystemReplication(
 			}
 		}
 	} else {
-		replication = cassandra.ComputeReplicationFromDatacenters(3, kc.Spec.ExternalDatacenters, kc.Spec.Cassandra.Datacenters...)
+		// Cassandra 4.1 writes the superuser role with CL=EACH_QUORUM, so we can't reference DCs before they are created.
+		// Only reference the first DC in the replication; for subsequent DCs, the replication will be altered.
+		firstDc := kc.Spec.Cassandra.Datacenters[0]
+		replication = cassandra.ComputeReplicationFromDatacenters(3, kc.Spec.ExternalDatacenters, firstDc)
 	}
 
 	bytes, err := json.Marshal(replication)
@@ -140,8 +143,7 @@ func (r *K8ssandraClusterReconciler) updateReplicationOfSystemKeyspaces(
 		return recResult
 	}
 
-	datacenters := cassandra.GetDatacentersForSystemReplication(kc)
-	replication := cassandra.ComputeReplicationFromDatacenters(3, kc.Spec.ExternalDatacenters, datacenters...)
+	replication := cassandra.ComputeReplicationFromDatacenters(3, kc.Spec.ExternalDatacenters, kc.GetInitializedDatacenters()...)
 
 	logger.Info("Preparing to update replication for system keyspaces", "replication", replication)
 
--- a/CHANGELOG/CHANGELOG-1.5.md
+++ b/CHANGELOG/CHANGELOG-1.5.md
@@ -15,6 +15,8 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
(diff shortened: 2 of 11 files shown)


controllers/k8ssandra/schemas.go
```diff
@@ -86,6 +86,7 @@ func (r *K8ssandraClusterReconciler) checkInitialSystemReplication(
 
 	replication := make(map[string]int)
 	if val := annotations.GetAnnotation(kc, api.InitialSystemReplicationAnnotation); val != "" {
+		// TODO if a value exists, do we need to filter it to only keep the first DC?
```
olim7t, line 89: I'm not sure about this, but I would guess that if we already have the annotation, then there is at least one DC so we  won't need the value anymore.

adejanovski: I guess we'd need to test this. The annotation is only important when the cluster is bootstrapped indeed, and I think the operator creates that annotation, right?
Now, can a user override it and put something that would break it?
If so, you'll need a check and fail the reconcile.

dnugmanov:
Hi, @olim7t @adejanovski, could you pls clarify, is it will be released as 1.5.1?

controllers/k8ssandra/schemas.go
olim7t: Yes, the operator creates the annotation. Not only is it only used for bootstrapping, but with my changes it's now only applied to the first DC. The code already handles bad input.
So all things considered I think we're good, I'll remove the TODO.

olim7t:
> Hi, @olim7t @adejanovski, could you pls clarify, is it will be released as 1.5.1?

Hi @dnugmanov,
Yes, this will be included in 1.5.1.

adejanovski (approved):
Great stuff 👍 Successfully reviewed and tested :shipit:

PR #834, opened 2023-01-30
Default metrics filters for new metrics agent

### LINKED ISSUE #816: Port MCAC default filters to the new metrics endpoint
Filters will need to be set by default in the new metrics endpoint config, so that we apply the same ones as [with MCAC](https://github.com/k8ssandra/k8ssandra-operator/blob/7f5687f9b78f6189e7094b1c72a2e1913ae43eb5/pkg/telemetry/cassandra_metrics_filters.go#L16-L29).
Metrics names are fairly similar, so it shouldn't be much of a problem.

Filters priority is a little different though, so there could be some converting needed.

Miles-Garnsey:
**What this PR does**:

This PR hopefully replicates the previous metrics filters that applied to MCAC over to the new metrics agent's config file which we've built and mounted via [this](https://github.com/k8ssandra/k8ssandra-operator/pull/813) PR.

I haven't been able to track down docs on how the metrics relabelling/filtering works for the new agent, so I'm flying slightly blind here (if anyone knows where docs are please feel free to chime in). 

Having said that, if it works similarly to Prometheus then I think this will be pretty close to what we need.

Fixes #816 

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
I'm a bit concerned that we introduce a breaking change here. Apparently in the course of delivering the metrics agent work, the config field `filters` was changed to `relabels`. 

As a result, I have updated the CR's API so that filters is now `relabels` but I think that API was in the prior release. We should check this and discuss prior to approving this PR.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
 	agentConfigLocation = "/opt/management-api/configs/metric-collector.yaml"
 	defaultAgentConfig  = telemetryapi.CassandraAgentSpec{
-		Endpoint: telemetryapi.Endpoint{
+		Endpoint: &telemetryapi.Endpoint{
 			Port:    "9000",
 			Address: "127.0.0.1",
 		},
+		Relabels: []promapi.RelabelConfig{
+			{
+				SourceLabels: []string{"__origname__"},
```
burmanm, line 34: Use ``__table__`` here instead and check for ``.+``. That way even with changes to original feed it would stay relevant.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.Table.*",
+				TargetLabel:  "should_drop",
+				Replacement:  "true",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table.*",
```
burmanm, line 41: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				TargetLabel:  "should_drop",
+				Replacement:  "true",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.live_ss_table_count",
```
burmanm, line 47: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.Table\\.LiveSSTableCount",
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.live_disk_space_used",
```
burmanm, line 59: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.Table\\.Compaction",
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.read",
```
burmanm, line 83: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.write",
```
burmanm, line 89: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.range",
```
burmanm, line 95: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.coordinator",
```
burmanm, line 101: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -22,10 +25,95 @@ import (
...
+				TargetLabel:  "should_drop",
+				Replacement:  "false",
+			},
+			{
+				SourceLabels: []string{"__origname__"},
+				Regex:        "org\\.apache\\.cassandra\\.metrics\\.table\\.dropped_mutations",
```
burmanm, line 107: This metric name does not exists in Cassandra.

pkg/telemetry/cassandra_agent/cassandra_agent_config_test.go
```diff
@@ -78,6 +166,47 @@ func Test_GetTelemetryAgentConfigMap(t *testing.T) {
...
+			Separator:    ";",
+			Regex:        "(.*);(b.*)",
+			Action:       "drop",
+		},
+	}
+	Cfg.TelemetrySpec.Cassandra.Endpoint = nil
+	cm, err := Cfg.GetTelemetryAgentConfigMap()
+	println(cm.Data)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedCm.Data["metric-collector.yaml"], cm.Data["metric-collector.yaml"])
```
burmanm, line 205: The correct filename is ``/configs/metrics-collector.yaml``, notice the *s*

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -43,7 +131,8 @@ func (c Configurator) GetTelemetryAgentConfigMap() (*corev1.ConfigMap, error) {
 	var yamlData []byte
 	var err error
 	if c.TelemetrySpec.Cassandra != nil {
-		yamlData, err = yaml.Marshal(&c.TelemetrySpec.Cassandra)
+		mergedSpec := goalesce.MustDeepMerge(&defaultAgentConfig, c.TelemetrySpec.Cassandra)
```
burmanm, line 134: @adejanovski and I have a slight problem with this. If the fields are merged, what's the ordering here? Can user override everything that's set in the defaults? How are they merged, before, after, what about duplicates?

adejanovski: I confirm that my expectation here is that any custom filter from users would override all the default filters. They shouldn't get merged with it.

Miles-Garnsey: And that's exactly how this works. MustDeepMerge uses an atomic merge for slices by default.

pkg/telemetry/cassandra_agent/cassandra_agent_config_test.go
Miles-Garnsey: That isn't what is specified [here](https://github.com/datastax/metric-collector-for-apache-cassandra/blob/master/config/metric-collector.yaml)?

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: I am building to the original [MCAC filters](https://github.com/k8ssandra/k8ssandra-operator/blob/7f5687f9b78f6189e7094b1c72a2e1913ae43eb5/pkg/telemetry/cassandra_metrics_filters.go#L16-L29) where that metric name is filtered. Perhaps this metrics differs in name between Cassandra versions.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -20,12 +23,91 @@ import (
...
 var (
-	agentConfigLocation = "/opt/management-api/configs/metric-collector.yaml"
+	agentConfigLocation = "/opt/management-api/configs/metrics-collector.yaml"
 	defaultAgentConfig  = telemetryapi.CassandraAgentSpec{
-		Endpoint: telemetryapi.Endpoint{
+		Endpoint: &telemetryapi.Endpoint{
 			Port:    "9000",
 			Address: "127.0.0.1",
 		},
+		Relabels: []promapi.RelabelConfig{
```
Miles-Garnsey, line 32: @adejanovski can you please take a look at this to confirm that I've captured the intention behind the [original filters](https://github.com/k8ssandra/k8ssandra-operator/blob/7f5687f9b78f6189e7094b1c72a2e1913ae43eb5/pkg/telemetry/cassandra_metrics_filters.go#L16-L29)?

Some metric names present quite differently (e.g. org.apache.cassandra.metrics.table.read doesn't appear directly anymore, now it seems to just be read_latency, similar with org.apache.cassandra.metrics.Table.LiveSSTableCount which now appears to be ... live_ss_table_count).

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: It appears that I actually needed to use `table`, not `__table__`, but once I made that change it appears that this works.

Miles-Garnsey:
I've done some manual testing and this is ready for merge, subject to my interpretation of the original metrics filters (which I've commented on above) being right.

pkg/telemetry/cassandra_agent/cassandra_agent_config_test.go
burmanm: That's MCAC?

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
burmanm: No, MCAC just modified the names to be non-original. The real documentation (as mentioned here: https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-common/src/main/resources/default-metric-settings.yaml#L2) is https://cassandra.apache.org/doc/latest/cassandra/operating/metrics.html

pkg/telemetry/cassandra_agent/cassandra_agent_config_test.go
Miles-Garnsey: Oops... I'm glad I mentioned this since I need to change some documentation back as I think it refers to MCAC, not the new collector. 

Thanks for clearing this up for me anyway, didn't realise this had also changed from the original PR.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
adejanovski: Awesome! I wasn't sure about this because Goalesce is very powerful and has a lot of different merge techniques.

burmanm: Well, in this case I'd say the "Merge" is a bit of a confusing (sorry, insanely confusing) name for a method (and perhaps this is why this would require a comment to say we don't ever want to merge these despite calling a function to merge). It's not merging anything. It either returns a copy of defaultAgentConfig (if c.TelemetrySpec.Cassandra is nil) or a copy of c.TelemetrySpec.Cassandra (if it isn't nil), it never merges values from those two. 

Now finding out the intended behavior requires to go through 4-5 jumps in files while reading which method to jump to next in the goalesce. And even then next one has no idea what's the intention - was it a bug that it didn't merge when one user asks "why isn't it merging". The fact that the behavior depends of the type (map or slice have different behaviors, so if for whatever reason someone changes the slices to be maps for some preprocessing requirements then the entire behavior will change) makes it even more difficult to remember these.

docs/content/en/tasks/monitor/_index.md
```diff
@@ -3,4 +3,203 @@ title: "Monitor K8ssandra"
 linkTitle: "Monitor"
 weight: 6
 description: "Access tools to monitor your Apache Cassandra® cluster running in Kubernetes."
----
\ No newline at end of file
+---
```
Miles-Garnsey, line 6: It looks like these docs were removed in a previous PR, I'm not sure if that was a good idea since some folks still need to use MCAC given that older patch versions aren't supported by the new metrics agent. 

Just getting thoughts here, not insisting it be left it.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -79,7 +162,7 @@ func (c Configurator) ReconcileTelemetryAgentConfig(dc *cassdcapi.CassandraDatac
 	recRes := reconciliation.ReconcileObject(c.Ctx, c.RemoteClient, c.RequeueDelay, *desiredCm)
 	switch {
 	case recRes.IsError():
-		fallthrough
```
Miles-Garnsey, line 82: @burmanm why did you change this to fallthrough? This will give you a `Done` result AFAIK, which is wrong if the ConfigMap reconciliation has failed.

Miles-Garnsey:
I have a few questions in my comments RE docs, and some changes that I didn't previously catch to the configmap reconciliation. 

Beyond that, I believe that this now works.

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
adejanovski: The behavior looks different than what was done by the original filters.
Instead of dropping by default all `org_apache_cassandra_metrics_table.*` metrics, you're dropping everything that has a `table` label, and then allow specific metrics by changing the `should_drop` label value.

Based on my testing, none of the metrics with a `table` label survives these filters, despite the following rules.

With the default filters on, I can only see 3 lines when looking for for `org_apache_cassandra_metrics_table_live_ss_table_count`:

```
# HELP org_apache_cassandra_metrics_table_live_ss_table_count_all 
# TYPE org_apache_cassandra_metrics_table_live_ss_table_count_all gauge
org_apache_cassandra_metrics_table_live_ss_table_count_all{host="67140d82-7ae5-45a0-b3c8-cac99530fde2",instance="172.24.0.4",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-1",node_name="k8ssandra-0-worker2",} 12.0
```

But without the filters, I see 45 such lines. Here's a sample:

```
# HELP org_apache_cassandra_metrics_table_live_ss_table_count 
# TYPE org_apache_cassandra_metrics_table_live_ss_table_count gauge
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="built_views",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="view_builds_in_progress",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_traces",table="sessions",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_schema",table="aggregates",} 2.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="available_ranges",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="size_estimates",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_schema",table="dropped_columns",} 2.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_auth",table="role_members",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="transferred_ranges_v2",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="peers",} 2.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_schema",table="tables",} 2.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system_auth",table="roles",} 3.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="peer_events_v2",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="peer_events",} 0.0
org_apache_cassandra_metrics_table_live_ss_table_count{host="1f8db44a-2fb4-4a65-8f60-82fdb75e3b07",instance="172.24.0.5",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker3",keyspace="system",table="local",} 1.0
```

The above metrics should survive the filters, given the name of the metrics, but they don't 🤔 
So, too many metrics are being dropped.

Switching the first relabeling to: 

```
- regex: org_apache_cassandra_metrics_table.*
  replacement: "true"
  sourceLabels:
  - __name__
  targetLabel: should_drop
```

generates even weirder behaviors, since I now lose all the `live_ss_table_count` metrics, but end up with metrics that have both `should_drop=true` and `should_drop=false` 😆 

```
org_apache_cassandra_metrics_table_all_memtables_live_data_size_all{host="af660e57-38c4-4863-af8a-31c08a4314dc",instance="172.24.0.4",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker2",should_drop="true",should_drop="false",} 125755.0
```

@burmanm, there seems to be something wrong with the relabelings here.
Could you tell us what you think?

docs/content/en/tasks/monitor/_index.md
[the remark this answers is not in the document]
Miles-Garnsey: Ah, I'll delete it again, thanks!

Miles-Garnsey:
> Instead of dropping by default all org_apache_cassandra_metrics_table.* metrics, you're dropping everything that has a table label, and then allow specific metrics by changing the should_drop label value.

The former part of the solution was @burmanm's suggestion, the latter part is required due to the differences in the way allow/deny rules work in MCAC (additive) vs the new metrics agent (subtractive).

w.r.t. some metrics not appearing, that's exactly what I was hoping you could confirm @adejanovski. I do see table related metrics appearing, however, having just re-tested I've realised that you're right - the ones I can see don't have the `table` label set. They appear to be keyspace related metrics which are just named similarly to table related metrics (e.g. `org_apache_cassandra_metrics_table_read_latency_all_bucket`).

So it appears that something is indeed wrong.

Miles-Garnsey:
Coming back to this now that this [PR](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/278) has been merged. 

The previous iteration of management api's new metrics agent had some issues where a capitalised metric name sometimes appeared, which meant that our default rules in this PR did not recognise the metric name via regex and failed to mark it with `should_drop=false`. This led to metrics being expectedly dropped.

I have manually tested and found that:

1. The metrics we do not want dropped (e.g. `org_apache_cassandra_metrics_table_live_ss_table_count` for `table="views"`) are not dropped: `org_apache_cassandra_metrics_table_live_ss_table_count{host="9fb6377b-62ba-4c7d-a869-ff0c926036ed",instance="10.244.4.4",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker4",keyspace="system_schema",table="views",should_drop="false",} 1.0`.
2. Metrics we do want dropped (e.g. CompressionRatio at table level) do not appear in the metrics (see below grep to confirm):

```
$ curl localhost:9000/metrics | grep org_apache_cassandra_metrics_table_compression
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0# HELP org_apache_cassandra_metrics_table_compression_ratio
# TYPE org_apache_cassandra_metrics_table_compression_ratio gauge
org_apache_cassandra_metrics_table_compression_ratio{host="9fb6377b-62ba-4c7d-a869-ff0c926036ed",instance="10.244.4.4",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker4",} 0.24513263925651432
# HELP org_apache_cassandra_metrics_table_compression_metadata_off_heap_memory_used
# TYPE org_apache_cassandra_metrics_table_compression_metadata_off_heap_memory_used gauge
org_apache_cassandra_metrics_table_compression_metadata_off_heap_memory_used{host="9fb6377b-62ba-4c7d-a869-ff0c926036ed",instance="10.244.4.4",cluster="test",datacenter="dc1",rack="default",pod_name="test-dc1-default-sts-0",node_name="k8ssandra-0-worker4",} 112.0
100 6998k  100 6998k    0     0  29.4M      0 --:--:-- --:--:-- --:--:-- 29.5M
```

These tests are run with `serverImage: docker.io/datastax/dse-mgmtapi-6_8:6.8.33-ubi7-v0.1.59`.

I believe that makes this PR ready for merge.

burmanm (approved):

PR #837, opened 2023-01-30
Add topology env variables to the Cassandra vector agent

### LINKED ISSUE #831: Add CLUSTER, DC and RACK env vars to the Vector agent sidecar container
As we extract container level metrics (cpu, load, ...) using the host_metrics source in the Vector agent, we need to enrich the metrics with cluster, dc and rack information.
This will allow to make the dashboards relevant and align with Cassandra level metrics.

Once the env vars are added, we'll get to enrich the metrics using [the remap transformer](https://vector.dev/docs/reference/configuration/transforms/remap/#examples-modify-metric-tags).

### LINKED ISSUE #836: Support dc name overrides in cass-operator
cass-operator recently got a new feature which allows to override a Cassandra DC name with a value that's different than the cassdc object name. This allows creating Cassandra DCs with names that aren't valid kubernetes names, with spaces, uppercase characters, ...

We need to expose this feature in k8ssandra-operator through a new `DatacenterOptions.DatacenterName` field which will only be considered at the DC level, and not the global level (obviously).

adejanovski:
**What this PR does**:
Adds `CLUSTER_NAME`, `DATACENTER_NAME` and `RACK_NAME` env variables to the Vector agent container, allowing to enrich the extracted metrics/logs with these informations as labels.
Support DC names overrides in cass-operator.

Fixes #836 
Fixes #831 

**Checklist**
- [x] Changes manually tested


pkg/cassandra/datacenter.go
```diff
@@ -124,6 +125,20 @@ type DatacenterConfig struct {
 	InitialTokensByPodName map[string][]string
 }
 
+// SanitizedName returns a sanitized version of the name returned by CassClusterName()
+func (in *DatacenterConfig) SanitizedName() string {
```
olim7t, line 129: I don't see this getting called anywhere, yet the CRD description says it's used to name Kubernetes objects.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -422,6 +422,13 @@ type DatacenterOptions struct {
 
 	// The k8s service account to use for the Cassandra pods
 	ServiceAccount string `json:"serviceAccount,omitempty"`
+
+	// DatacenterName allows to override the name of the Cassandra datacenter. Kubernetes objects will be named after a sanitized version of it if set, and if not metadata.name. In Cassandra the DC name will be overridden by this value.
+	// It may generate some confusion as objects created for the DC will have a different name than the CasandraDatacenter object itself.
```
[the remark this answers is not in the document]
adejanovski: > does this mean that objects created for the DC will use DatacenterName, but the CassandraDatacenter still uses metadata.name?

Yes, and that's a decision we made a while ago for cluster name overrides in K8ssandraCluster objects, which we're trying to be consistent with.
While this may seem confusing, there's another way to view this which is that clusters aren't referenced by their Kubernetes technical name but rather the Cassandra cluster name. Naming the objects accordingly makes it easier to spot which pod/configmap relates to what cluster.
Not an easy call, and opinions surely differ regarding what the "right" decision should be here.
Note that overrides aren't common and that sane operators would use a K8ssandraCluster name that is a sanitized version of the override, hopefully.

pkg/cassandra/datacenter.go
adejanovski: Indeed, I think I initially imagined I would need the sanitized name, but in the end I only need the non sanitized version of the datacenter name to populate the env variable.
I'll remove the function.

olim7t (approved):

PR #849, opened 2023-02-06
Prototype: generic reconciliation function

### LINKED ISSUE #827: Generic Reconciliation function
**What is missing?**

There is a lot of duplication in our reconciliation logic. Would it be possible to create a ReconcileAllObjects function which takes a slice of different k8s API objects and brings them all into the desired state based on their name?

Miles-Garnsey:
**What this PR does**:

Prototype of generic reconciliation function.

Fixes #827 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/reconciliation/reconciliation.go
+++ b/pkg/reconciliation/reconciliation.go
@@ -11,20 +11,24 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type Reconcilable interface {
+type k8sResource interface {
 	client.Object
+	annotations.Annotated
 	DeepCopy()
 	DeepCopyInto(out *client.Object)
 }
 
-func ReconcileObject[G Reconcilable](ctx context.Context, desiredObj G, client client.Client, requeueDelay time.Duration) result.ReconcileResult {
+func ReconcileObject[G k8sResource](ctx context.Context,
+	desiredObj k8sResource,
+	currentObj *k8sResource,
+	client client.Client,
+	requeueDelay time.Duration) result.ReconcileResult {
 	objectKey := types.NamespacedName{
 		Name:      desiredObj.GetName(),
 		Namespace: desiredObj.GetNamespace(),
 	}
 	annotations.AddHashAnnotation(desiredObj)
-	var currentObj G
-	err := client.Get(ctx, objectKey, currentObj)
+	err := client.Get(ctx, objectKey, *currentObj)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			if err := client.Create(ctx, desiredObj); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reconciliation/reconciliation.go
```diff
@@ -0,0 +1,53 @@
...
+	requeueDelay time.Duration) result.ReconcileResult {
+	objectKey := types.NamespacedName{
+		Name:      desiredObj.GetName(),
+		Namespace: desiredObj.GetNamespace(),
+	}
+	annotations.AddHashAnnotation(desiredObj)
+	err := client.Get(ctx, objectKey, *currentObj)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			if err := client.Create(ctx, desiredObj); err != nil {
```
emerkle826, line 34: What about setting controller references here? For example, in the Stargate controller, we have a ConfigMap [reconcile function](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/stargate/vector.go#L49-L51) for reconciling the Vector ConfigMap for Stargate metrics. We set the config map's controller reference to the stargate object so that cleanup happens when necessary.

pkg/reconciliation/reconciliation.go
```diff
@@ -0,0 +1,53 @@
...
+	if !annotations.CompareHashAnnotations(*currentObj, desiredObj) {
+		resourceVersion := (*currentObj).GetResourceVersion()
+		desiredObj.DeepCopyInto(currentObj)
+		currentObj.SetResourceVersion(resourceVersion)
+		if err := client.Update(ctx, currentObj); err != nil {
+			return result.Error(err)
+		}
+		return result.RequeueSoon(requeueDelay)
+	}
+	return result.Done()
```
emerkle826, line 52: This is more of a question than a change request. Would it be better to return `result.Continue()` here? I can see both ways. I guess it just depends on who should own the logic of deciding when a reconcile is done.

emerkle826, line 52: Also, don't we need logic to delete objects here if they existed and are no longer desired?

emerkle826 (changes requested):

pkg/reconciliation/reconciliation.go
Miles-Garnsey: That part of the reconciliation process is discussed over [here](https://github.com/k8ssandra/k8ssandra-operator/issues/827#issuecomment-1418633875).

pkg/reconciliation/reconciliation.go
Miles-Garnsey: I'm certainly open to that. I guess the reconciliation of this particular object is complete, so `Done()` made some sense here. We should comment the reconcile result types so developers can have some sense of what they're intended to mean.

Miles-Garnsey: That's a good point for sure. I wonder how we should communicate that via the function's arguments... Pass an empty object just with name/namespace?

emerkle826: Yes, for what you are doing here, I think `Done()` makes the most sense.

emerkle826: After a small private discussion, I think an empty object works. It's certainly better than an explicit delete function, which would push duplicate logic out to code that is reconciling objects, having to decide if it wants a reconcile, or a special reconcile of removing an object. Handling it here somehow makes the most sense.

Miles-Garnsey: Having looked into this, it looks like we can't establish whether an object is empty, because there is no way to access the `.Spec` field on the `client.Object`, or express to the type system that the types of objects we will be reconciling will always have a `.Spec` field. 

There might be some way to solve this, but I think the type system is hurting us here again. We might need to handle deletions separately which is a pity.

Miles-Garnsey:
I can't progress this at the moment. I'll ask questions on Stack Overflow (oh the ignominy 😔) and see if someone external can helm me figure out why it isn't working.

PR #852, opened 2023-02-06
Check that the build does not generate additional git changes in CI (fixes #851)

### LINKED ISSUE #851: Check that the build does not generate additional git changes in CI
**What is missing?**
Our build generates source files, such as CRDs. Some of these files are tracked in git. I propose that we add a check in CI that everything was checked in properly, i.e. the build process does not "dirty" the local working copy.

**Why do we need it?**
This is easy to overlook. The changes usually end up getting committed (consciously or not) as part of another PR, but that's not good hygiene, for example if we need to revert stuff later.
Also, it will help detect when someone has the wrong version of a tool (e.g. `controller-gen`) in their local cache.

olim7t:
Fixes #851

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/test_and_build_image.yaml
+++ b/.github/workflows/test_and_build_image.yaml
@@ -41,6 +41,12 @@ jobs:
         run: |
           export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
           make envtest test
+      - name: Check that the build did not generate additional git changes
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            git status
+            exit 1
+          fi
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
   build_image:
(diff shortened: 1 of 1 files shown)


olim7t:
Currently failing since there are uncommitted changes.

adejanovski (commented):
Looks good, but it seems like the build is creating additional files, right?
Maybe we should add those to `.gitignore` ?

olim7t:
> Looks good, but it seems like the build is creating additional files, right? Maybe we should add those to `.gitignore` ?

Hrmm yeah, at first there were some modified manifests coming from 4443fc6, but I rebased and they must have been committed in the meantime.
Now I see `install-man-page.sh`, `yq.1`and `yq_linux_amd64.tar.gz` that come from the yq install. We could work around that by using a temporary directory.

But also I'm starting to think that maybe we shouldn't commit generated manifests at all. Can we easily identify where they are located? If we add them to `.gitignore`, we wouldn't even need this PR.

adejanovski (approved):

adejanovski:
@olim7t, can this PR be merged now?

PR #857, opened 2023-02-07
Create a mutating webhook to mount secrets

### LINKED ISSUE #605: K8SSAND-1623 ⁃ Create a mutating webhook for the internal secrets provider

sseidman:
**What this PR does**:
Creates a webhook that, as a default, creates volumes and volume mounts on created/updated pods for secrets that are specified via a specific annotation.

For example, a pod with the annotation k8ssandra.io/inject-secret-cassandraSuperuser=[{"secretName": "cassandraSuperUser", "path": "/etc/credentials/cassandra"}] will create a volume for the cassandraSuperuser secret and mount the secret to the path /etc/credentials/cassandra on each container for the pod. This will mount secrets in the default Kubernetes style, which will create a single file for each key/value pair in the Secret (e.g. there will be username file and password file for credential sets).

Fixes #605

**Checklist**
- [x] Changes manually tested


controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	log "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/mutate-v1-pod-secrets-inject,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,admissionReviewVersions=v1,sideEffects=None
+
+func SetupSecretsInjectorWebhook(mgr ctrl.Manager) {
```
Miles-Garnsey, line 19: I think you'll need to call this as part of the envtest setup around [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/test/testenv.go#L104), I think that's what is causing the bulk of the test failures in the envtests right now.

Miles-Garnsey (changes requested):
I've left a couple of comments already. I have done a manual test and the webhook certainly appears to behave as expected.

There are a couple of additional areas I think we need to think about:

1. It would be good to have a CLI flag to disable the webhook for users who don't want it turned on.
2. There are test failures. I think I may have figured out the failures in the integration tests. I am still looking at the failures in the e2e tests. These are all multi-cluster, so I suspect that maybe the webhook isn't being installed on data planes or something, but I haven't explored that path yet.

I'll comment further tomorrow.

controllers/secrets-webhook/secretswebhook.go
Miles-Garnsey: Looks like you may need to do the same in the Medusa controller tests too FYI.

config/webhook/kustomization.yaml
```diff
@@ -20,3 +20,24 @@ patchesJson6902:
     - op: replace
       path: /webhooks/0/clientConfig/service/name
       value: k8ssandra-operator-webhook-service
+
+# adding the objectSelector prevents the bootstrapping problem
+# where the mutation request for the operator pod would be 
+# sent before the operator pod is created
+patchesJson6902:
```
Miles-Garnsey, line 27: I'm still trying to get tests to run for me locally so I can debug your e2e tests which are failing. 

The observation I will make, is that there are no failures on the cluster scoped test. That leads me to wonder whether there is an issue around Rolebindings/ClusterRoleBindings which you need to address to ensure that your data plane deployments have perms to access Pods in namespaces other than their own.

I'll keep trying to get a kind cluster up with the objective of inspecting the errors as the test runs.

sseidman: yeah I'm having a bit of an issue running the e2e tests locally as well; they seem to just hang and never complete

sseidman: ah ok I think I've fixed it. I was missing a namespace patch in the test fixtures

controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+			"secret", secretName,
+			"secret path", mountPath,
+		)
+	}
+
+	return nil
+}
+
+// injectVolume attaches a volume to the pod spec
+func injectVolume(pod *corev1.Pod, volume corev1.Volume) {
```
Miles-Garnsey, line 124: Can we use the functions from [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L414) for some of this functionality?

controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+	if !hasVolume(pod.Spec.Volumes, volume) {
+		pod.Spec.Volumes = append(pod.Spec.Volumes, volume)
+	}
+}
+
+// injectVolumeMount attaches a volumeMount to all containers in the pod spec
+func injectVolumeMount(pod *corev1.Pod, volumeMount corev1.VolumeMount) {
```
Miles-Garnsey, line 131: Again, there might be functionality [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/cassandra/datacenter.go#L414) that could be used.

controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+
+type SecretInjection struct {
+	SecretName string `json:"secretName"`
+	Path       string `json:"path"`
+}
+
+// mutatePods injects the secret mounting configuration into the pod
+func (p *podSecretsInjector) mutatePods(ctx context.Context, pod *corev1.Pod, logger logr.Logger) error {
+	if pod.Annotations == nil {
+		logger.Info("no annotations exist")
```
Miles-Garnsey, line 74: If we're going to log here (and I think this logging might be a bit too chatty), can we include the pod name and namespace so that any unexpected behaviour can be traced?

controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := json.Unmarshal([]byte(secretsStr), &secrets); err != nil {
+		logger.Error(err, "unable to unmarhsal secrets annotation", "annotation", secretsStr)
+		return err
+	}
+
+	for _, secret := range secrets {
+		// get secret name from injection annotation
+		secretName := secret.SecretName
+		mountPath := secret.Path
+		logger.Info("creating volume and volume mount for secret",
```
Miles-Garnsey, line 94: Again, it would be good to include the pod name here.

controllers/secrets-webhook/secretswebhook_test.go
```diff
@@ -0,0 +1,134 @@
...
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{{
+				Name: "test",
+			}},
+		},
+	}
+
+	p := &podSecretsInjector{}
+
+	ctx := context.TODO()
```
Miles-Garnsey, line 56: I think tests normally use context.Background().

controllers/secrets-webhook/secretswebhook_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	log "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func TestInjectSecretsc(t *testing.T) {
```
Miles-Garnsey, line 13: Could we possibly take a look at the test names? I think this one might be better described by something like `Test_mutatePods_OneSecret` or similar. Tests normally have `Test_` as a prefix, and it is good for the test name to explicitly mention the function name.

controllers/secrets-webhook/secretswebhook_test.go
```diff
@@ -0,0 +1,134 @@
...
+	p := &podSecretsInjector{}
+
+	ctx := context.TODO()
+	err := p.mutatePods(ctx, pod, log.FromContext(ctx))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, want, pod)
+}
```
Miles-Garnsey, line 134: Would it be possible to engineer tests for both the happy and failure paths of `Handle()`? There might be some need for mocks to make that fly, but I think it would be worthwhile.

Miles-Garnsey:
@sseidman, nice work tracking down the e2e failures! I've completed my review now. This broadly looks really good, there are only one or two naming, unit testing and refactoring related things I'm going to suggest.

Miles-Garnsey:
@sseidman, apologies, I totally missed that you'd been pushing some commits over here. I'll follow up with a review tomorrow.

controllers/secrets-webhook/secretswebhook.go
```diff
@@ -0,0 +1,162 @@
...
+
+	marshaledPod, err := json.Marshal(copy)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+}
+
+// e.g. k8ssandra.io/inject-secret=[{secretName=my-secret, path=/etc/credentials/cassandra}]
```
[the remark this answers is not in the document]
Miles-Garnsey: Just mentioning this again, the format in the commented example didn't work for me.

Miles-Garnsey (approved):
Looks perfect, thanks for addressing all my comments from the previous iterations!

PR #858, opened 2023-02-07
Remove current usages of managed-by label (fixes #859)

### LINKED ISSUE #859: Remove current usages of managed-by label
We currently use `app.kubernetes.io/managed-by` (and its associated functions such as `labels.IsManagedBy`):
- for replicated secrets: as part of the selector expression that determines which secrets to replicate;
- for per-node config maps and Vector config maps: as a way to set the two other label that allows us to "watch" those components in the `K8ssandraCluster` controller.

`managed-by` is part of the [Recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) that Kubernetes suggests applying to every component; it should be for information purposes only, associating semantic meaning with it can be misleading.

We propose removing its current usages:
- for replicated secrets, replace it with a custom label.
- for config maps, it is not needed, the other two are sufficient (and in fact the code that sets up the watch in the controller does not even test for it).

`managed-by` will be reintroduced in #861, but as a common label that gets applied to every component, without any underlying semantics.

olim7t:
**What this PR does**:

Fixes #859 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/vector.go
+++ b/controllers/k8ssandra/vector.go
@@ -37,7 +37,7 @@ func (r *K8ssandraClusterReconciler) reconcileVector(
 		}
 
 		desiredVectorConfigMap := telemetry.BuildVectorAgentConfigMap(namespace, kc.SanitizedName(), dcConfig.Meta.Name, kc.Namespace, toml)
-		k8ssandralabels.SetManagedBy(desiredVectorConfigMap, kcKey)
+		k8ssandralabels.SetWatchedByK8ssandraCluster(desiredVectorConfigMap, kcKey)
 
 		// Check if the vector config map already exists
 		recRes := reconciliation.ReconcileConfigMap(ctx, remoteClient, r.DefaultDelay, *desiredVectorConfigMap)
--- a/pkg/secret/replicated.go
+++ b/pkg/secret/replicated.go
@@ -168,7 +168,7 @@ func generateReplicatedSecret(kcKey client.ObjectKey, replicationTargets []repli
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      kcKey.Name,
 			Namespace: kcKey.Namespace,
-			Labels:    labels.ManagedByLabels(kcKey),
+			Labels:    labels.WatchedByK8ssandraClusterLabels(kcKey),
 		},
 		Spec: replicationapi.ReplicatedSecretSpec{
 			Selector: &metav1.LabelSelector{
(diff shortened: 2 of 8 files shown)


controllers/k8ssandra/vector.go
```diff
@@ -38,7 +38,7 @@ func (r *K8ssandraClusterReconciler) reconcileVector(
 
 		desiredVectorConfigMap := telemetry.BuildVectorAgentConfigMap(namespace, kc.SanitizedName(), dcConfig.Meta.Name, kc.Namespace, toml)
 		annotations.AddHashAnnotation(desiredVectorConfigMap)
-		k8ssandralabels.SetManagedBy(desiredVectorConfigMap, kcKey)
+		k8ssandralabels.SetWatchedByK8ssandraCluster(desiredVectorConfigMap, kcKey)
```
adejanovski, line 41: We should be able to remove this as we already put the corresponding labels when creating the configmap in `BuildVectorAgentConfigMap()` [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/pkg/telemetry/vector.go#L163-L170).

pkg/secret/replicated.go
```diff
@@ -161,10 +165,14 @@ func HasReplicatedSecrets(ctx context.Context, c client.Client, kcKey client.Obj
 
 func generateReplicatedSecret(kcKey client.ObjectKey, replicationTargets []replicationapi.ReplicationTarget) *replicationapi.ReplicatedSecret {
 	return &replicationapi.ReplicatedSecret{
-		ObjectMeta: getManagedObjectMeta(kcKey.Name, kcKey),
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kcKey.Name,
+			Namespace: kcKey.Namespace,
+			Labels:    labels.WatchedByK8ssandraClusterLabels(kcKey),
```
adejanovski, line 171: Shouldn't we consider that a ReplicatedSecret is "part of" the K8ssandraCluster? Since it's been created by the controller, I'd set it part of, which will naturally add the WatchedBy labels.
I think WatchedBy should only be applied to externally provided resources.

olim7t: A lot of the other label sets are supersets of "watched by". So yes, we could document that they "imply" it, and avoid redundant declaration.
I'd like to address that after #860 if that's OK.

adejanovski: sounds good

adejanovski (approved):

adejanovski (approved):

PR #866, opened 2023-02-08
Configmap reconciler

### LINKED ISSUE #827: Generic Reconciliation function

### LINKED ISSUE #843: Refactor duplicated ConfigMap code
With the introduction of Vector metrics collection, we have added duplicated code around handling of ConfigMaps. Duplicate code increases the maintenance burden, so it would be good to try to consolidate some of that code.

For example:
We have this to deal with Vector ConfigMaps  for Cassandra:
https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/k8ssandra/vector.go#L19
And this to deal with them for Stargate:
https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/stargate/vector.go#L22
And are about to duplicate it again for Reaper:
https://github.com/k8ssandra/k8ssandra-operator/pull/838/files#diff-49b4878a0509bc364d174219a6906417836a62323a8fef85c0cd2b035c42635dR22

Miles-Garnsey:
**What this PR does**:

Reconciles configmaps.

Partially Fixes #827, completely fixes #843

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/telemetry/cassandra_agent/cassandra_agent_config.go
+++ b/pkg/telemetry/cassandra_agent/cassandra_agent_config.go
@@ -8,12 +8,11 @@ import (
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	k8ssandraapi "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	telemetryapi "github.com/k8ssandra/k8ssandra-operator/apis/telemetry/v1alpha1"
-	"github.com/k8ssandra/k8ssandra-operator/pkg/annotations"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/labels"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/reconciliation"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/result"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


pkg/telemetry/cassandra_agent/cassandra_agent_config.go
```diff
@@ -71,41 +70,19 @@ func (c Configurator) ReconcileTelemetryAgentConfig(dc *cassdcapi.CassandraDatac
...
-		if err := c.RemoteClient.Update(c.Ctx, currentCm); err != nil {
-			return result.Error(err)
-		}
-		return result.RequeueSoon(c.RequeueDelay)
+	recRes := reconciliation.ReconcileConfigMap(c.Ctx, c.RemoteClient, c.RequeueDelay, *desiredCm)
+	switch {
+	case recRes.IsError():
+		return recRes
+	case recRes.IsRequeue():
+		return recRes
```
adejanovski, line 85: nit: Why us a switch instead of an `if` ?

```suggestion
	if recRes.IsError() || recRes.IsRequeue(){
		return recRes
	}
```

adejanovski (commented):
This is looking good!

There are other configmaps that are reconciled [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/k8ssandra/medusa_reconciler.go#L89) and [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/stargate/stargate_controller.go#L411) for example.

Do you think they could use your function?

pkg/telemetry/cassandra_agent/cassandra_agent_config.go
Miles-Garnsey: Firstly, because I think it is easier to read than an if/else. Secondly, because longer term I think we should think about making the reconcile.Result types public so that we don't have to have so many silly isDone/isRequeue type helper functions, as they aren't that maintainable.

Miles-Garnsey: Oh, and also, the switch is needed in other places where you (for e.g.) don't return an error. I was just keeping it consistent in this case where you directly return the result.

Miles-Garnsey:
> There are other configmaps that are reconciled [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/k8ssandra/medusa_reconciler.go#L89) and [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/stargate/stargate_controller.go#L411) for example.

Yes, probably, can we get this one merged first though and do them as a second PR? I'm keen to be incremental with refactorings.

adejanovski:
> > There are other configmaps that are reconciled [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/k8ssandra/medusa_reconciler.go#L89) and [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/stargate/stargate_controller.go#L411) for example.
> 
> Yes, probably, can we get this one merged first though and do them as a second PR? I'm keen to be incremental with refactorings.

works for me 👍

adejanovski (approved):
Awesome stuff!

PR #873, opened 2023-02-10
More ConfigMap refactoring + handle edge case

### LINKED ISSUE #843: Refactor duplicated ConfigMap code

Miles-Garnsey:
**What this PR does**:

More refactoring for ConfigMaps to address PR feedback from #866. Uses the new reconciliation function for Stargate and Medusa ConfigMaps.

Also fixes a potential edge case where timing issues might prevent creation of a ConfigMap when the call to `Get` fails with a not found, but the call to `Create` then fails with an "already exists" error. 

This logic was not previously used in every reconciliation, so was missed when building the general reconciler.

Fixes #843

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/medusa_reconciler.go
+++ b/controllers/k8ssandra/medusa_reconciler.go
@@ -6,14 +6,12 @@ import (
 
 	"github.com/go-logr/logr"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
-	"github.com/k8ssandra/k8ssandra-operator/pkg/annotations"
 	cassandra "github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
 	medusa "github.com/k8ssandra/k8ssandra-operator/pkg/medusa"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/reconciliation"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/result"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/secret"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


Miles-Garnsey:
A couple of notes here. For the changes to `[controllers/stargate/stargate_controller.go](https://github.com/k8ssandra/k8ssandra-operator/pull/873/files#diff-200750b6d0cbf8caf7d94d1b07fe07f2634caa507c62b3e5334999a3e05e1bab)`: there is something strange going on in that reconciliation. It doesn't set a managed by reference or anything - I don't know how it gets GC'd, but it does manually set part-of etc. when creating the `desiredConfigMap` so maybe that's it.


In the Medusa one, also something weird. There is [no explicit cleanup](https://github.com/k8ssandra/k8ssandra-operator/pull/873/files#diff-f67b7b530aca36d67953e11a3fe941fcaca3bd934c66d681973b6940bcb63065R95) if `kc.Spec.Medusa` is nil. 

In any event I don't think I'm breaking anything any worse than it might already be broken, I might take a deeper look on Monday though and confirm if we have a test asserting that these resources are cleaned up under the right conditions.

PR #926, opened 2023-03-24
Don't set jmxremote.authenticate system property by default

olim7t:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


adejanovski:
By making this change we're changing our defaults. I think we should allow to opt out of having jmx auth enabled instead of changing the default, which will trigger a rolling restart upon upgrade ;)

olim7t:
> By making this change we're changing our defaults. I think we should allow to opt out of having jmx auth enabled instead of changing the default, which will trigger a rolling restart upon upgrade ;)

I don't know, that sounds overkill for such a small change.

adejanovski:
@olim7t, should we close this and fully remove changes we make on the JMX config now that Reaper moved to using the management api?

PR #936, opened 2023-03-29
Add updated dashboard for the new metrics endpoint

### LINKED ISSUE #935: The dashboards don't work with the new metrics names
The new metrics endpoint exposes different metrics names than MCAC, which prevents the documented dashboards from displaying the metrics.

We need to provide an updated dashboard using the new names.

adejanovski:
**What this PR does**:
Provides an updated dashboard for the new metrics endpoint.
Adds proper documentation for setting it up.

Fixes #935

**Checklist**
- [x] Changes manually tested
- [ ] ~~Automated Tests added/updated~~

### DIFF
--- a/docs/content/en/tasks/monitor/prometheus-grafana/_index.md
+++ b/docs/content/en/tasks/monitor/prometheus-grafana/_index.md
@@ -160,26 +160,27 @@ The new metrics endpoint doesn't extract os level metrics (cpu, memory, disk, et
 An alternative is to enable Vector instead and include the following components in the Vector configuration:
 
 ```
-hostMetricsSource := telemetryapi.VectorSourceSpec{
-		Name: "host_metrics",
-		Type: "host_metrics",
-		Config: `filesystem.devices.excludes = ["binfmt_misc"]
-filesystem.filesystems.excludes = ["binfmt_misc"]
-filesystem.mountpoints.excludes = ["*/proc/sys/fs/binfmt_misc"]
-scrape_interval_secs = 30`,
-	}
-
-	enrichHostMetricsTransform := telemetryapi.VectorTransformSpec{
-		// We enrich the host metrics with the cluster, datacenter and rack information which are available as env variables in the Vector container
-		Name:   "enrich_host_metrics",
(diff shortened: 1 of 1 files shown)


docs/content/en/tasks/monitor/prometheus-grafana/_index.md
```diff
@@ -144,12 +144,47 @@ test-dc1-stargate-stargate-servicemonitor            5m47s
...
 ``` 
 
+**When using the new metrics endpoint** (and MCAC is disabled), create the overview, condensed and stargate dashboards (download the manifest [here](grafana-dashboards-new.yaml)) configmaps:
+
+```bash
+kubectl apply -f grafana-dashboards-new.yaml -n k8ssandra-operator
+```
+
+The new metrics endpoint doesn't extract os level metrics (cpu, memory, disk, etc) so the overview dashboard will be missing those metrics by default. The Prometheus node_exporter can be deployed to expose those metrics for scraping, and the corresponding charts will have to be adjusted accordingly in the dashboard.  
+An alternative is to enable Vector instead and include the following components in the Vector configuration:
```
burmanm, line 160: This assumes user must understand Go struct -> YAML to use this information. Could we just provide yaml here?

burmanm: This is open source repo @Miles-Garnsey ;)

Miles-Garnsey: Hah, now I'm the one getting confused as to which repo we're in. In that case, that being the case, I kind of stand by keeping the API references but agree that maybe we should add a yaml example since we don't get as much docs support over here in OSS land.

adejanovski: > This assumes user must understand Go struct -> YAML to use this information. Could we just provide yaml here?

Good point!
I'll provide the yaml representation as well.

adejanovski: Done! (finally...)

PR #942, opened 2023-04-05
Add variable support to the secret webhook injector

### LINKED ISSUE #932: Enhance webhook to allow templating of pod name for per-pod secret injection
There is a requirement to allow each pod in our statefulsets to mount distinct keystores/truststores. 

To enable a keystore/truststore per Cassandra node, we need to mount a different secret per statefulset pod. Since the statefulset API doesn't allow doing this, we need to leverage the new secret injection webhook created in #605 for this.

We need to add a feature to the webhook so that the annotation can contain a placeholder for the pod name:

```
k8ssandra.io/inject-secret=[{"name": "encryption-stores-${POD_NAME}", "path": "/etc/credentials/cassandra"}]
```

For a pod named `test-dc2-default-sts-1`, the above would mount the `encryption-stores-test-dc2-default-sts-1` secret at `/etc/credentials/cassandra`. 

```[tasklist]
### Definition of Done
- [x] The json format of the annotation should be modified to accept "name" instead of "secretName" to avoid repeating "secret"
- [x] The webhook is capable of substituting ${POD_NAME} with the name of the pod being mutated in the secret name to mount
```

burmanm:
**What this PR does**:
Support variables in the annotation for secret-injection

Fixes #932 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/secrets-webhook/secretswebhook.go
+++ b/controllers/secrets-webhook/secretswebhook.go
@@ -3,9 +3,12 @@ package secrets_webhook
 import (
 	"context"
 	"encoding/json"
+	"net/http"
+	"os"
+	"strings"
+
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"net/http"
 
 	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


controllers/secrets-webhook/secretswebhook.go
```diff
@@ -61,11 +64,11 @@ func (p *podSecretsInjector) Handle(ctx context.Context, req admission.Request)
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
-// e.g. k8ssandra.io/inject-secret: '[{ "secretName": "test-secret", "path": "/etc/test/test-secret" }]'
+// e.g. k8ssandra.io/inject-secret: '[{ "name": "test-secret", "path": "/etc/test/test-secret" }]'
 const secretInjectionAnnotation = "k8ssandra.io/inject-secret"
 
 type SecretInjection struct {
-	SecretName string `json:"secretName"`
+	SecretName string `json:"name"`
```
olim7t, line 71: The ticket also suggests that the old name still be supported for backward compatibility. It looks like there's no way to have multiple name tags, so we could have both fields and pick the first non nil.

Although that raises the question of whether changing the name in the first place is worth it...

olim7t (changes requested):

controllers/secrets-webhook/secretswebhook.go
adejanovski: I'm to blame for trying to keep some backwards compatibility here. It's probably not worth it and we can move forward with just `name`.

PR #949, opened 2023-04-25
Fix MedusaController Kubernetes Client to fetch pods from one namespace

### LINKED ISSUE #946: MedusaBackupJob fetches backup pods from all namespaces if installed in cluster-scope
**What is missing?**
Curretnly, MedusaBackupSchedule will look for the cassandraDatacenter name only to launch the backup.
Maybe it's better to add more filter, like namespace.

**Why do we need it?**
In my environment, I create a k8ssandracluster per namespace. Each k8ssandracluster has the same cassandraDatacenter name.
It's impossible to set the backup separately for each k8ssandracluster in the different namespace.

**Environment**

kubelet V1.21.14-00

* K8ssandra Operator version: V1.5.2

May I missed something. Any other filter for MedusaBackupSchedule ?

burmanm:
**What this PR does**:
The medusa job controller fetched pods from all the namespaces matchi…ng Datacenter name. Add a namespace restrictor to fetch only the correct DC, not everything with the same name

Fixes #946 

**Checklist**
- [x] Changes manually tested


adejanovski (approved):
Successfully reviewed and tested the change.

PR #960, opened 2023-05-03
Refactor InitClientConfigs

### LINKED ISSUE #959: Refactor InitClientConfigs
Refactor InitClientConfigs to reduce code complexity:
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=k8ssandra_k8ssandra-operator&open=AYWmfaSCjyR6sSVTtRYY

emerkle826:
**What this PR does**:
Refactor InitClientConfigs to reduce code complexity. Should address this one:
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=k8ssandra_k8ssandra-operator&open=AYWmfaSCjyR6sSVTtRYY
Fixes #959 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/config/clientconfig_controller.go
+++ b/controllers/config/clientconfig_controller.go
@@ -134,7 +134,7 @@ func (r *ClientConfigReconciler) InitClientConfigs(ctx context.Context, mgr ctrl
 	r.secretFilter = make(map[types.NamespacedName]types.NamespacedName, len(clientConfigs))
 
 	for _, cCfg := range clientConfigs {
-		c, err := initAdditionalCLusterConfig(r, ctx, cCfg, mgr, watchNamespace)
+		c, err := r.initAdditionalClusterConfig(ctx, cCfg, mgr, namespaces)
 		if err != nil {
 			return nil, err
 		}
@@ -161,9 +161,8 @@ func calculateHashes(ctx context.Context, anyClient client.Client, clientCfg con
 }
 
 // initAdditionalCLusterConfig fetches the clientConfigs for additional clusters
-func initAdditionalCLusterConfig(r *ClientConfigReconciler, ctx context.Context, cCfg configapi.ClientConfig, mgr ctrl.Manager, watchNamespace string) (cluster.Cluster, error) {
+func (r *ClientConfigReconciler) initAdditionalClusterConfig(ctx context.Context, cCfg configapi.ClientConfig, mgr ctrl.Manager, namespaces []string) (cluster.Cluster, error) {
 	uncachedClient := r.ClientCache.GetLocalNonCacheClient()
-	namespaces := strings.Split(watchNamespace, ",")
 
 	// Calculate hashes
 	cCfgName := types.NamespacedName{Name: cCfg.Name, Namespace: cCfg.Namespace}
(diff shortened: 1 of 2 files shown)


controllers/config/clientconfig_controller.go
```diff
@@ -205,3 +159,61 @@ func calculateHashes(ctx context.Context, anyClient client.Client, clientCfg con
 
 	return cfgHash, secretHash, nil
 }
+
+// initAdditionalCLusterConfig fetches the clientConfigs for additional clusters
+func initAdditionalCLusterConfig(r *ClientConfigReconciler, ctx context.Context, cCfg configapi.ClientConfig, mgr ctrl.Manager, watchNamespace string) (cluster.Cluster, error) {
```
olim7t, line 164: C**L**uster => Cluster:
```suggestion
func initAdditionalClusterConfig(r *ClientConfigReconciler, ctx context.Context, cCfg configapi.ClientConfig, mgr ctrl.Manager, watchNamespace string) (cluster.Cluster, error) {
```
Also: this could be turned into a method instead of passing `r` as a parameter.

controllers/config/clientconfig_controller.go
```diff
@@ -205,3 +159,61 @@ func calculateHashes(ctx context.Context, anyClient client.Client, clientCfg con
...
+	uncachedClient := r.ClientCache.GetLocalNonCacheClient()
+	namespaces := strings.Split(watchNamespace, ",")
```
olim7t, line 166: You can pass `namespaces` as a parameter instead of `watchNamespace`, then the two lines that used the latter can be rewritten:
- `if strings.Contains(watchNamespace, ",") {` => `if len(namespaces) > 1 {`
- `o.Namespace = watchNamespace` => `o.Namespace = namespaces[0]`

I feel like we should probably check `len(namespaces) == 0` somewhere (the original code ignores it, but this probably causes a failure down the line).

olim7t (changes requested):

controllers/config/clientconfig_controller.go
emerkle826: `namespaces` is created from a string using `strings.Split`, so the only way it's length would be `0` is if the namespace string passed in was empty. 

> I feel like we should probably check `len(namespaces) == 0` somewhere (the original code ignores it, but this probably causes a failure down the line).

I agree. I looked a little at the flow, but it seems a little tricky how it handles watchNamepsace. In [main.go](https://github.com/k8ssandra/k8ssandra-operator/blob/main/main.go#L109-L115) if there is an error getting `watchNamespace`, the error indicates it will watch all namespaces. This seems to be done by returning an [empty string](https://github.com/k8ssandra/k8ssandra-operator/blob/main/main.go#L289-L300) as the `watchNamespace`. So I guess a length of 0 does come into InitClientConfigs.....

emerkle826: I guess my point is, it's a little more involved to handle length==0 it in this scope....

olim7t: > it will watch all namespaces. This seems to be done by returning an [empty string](https://github.com/k8ssandra/k8ssandra-operator/blob/main/main.go#L289-L300) as the watchNamespace

Ok, so in that case I think we need to change the first check to `if len(namespaces) != 1` to achieve the same result. Otherwise we'll enter the second branch and fail with an array index out of bounds.

olim7t: Actually forget this last comment.

`strings.split` returns the string itself if the separator is not present.
> If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.

So in this case we'll get an array of size 1 containing the empty string, and things work as expected.

olim7t (approved):

PR #962, opened 2023-05-03
Refactor/cleanup

### LINKED ISSUE #961: Refactor cleanup

emerkle826:
**What this PR does**:
Refactors [cleanup]() functions to reduce code complexity. 
Fixes #961 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/cleanup.go
+++ b/controllers/k8ssandra/cleanup.go
@@ -218,34 +218,22 @@ func (r *K8ssandraClusterReconciler) findStargateForDeletion(
 	dcName string,
 	remoteClient client.Client) (*stargateapi.Stargate, client.Client, error) {
 
-	selector := k8ssandralabels.PartOfLabels(kcKey)
-	options := &client.ListOptions{LabelSelector: labels.SelectorFromSet(selector)}
-	stargateList := &stargateapi.StargateList{}
 	stargateName := kcKey.Name + "-" + dcName + "-stargate"
-
 	if remoteClient == nil {
 		for _, remoteClient := range r.ClientCache.GetAllClients() {
-			err := remoteClient.List(ctx, stargateList, options)
+			stargate, err := finStargateForDeletionWithRemoteClient(ctx, kcKey, dcName, stargateName, remoteClient)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
 			}
-			for _, stargate := range stargateList.Items {
-				if stargate.Name == stargateName {
-					return &stargate, remoteClient, nil
-				}
-			}
+			return stargate, remoteClient, nil
 		}
 	} else {
-		err := remoteClient.List(ctx, stargateList, options)
+		stargate, err := finStargateForDeletionWithRemoteClient(ctx, kcKey, dcName, stargateName, remoteClient)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
 		}
 
-		for _, stargate := range stargateList.Items {
-			if stargate.Name == stargateName {
-				return &stargate, remoteClient, nil
-			}
-		}
+		return stargate, remoteClient, nil
 	}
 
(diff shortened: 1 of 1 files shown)


controllers/k8ssandra/cleanup.go
```diff
@@ -153,9 +153,144 @@ func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *ap
...
+		return result
+	}
+
+	// delete Reaper
+	if result := r.deleteReaper(ctx, kc, dcName, logger); result != nil {
+		return result
+	}
+
+	// delete the DC
+	if result := r.deleteDatacenter(ctx, kc, dcName, logger); result != nil {
```
olim7t, line 168: We need to improve naming here, there's a `deleteDc` and a `deleteDatacenter` and they do different things.
Maybe `deleteDcComponents` for the first one, and `deleteCassandraDatacenter` for the second (matches the name of the CR that it deletes)?

controllers/k8ssandra/cleanup.go
```diff
@@ -206,154 +361,64 @@ func (r *K8ssandraClusterReconciler) deleteDc(ctx context.Context, kc *api.K8ssa
...
+	return nil
 }
 
-func (r *K8ssandraClusterReconciler) findStargateForDeletion(
-	ctx context.Context,
-	kcKey client.ObjectKey,
-	dcName string,
-	remoteClient client.Client) (*stargateapi.Stargate, client.Client, error) {
-
+func (r *K8ssandraClusterReconciler) findRemoteClientForReaper(ctx context.Context, kcKey client.ObjectKey, dcName string) (client.Client, error) {
```
olim7t, line 367: We know the remote client is going to be the same for all components.

In the original code, as soon as we found a client for Stargate or Reaper, it would get passed to the following methods instead of looking it up again. I think we should preserve that in some form.

In fact we can even improve the lookup, because the current approach is very inefficient. We don't need to look for the components to delete in every context, we already know the context because it's declared in the DC's `K8sContext` field.
The only problem is that we only have the DC name right now, but that's an easy fix:
- in `checkDcDeletion()`, either modify `GetDatacenterForDecommission` so that it returns the full `CassandraDatacenterTemplate`, or re-iterate `kc.Spec.Cassandra.Datacenters` to look it up by name.
- then look up the client:
    ```go
    remoteClient, err := r.ClientCache.GetRemoteClient(dcTemplate.K8sContext)
    if err != nil { ... }
    ```
- and finally pass it into `deleteDc` and its child methods.

That should allow us to remove all the `findRemoteClient` methods.

controllers/k8ssandra/cleanup.go
```diff
@@ -153,9 +153,144 @@ func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *ap
...
+
+	err := remoteClient.List(ctx, stargateList, options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
+	}
+	for _, stargate := range stargateList.Items {
+		if stargate.Name == stargateName {
+			return &stargate, nil
+		}
+	}
```
olim7t, line 198: (Also applied to the original code) Can't we do a simple lookup by key with `remoteClient.Get()` instead of this?
We just need the namespace but it should be the same as `kcKey` if I'm not mistaken.

olim7t (changes requested):

controllers/k8ssandra/cleanup.go
emerkle826: Ok, I had trouble figuring out this bit, and the old code was confusing. But this makes sense now.

olim7t: > The only problem is that we only have the DC name right now, but that's an easy fix:
> 
> * in `checkDcDeletion()`, either modify `GetDatacenterForDecommission` so that it returns the full `CassandraDatacenterTemplate`, or re-iterate `kc.Spec.Cassandra.Datacenters` to look it up by name.

Actually it's not that easy, because the DC has already been removed from `kc.Spec.Cassandra.Datacenters` by the time we start the cleanup.

PR #980, opened 2023-05-24
Standalone medusa pod

### LINKED ISSUE #483: K8SSAND-1354 ⁃ Deploy a Medusa grpc pod for operator direct interactions

### LINKED ISSUE #486: K8SSAND-1357 ⁃ Create Medusa restore mapping in the operator
Depending on the results of #485, we should be able to compute restore mappings in the operator directly.
After getting the backup tokenmap from the storage backend, we can generate a mapping file in a config map that will be mounted on the pods for Medusa to read it.
The mapping should look like the following:

```
{
  'in_place': false,
  'host_map':
  {
     'target-node-0': 'source-node-1',
     'target-node-1': 'source-node-3',
     'target-node-2': 'source-node-0',
    ...
  }
}
```

The mapping should respect rack assignments and detecting whether restore is in place or not can be based on overlaps between source node names and projected target pod names.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1357) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1357
┆priority: Medium

### LINKED ISSUE #821: Medusa 0.14.0 breaks restore mappings
**What happened?**
The latest release of Medusa, v0.14.0, broke restore mappings by selecting the same node (the last one) as source for mapping to each of the existing pods.
This results in broken clusters after a restore.

Restore mappings should be done consistently with one source node mapping to a single target node. This was working correctly in Medusa v0.13.4.

**Environment**

* K8ssandra Operator version:

    `v1.4.0`

### LINKED ISSUE #990: TestHandleInjectSecretSuccess is flaky
**What happened?**
The order of volumes / volumeMounts is not guaranteed in the output of the method, so the test occasionally fails to the following:

```
{[{add /spec/containers/0/volumeMounts [map[mountPath:/my/secret/path name:mySecret-secret]]} {add /spec/volumes [map[name:mySecret-secret secret:map[secretName:mySecret]]]}] { true nil [] 0xc0004bef20 map[] []}}
--- FAIL: TestHandleInjectSecretSuccess (0.01s)
    secretswebhook_test.go:45: 
                Error Trace:    /Users/michael.burman/projects/datastax/k8ssandra-operator/controllers/secrets-webhook/secretswebhook_test.go:45
                Error:          Not equal: 
                                expected: "/spec/containers/0/volumeMounts"
                                actual  : "/spec/volumes"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -/spec/containers/0/volumeMounts
                                +/spec/volumes
                Test:           TestHandleInjectSecretSuccess
    secretswebhook_test.go:47: 
                Error Trace:    /Users/michael.burman/projects/datastax/k8ssandra-operator/controllers/secrets-webhook/secretswebhook_test.go:47
                Error:          Not equal: 
                                expected: "/spec/volumes"
                                actual  : "/spec/containers/0/volumeMounts"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -/spec/volumes
                                +/spec/containers/0/volumeMounts
                Test:           TestHandleInjectSecretSuccess
```

**Environment**

* K8ssandra Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* K8ssandra Operator Logs:

```
```

adejanovski:
**What this PR does**:
The creation of the main Medusa container has been separated from its update process. The medusa.CreateMedusaMainContainer method now handles creation, while medusa.UpdateMedusaMainContainer only updates the created container.

The creation and updating of Medusa volumes has been improved. Volumes are generated using the medusa.GenerateMedusaVolumes method and then updated or added to the CassandraDatacenter configuration (dcConfig).

Both these changes allow creating a standalone Medusa pod that is using the medusa.StandaloneMedusaDeployment method, by reusing the previously created Medusa main container and volumes.

**14/6/23 update:** The `local` storage option for Medusa has been removed. It was only available for testing purposes but since we now use a separate pod to compute restore mappings, local storage is no longer an option. As such, there are no more additionalVolumes added to the CassandraDatacenter definition as part of reconciling Medusa. **This is a breaking change**.

A standalone Medusa service is also created and reconciled. The medusa.StandaloneMedusaService method creates the service, which is then reconciled like other resources.

The reconciliation process for the Medusa ConfigMap has been updated. It now includes a key to the K8ssandraCluster resource, which can be used for owner references or other purposes.
Support was added for the definition of readiness and liveness probe settings for Medusa containers. This allows developers to specify custom probe parameters, such as initialDelaySeconds, periodSeconds, failureThreshold, successThreshold, and timeoutSeconds.
In the cleanup.go file, the checkDeletion function now includes calls to the newly introduced deleteDeployments and deleteServices methods. These methods aim to delete all Kubernetes Services and Deployments associated with the K8ssandraCluster in question. If the deletion of any of these resources fails, an error is logged and the function will return true, indicating the presence of errors.
The deleteServices method lists all services part of the K8ssandraCluster, as determined by label selectors, and attempts to delete them. Similarly, the deleteDeployments method lists and tries to delete all deployments associated with a deleted K8ssandraCluster.

~~This pull request also introduces an enhancement to the generic reconciliation process. The changes aim to improve the labeling of reconciled objects, linking them back to the associated K8ssandra object:
The ReconcileObject function signature has been updated to take an additional parameter, klKey. This parameter represents the key of a K8ssandra object, which includes its name and namespace.
Inside the ReconcileObject function, if a klKey is provided, the reconciled object is given a new set of labels. These labels are generated using the labels.PartOfLabels function with the provided klKey. These labels signify that the reconciled object is part of the associated K8ssandra object which then allows proper cleanup of resources upon deletion of the K8ssandraCluster object.~~

Fixes #483
Fixes #486 
Fixes #821
Fixes #990

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -44,9 +46,27 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
...
 
+		medusaContainer := medusa.CreateMedusaMainContainer(dcConfig, medusaSpec, kc.Spec.UseExternalSecrets(), kc.SanitizedName(), logger)
 		medusa.UpdateMedusaInitContainer(dcConfig, medusaSpec, kc.Spec.UseExternalSecrets(), kc.SanitizedName(), logger)
-		medusa.UpdateMedusaMainContainer(dcConfig, medusaSpec, kc.Spec.UseExternalSecrets(), kc.SanitizedName(), logger)
-		medusa.UpdateMedusaVolumes(dcConfig, medusaSpec, kc.SanitizedName())
+		medusa.UpdateMedusaMainContainer(dcConfig, medusaContainer)
+
+		// Create required volumes for the Medusa containers
+		volumes, additionalVolumes := medusa.GenerateMedusaVolumes(dcConfig, medusaSpec, kc.SanitizedName())
+		for _, volume := range volumes {
```
Miles-Garnsey, line 55: Not a big fan of using both volumes and additionalVolumes. Can't we just use one or the other?

pkg/medusa/reconcile.go
```diff
@@ -9,25 +9,37 @@ import (
 	k8ss "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
```
Miles-Garnsey, line 9: Could we please split this file up, it doesn't have great cohesion and 'reconcile' is too broad a topic for a single file.

Miles-Garnsey (changes requested):
I've left a few comments, not yet sure how to test this or whether you're ready for me to kick off manual tests.

I have one more general comment too, which is non-blocking, but hrmmmm... The way you're creating the Medusa containers/deployment feels convoluted/tortured to me. Each of them seems to go through multiple phases of mutation, e.g. `UpdateMedusaInitContainer`, `medusaInitContainerResources`, `CreateMedusaMainContainer`, `setImage`, `medusaVolumeMounts`, `GenerateMedusaVolumes`, `medusaEnvVars`, 

There are also code paths that may not be needed anymore. E.g. do we still allow local storage [here](https://github.com/k8ssandra/k8ssandra-operator/blob/standalone-medusa-pod/pkg/medusa/reconcile.go#L281)? 

Maybe Medusa just has a lot of configurables, but I wonder if better API design on the CRD (so that it uses more k8s native types) together with more extensive usage of goalesce might allow for better configurability while simplifying the reconciliation code paths?

pkg/reconciliation/generic.go
```diff
@@ -19,11 +19,15 @@ type Reconcileable[T any] interface {
 }
 
 // Try with U, a type of any whose POINTER still fulfils Reoncilable...
-func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U) result.ReconcileResult {
+func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U, labels map[string]string) result.ReconcileResult {
```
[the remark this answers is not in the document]
adejanovski: fair enough, I thought it was a nice improvement but I'll switch it back.

CHANGELOG/CHANGELOG-1.7.md
```diff
@@ -15,6 +15,9 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+* [CHANGE] [#483](https://github.com/k8ssandra/k8ssandra-operator/issues/483) Deploy a standalone Medusa pod for operator to Medusa direct interactions
```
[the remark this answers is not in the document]
adejanovski: This PR contains both the PRs you and I worked on last year actually, plus the things that were missing to actually use this mapping.

controllers/control/k8ssandratask_controller_test.go
```diff
@@ -69,7 +69,7 @@ func TestK8ssandraTask(t *testing.T) {
 	defer cancel()
 
 	t.Run("ExecuteParallelK8ssandraTask", testEnv.ControllerTest(ctx, executeParallelK8ssandraTask))
-	t.Run("ExecuteSequentialK8ssandraTask", testEnv.ControllerTest(ctx, executeSequentialK8ssandraTask))
+	//t.Run("ExecuteSequentialK8ssandraTask", testEnv.ControllerTest(ctx, executeSequentialK8ssandraTask))
```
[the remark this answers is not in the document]
adejanovski: Forgot about it in the description indeed.
This test is failing a lot, even locally, and I've created a refactoring ticket to fix it: https://github.com/k8ssandra/k8ssandra-operator/issues/981

controllers/k8ssandra/cleanup.go
```diff
@@ -357,3 +366,69 @@ func (r *K8ssandraClusterReconciler) deleteK8ssandraConfigMaps(
...
+					"Context", dcTemplate.K8sContext)
+				hasErrors = true
+			}
+		}
+	}
+
+	return
+}
+
+func (r *K8ssandraClusterReconciler) deleteDeployments(
```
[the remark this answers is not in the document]
adejanovski: Sure, I've created [a refactoring ticket](https://github.com/k8ssandra/k8ssandra-operator/issues/992) for this.

controllers/k8ssandra/cleanup.go
```diff
@@ -357,3 +366,69 @@ func (r *K8ssandraClusterReconciler) deleteK8ssandraConfigMaps(
 	}
 	return
 }
+
+func (r *K8ssandraClusterReconciler) deleteServices(
```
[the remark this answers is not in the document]
adejanovski: Covered in [the refactoring ticket](https://github.com/k8ssandra/k8ssandra-operator/issues/992).

controllers/k8ssandra/medusa_reconciler.go
adejanovski: I think we should get rid of the additionalVolumes now, because the testing strategy can't rely anymore on local volumes anyway.
It's the only reason why we were using them I think.
One thing I'm worried about is that it would break updates if someone is using the volumes 🤔 cass-operator won't allow changing the list of additional volumes on an existing cluster.

pkg/reconciliation/generic.go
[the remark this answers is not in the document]
Miles-Garnsey: If you wanted to, for the purposes of this repo, you could wrap it in ReconcileObjectWithLabels or similar. But honestly my preference is just to see the labels added by the caller as a step before calling this function. I don't love super-deep function call stacks.

CHANGELOG/CHANGELOG-1.7.md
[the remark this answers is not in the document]
Miles-Garnsey: OK, it'll just take a bit longer to review, but that's OK.

controllers/control/k8ssandratask_controller_test.go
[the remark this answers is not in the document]
Miles-Garnsey: If you have a refactoring ticket already then I'm fine with this.

controllers/k8ssandra/cleanup.go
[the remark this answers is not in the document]
Miles-Garnsey: Dude, I'm not letting you get away with unit tests in a refactoring ticket 😄 

I'd like the parameter list and return types fixed here too, unless there's some reason why those changes are more complex than I'd expect?

controllers/k8ssandra/medusa_reconciler.go
Miles-Garnsey: One for us to discuss later this evening?

controllers/k8ssandra/cleanup.go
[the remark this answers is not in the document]
adejanovski: > 'd like the parameter list and return types fixed here too, unless there's some reason why those changes are more complex than I'd expect?

It's just that it's out of scope I think. I've just copied the method that exists for configmaps and modified them to work for services and deployments.
As for testing, we do have envtests for this [here](https://github.com/k8ssandra/k8ssandra-operator/pull/980/files#diff-b6c97c9724c3403b1304d1d82e10b8e76299fbb470b629ee6fdf9e26513c8be6R309-R312).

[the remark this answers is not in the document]
adejanovski: Let me see if I can come up with some unit tests though.

controllers/k8ssandra/cleanup.go
```diff
@@ -357,3 +366,69 @@ func (r *K8ssandraClusterReconciler) deleteK8ssandraConfigMaps(
...
+	namespace string,
+	remoteClient client.Client,
+	kcLogger logr.Logger,
+) (hasErrors bool) {
+	selector := k8ssandralabels.CleanedUpByLabels(client.ObjectKey{Namespace: kc.Namespace, Name: kc.Name})
+	options := client.ListOptions{
+		Namespace:     namespace,
+		LabelSelector: labels.SelectorFromSet(selector),
+	}
+	serviceList := &corev1.ServiceList{}
```
Miles-Garnsey, line 383: Suggestion: A better option here would probably be `DeleteAllOf`, using the labels in the LabelSelector. I am happy for you to add this to the refactoring ticket you already have open to fix these functions up and combine them into a single generic function.

controllers/k8ssandra/cleanup_test.go
```diff
@@ -0,0 +1,102 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestK8ssandraClusterReconciler_DeleteServices(t *testing.T) {
```
Miles-Garnsey, line 19: These tests are nicely structured. Good to see them in here as unit tests rather than standing up the whole envtest framework.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2625,7 +2621,7 @@ func injectContainersAndVolumes(t *testing.T, ctx context.Context, f *framework.
 	verifyReplicatedSecretReconciled(ctx, t, f, kc)
 
 	verifySystemReplicationAnnotationSet(ctx, t, f, kc)
-
+	reconcileMedusaStandaloneDeployment(ctx, t, f, kc, "dc1", f.DataPlaneContexts[0])
```
Miles-Garnsey, line 2624: You're running the reconciliation here, but I'm not sure whether you ever assert that the deployment was created?

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -37,16 +39,34 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
...
+		}
 		medusa.UpdateMedusaInitContainer(dcConfig, medusaSpec, kc.Spec.UseExternalSecrets(), kc.SanitizedName(), logger)
-		medusa.UpdateMedusaMainContainer(dcConfig, medusaSpec, kc.Spec.UseExternalSecrets(), kc.SanitizedName(), logger)
-		medusa.UpdateMedusaVolumes(dcConfig, medusaSpec, kc.SanitizedName())
+		medusa.UpdateMedusaMainContainer(dcConfig, medusaContainer)
+
+		// Create required volumes for the Medusa containers
+		volumes := medusa.GenerateMedusaVolumes(dcConfig, medusaSpec, kc.SanitizedName())
+		for _, volume := range volumes {
+			cassandra.AddOrUpdateVolume(dcConfig, volume.Volume, volume.VolumeIndex, volume.Exists)
```
Miles-Garnsey, line 59: Suggestion: probably for later, but I think this function can probably be subsumed by logic from Goalesce going forward.

pkg/medusa/reconcile.go
```diff
@@ -427,6 +432,125 @@ func UpdateMedusaVolumes(dcConfig *cassandra.DatacenterConfig, medusaSpec *api.M
...
+func MedusaServiceName(clusterName string, dcName string) string {
+	return fmt.Sprintf("%s-%s-medusa-service", clusterName, dcName)
+}
+
+func MedusaStandaloneDeploymentName(clusterName string, dcName string) string {
+	return fmt.Sprintf("%s-%s-medusa-standalone", clusterName, dcName)
+}
+
+func StandaloneMedusaDeployment(medusaContainer *corev1.Container, clusterName, dcName, namespace string, logger logr.Logger) *appsv1.Deployment {
+	medusaStandaloneContainer := *medusaContainer.DeepCopy()
```
Miles-Garnsey, line 453: Issue: if the first thing we do is take a DeepCopy we should probably just be passing a copy, not a pointer.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -37,16 +39,34 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
...
+		}
+
+		// Create the Medusa standalone pod
+		desiredMedusaStandalone := medusa.StandaloneMedusaDeployment(medusaContainer, kc.Name, dcConfig.Meta.Name, namespace, logger)
+
+		// Add the volumes previously computed to the Medusa standalone pod
+		for _, volume := range volumes {
+			cassandra.AddOrUpdateVolumeToSpec(&desiredMedusaStandalone.Spec.Template, volume.Volume, volume.VolumeIndex, volume.Exists)
```
Miles-Garnsey, line 67: Issue (might have to be a suggestion): At least for this one (given that it isn't a Cassandra pod) can we please use Coalesce? 

This is tough to interpret for anyone who comes after us, because it is using logic originally developed for manipulating Cassandra pod's podTemplateSpecs in a totally different context. 

It also requires an additional for loop which goalesce would abstract away. 

I don't think I can actually block this PR on those grounds, but I do feel somewhat strongly about this one. Perhaps it needs to be a more wholistic review of the codebase with a view to removing this function completely and replacing it with Goalesce throughout all projects.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -56,13 +76,57 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 			}
 			secret.AddInjectionAnnotationMedusaContainers(&dcConfig.Meta.Pods, cassandraUserSecretName)
 		}
+
+		// Reconcile the Medusa standalone deployment
+		kcKey := utils.GetKey(kc)
+		desiredMedusaStandalone.SetLabels(labels.CleanedUpByLabels(kcKey))
```
Miles-Garnsey, line 82: RE CleanedUpByLabels, would the medusa deployment not always be in the same namespace as the K8ssandraCluster and therefore be cleaned up using standard k8s GC and OwnerReferences? 

Regardless, I understand that PartOf labels ought to be applied to identify the cluster, cluster-home-namespace that the deployment belongs to. However, I am confused as to why we're using labels based cleanup in this instance.

Additionally, I also wonder whether this feedback may also apply to the services and deployments cleanup code too. (Sorry, this only just occured to me :/ )

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -56,13 +76,57 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
...
 
+// Check if the Medusa standalone deployment is ready
+func (r *K8ssandraClusterReconciler) isMedusaStandaloneReady(ctx context.Context, remoteClient client.Client, desiredMedusaStandalone *appsv1.Deployment) (bool, error) {
+	// Get the medusa standalone deployment and check the number of ready replicas
+	deplKey := utils.GetKey(desiredMedusaStandalone)
+	medusaStandalone := &appsv1.Deployment{}
+	if err := remoteClient.Get(context.Background(), deplKey, medusaStandalone); err != nil {
+		return false, err
+	}
+	return medusaStandalone.Status.ReadyReplicas > 0, nil
```
Miles-Garnsey, line 127: Issue: I think this should more likely be built around DeploymentCondition: Available == true instead of ReadyReplicas.

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -80,6 +85,30 @@ func createMultiDcClusterWithMedusa(t *testing.T, ctx context.Context, f *framew
 				CassandraUserSecretRef: corev1.LocalObjectReference{
 					Name: cassandraUserSecret,
 				},
+				ReadinessProbe: &corev1.Probe{
```
Miles-Garnsey, line 88: Question: why are we fully specifying so much config here? What is the failure we're trying to catch?

Miles-Garnsey: From my experience, if you're worried about the way this stuff makes its way from the CR to the output resources, you're best off using an e2e test, since the API server can play tricks on you that won't be caught in an envtest.

controllers/k8ssandra/medusa_reconciler_test.go
```diff
@@ -89,12 +118,35 @@ func createMultiDcClusterWithMedusa(t *testing.T, ctx context.Context, f *framew
...
+	medusaDeploymentKey1 := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: medusa.MedusaStandaloneDeploymentName("test", "dc1")}, K8sContext: f.DataPlaneContexts[0]}
+	medusaDeployment1 := &appsv1.Deployment{}
+	require.Eventually(func() bool {
+		if err := f.Get(ctx, medusaDeploymentKey1, medusaDeployment1); err != nil {
+			return false
+		}
+		return true
+	}, timeout, interval)
+
+	require.True(f.ContainerHasEnvVar(medusaDeployment1.Spec.Template.Spec.Containers[0], "MEDUSA_RESOLVE_IP_ADDRESSES", "False"))
```
Miles-Garnsey, line 138: Question: what does this do?

Question2: can we test that the environment variable containing the mappings has been created?

controllers/k8ssandra/vector.go
```diff
@@ -40,6 +41,7 @@ func (r *K8ssandraClusterReconciler) reconcileVector(
 		k8ssandralabels.SetWatchedByK8ssandraCluster(desiredVectorConfigMap, kcKey)
 
 		// Check if the vector config map already exists
+		desiredVectorConfigMap.SetLabels(labels.CleanedUpByLabels(kcKey))
```
Miles-Garnsey, line 44: This doesn't seem related to this PR? Is this right?

controllers/medusa/medusabackupjob_controller_test.go
```diff
@@ -405,3 +408,44 @@ func verifyObjectDoesNotExist(ctx context.Context, t *testing.T, f *framework.Fr
...
+	t.Log("check ReplicatedSecret reconciled")
+
+	medusaDepl := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      medusa.MedusaStandaloneDeploymentName(kc.SanitizedName(), dcName),
+			Namespace: kc.Namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": medusa.MedusaStandaloneDeploymentName(kc.SanitizedName(), dcName)},
```
Miles-Garnsey, line 422: I'm a little confused about this one, shouldn't the kc.SanitizedName just be kc.meta.name?

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -244,44 +238,27 @@ func (r *MedusaRestoreJobReconciler) podTemplateSpecUpdateComplete(ctx context.C
...
-		}
-		if len(prepare.Status.InProgress) == 0 {
-			// No more pods are running the task but finish time is not set.
-			// This means the task failed.
-			return false, fmt.Errorf("prepare restore task failed for restore %s", request.RestoreJob.Name)
-		}
+// prepareRestore prepares the MedusaRestoreMapping for the restore operation.
+// It uses the Medusa client to get the host map for the restore operation, using the Medusa standalone service to get the backup topology.
+func (r *MedusaRestoreJobReconciler) prepareRestore(ctx context.Context, request *medusa.RestoreRequest) (*medusav1alpha1.MedusaRestoreMapping, error) {
+	medusaClient, err := r.ClientFactory.NewClient(medusaServiceUrl(
```
Miles-Garnsey, line 244: Nice seam.

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -244,44 +238,27 @@ func (r *MedusaRestoreJobReconciler) podTemplateSpecUpdateComplete(ctx context.C
...
+		request.Datacenter.Spec.ClusterName,
```
Miles-Garnsey, line 245: Observation: We really need to start distinguishing between the ClusterName for the k8s resource and the Cassandra cluster name. Impossible to tell what is going on throughout these codebases sometimes :/

controllers/k8ssandra/medusa_reconciler_test.go
Miles-Garnsey: Ignore q2, I found you're testing this in the restore job controller test.

test/framework/e2e_framework.go
```diff
@@ -348,6 +349,73 @@ func (f *E2eFramework) CreateCassandraEncryptionStoresSecret(namespace string) e
...
+func (f *E2eFramework) InstallMinio() error {
+	path := filepath.Join("..", "testdata", "fixtures", "minio.yaml")
+	for _, k8sContext := range f.DataPlaneContexts {
+		options := kubectl.Options{Namespace: MinioNamespace, Context: k8sContext}
+		f.logger.Info("Install Minio", "Namespace", MinioNamespace, "Context", k8sContext)
+		if err := kubectl.Apply(options, path); err != nil {
+			return err
+		}
+
+		// Wait for the minio-operator rollout to complete
```
Miles-Garnsey, line 361: Nit: I don't think we're using Minio operator now?

test/e2e/medusa_test.go
```diff
@@ -164,3 +169,28 @@ func verifyRestoreJobFinished(t *testing.T, ctx context.Context, f *framework.E2
 		return !restore.Status.FinishTime.IsZero()
 	}, polling.medusaRestoreDone.timeout, polling.medusaRestoreDone.interval, "restore didn't finish within timeout")
 }
+
+func checkMedusaStandaloneDeploymentExists(t *testing.T, ctx context.Context, dcKey framework.ClusterKey, f *framework.E2eFramework, kc *api.K8ssandraCluster) {
```
Miles-Garnsey, line 173: Question:I'm not very good at reading these e2e tests, so maybe I'm missing something. But is there an e2e test here which actually exercises that a cluster can be backed up, destroyed, and then restored into another cluster which has a different name or different rack/DC names?

test/testdata/fixtures/multi-dc-medusa/k8ssandra.yaml
```diff
@@ -33,11 +33,11 @@ spec:
...
       bucketName: k8ssandra-medusa
       prefix: test
-      podStorage:
-        storageClassName: standard
-        accessModes:
-          - ReadWriteOnce
-        size: 100Mi
+      storageSecretRef:
+        name: medusa-bucket-key
+      host: test-hl.minio.svc.cluster.local
```
Miles-Garnsey, line 41: Where does this service come from? The minio namespace already contains a minio-service which is created alongside the deployment, which might be better to use here to avoid confusion.

test/e2e/medusa_test.go
Miles-Garnsey: More generally, we are very deficient on tests here. I can't find any test fixtures that reference the `medusabackupjobs.medusa.k8ssandra.io`, `medusabackupschedules.medusa.k8ssandra.io`, `medusatasks.medusa.k8ssandra.io`, `medusabackups.medusa.k8ssandra.io`, `medusarestorejobs.medusa.k8ssandra.io` etc. from a brief inspection.

pkg/k8ssandra/util.go
```diff
@@ -29,3 +35,27 @@ func GetDatacenterForDecommission(kc *api.K8ssandraCluster) string {
 
 	return ""
 }
+
+func GetK8ssandraClusterFromCassDc(ctx context.Context, cassdc *cassdcapi.CassandraDatacenter, client client.Client, logger logr.Logger) *api.K8ssandraCluster {
```
Miles-Garnsey, line 39: Issue: this is wrong, and is causing the bug I found in testing.

Miles-Garnsey:
I've left a number of comments above RE the implementation details. Black box testing reveals a bug.

**Procedure**

1. I created a cluster with medusa options specified and three racks [rack1,rack2,rack3], clusterName: "First Cluster", DC name firstdc, K8ssandraCluster k8s resource name `test`. I also created the minio deployment, and minio setup job. 
2. I created a few dummy tables/keyspaces. 
3. I ran nodetool flush on one node.
4. I created a backup against that cluster which ran successfully (based on the presence of a finished time).
5. I deleted that cluster and created a new one with the same k8s resource name but `.spec.cassandra.clusterName: "Second Cluster"`.

I got the following error back from the k8ssandra-operator pod:

```
2023-06-19T06:14:45.508Z    ERROR Reconciler error {"controller": "medusarestorejob", "controllerGroup": "medusa.k8ssandra.io", "controllerKind": "MedusaRestoreJob", "MedusaRestoreJob": {"name":"restore-backup1","namespace":"k8ssandra-operator"}, "namespace": "k8ssandra-operator", "name": "restore-backup1", "reconcileID": "e313db82-43e8-4791-834c-6bf1b8aaff90", "error": "failed to create gRPC connection to Second Cluster-seconddc-medusa-service.k8ssandra-operator.svc:50051: context deadline exceeded"}
```

Somewhere (I think perhaps [here](https://github.com/k8ssandra/k8ssandra-operator/pull/980/files#diff-225b586e944a1fbbc3e985b9662ca0196267b898ba3b7655c85e86fdd835dc3bR39)) in the codebase you're using the `clusterName` field when you should be using the k8s resource name instead.

controllers/k8ssandra/medusa_reconciler.go
Miles-Garnsey: Where did we land on this? Are we just going to add release notes with a migration pathway?

pkg/medusa/reconcile.go
Miles-Garnsey: This appears to be OK now.

pkg/k8ssandra/util.go
adejanovski: Very good catch!!
We need to use the sanitized names instead for both the cluster and the dc, otherwise we'll end up with either the wrong name or a name that's incompatible with kubernetes.
I'll fix this.

controllers/k8ssandra/k8ssandracluster_controller_test.go
Miles-Garnsey: Ah, so the intention here is just to advance the reconciliation, rather than to test anything to do with Medusa? Maybe a comment is best to explain that then?

controllers/k8ssandra/medusa_reconciler_test.go
Miles-Garnsey: I think an envtest is OK in this case because you aren't doing anything that really relies on the API server (e.g. defaults, etc.) but given that you aren't relying much on the API server, I think you could simplify the test by not specifying a lot of these settings.

controllers/k8ssandra/medusa_reconciler_test.go
Miles-Garnsey: So it turns off reverse DNS resolution by the sound of it. OK.

controllers/medusa/medusabackupjob_controller_test.go
Miles-Garnsey: My understanding was that the SanitizedName was for getting a DNS-compatible k8s resource name from a Cassandra cluster name (which can contain more characters). But we should be naming dependant objects based on the k8s resource names, shouldn't we?

controllers/medusa/medusarestorejob_controller.go
Miles-Garnsey: > The only way to get the cluster name, which is used to build the service url, is .spec.ClusterName here.

Do you mean .spec.cassandra.clusterName? This can be set differently to the k8s resource name and doesn't need to bear any relation to it AFAIK.

I believe you could possibly use .metadata.labels to get the parent cluster name.

pkg/k8ssandra/util.go
Miles-Garnsey: See my comments above, I suspect we need to discuss the use of sanitizedName and similar in this PR. I think what you need to use is actually the cluster name from the labels throughout.

controllers/k8ssandra/k8ssandracluster_controller_test.go
adejanovski: I'll add a comment, but this matches the things we've been doing so far in envtests.

controllers/medusa/medusabackupjob_controller_test.go
adejanovski: As discussed offline, this is not the path we took with cluster/dc name overrides.
The SanitizedName will pick the override if it exists and sanitize it, or the resource name if not.
And all dependent objects will be named based on this.

controllers/medusa/medusarestorejob_controller.go
adejanovski: Conversation resolved offline: we don't want to use the k8c resource name here, we really need the sanitized name.

Miles-Garnsey:
Oops, looks like I approved the old PR. My comments were as follows:

```
@adejanovski, as discussed, it does appear that the sanitized clusterName field is the point of reference for downstream resource names, so that's all good. Apologies for my error there.

I've raised a ticket RE using goalesce to disentangle things a bit here, but as it is an implementation detail I won't block merge on this point.

I tested once again, changing the k8s resource name, cluster name, and rack names (but leaving the DC name unchanged).

That tests shows that I can use cqlsh to extract data added to the original cluster from the restored cluster. So I think this is good to merge. Approving once again. Nice work on getting this monster across the line!
```

Approving this one now.

Miles-Garnsey (approved):

PR #998, opened 2023-06-07
Publish images even when only documentation changes are commited

### LINKED ISSUE #985: CI/CD does not produce images for some commits
**What happened?**

CI/CD runs for documentation commits don't produce images. This leads to errors when you try to reference the k8ssandra-operator kustomizations from downstream kustomizations.

All commits on `main` must create an image to avoid kustomizations being broken.

**Definition of done**

All commits (inc. docs commits/whatever) generate an image and kustomizations that can be deployed.

emerkle826:
**What this PR does**:
Removes the `paths-ignore` for commits that hit main or release branches so that documentation-only commits will still trigger an image build.
Fixes #985 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/test_and_build_image.yaml
+++ b/.github/workflows/test_and_build_image.yaml
@@ -4,9 +4,6 @@ on:
     branches:
       - main
       - 'release/**'
-    paths-ignore:
-      - 'docs/**'
-      - 'CHANGELOG/**'
   pull_request:
     branches:
       - main


Miles-Garnsey:
I'll pick this up for review as I raised the original issue.

Miles-Garnsey (approved):
A note that we normally raise branches on our own fork off this repo, rather than from origin. It doesn't matter too much in this case.

Otherwise, looking at this we aren't able to manually verify that it works as intended without merging to main because of the fact that this change affects only merges to `main`:

```
branches:
      - main
      - 'release/**'
```

It LGTM so I suggest we merge this and verify it works in a second PR updating the changelog. While that should normally be done as part of this PR, in this case splitting it out allows us to verify that that change has worked without merging unnecessary things to main (although it will make our commit history less clear).

@emerkle826 are you OK with that approach? Please merge this if so.

emerkle826:
> I suggest we merge this and verify it works in a second PR updating the changelog. While that should normally be done as part of this PR, in this case splitting it out allows us to verify that that change has worked without merging unnecessary things to main (although it will make our commit history less clear).
> 
> @emerkle826 are you OK with that approach? Please merge this if so.

Sounds like a plan. I can merge this and then I'll PR a changelog entry.

PR #1021, opened 2023-07-12
Fix race condition in K8ssandraTask status update (fixes #981)

### LINKED ISSUE #981: K8ssandraTask envtests are failing often
We're seeing a lot of flakes due to the K8ssandraTask envtests.
This is becoming a problem as it requires multiple runs before we can be sure a PR is ready to be merged.
Let's take the time to fix it.

```[tasklist]
### Definition of Done
- [x] The sources of flakiness in the K8ssandraTask envtests are identified
- [x] Such tests don't fail on 10 consecutive runs
```

olim7t:
**What this PR does**:
The error is happening in `TestK8ssandraTask/ExecuteSequentialK8ssandraTask`, about 10% of the time , only when the whole `TestK8ssandraTask` suite runs. Here's a script to reproduce it:
```shell
#!/bin/bash
export KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.25.0-darwin-amd64"

RUNS=10
FAILURES=0

for i in $( seq 1 $RUNS )
do
  echo "Run $i/$RUNS ($FAILURES failures so far)"
  go test ./controllers/control/... -v -count=1
  if [ $? -ne 0 ]; then
    FAILURES=$((FAILURES+1))
  fi
done

echo "Done ($FAILURES failures)"
```
The test creates a `K8ssandraTask`, which spawns two `CassandraTask`s that both succeed. So the status should end up as `Active=0, Succeeded=2`, but what the test observes is `Active=1, Succeeded=2`.

Debug logs showed the correct counts in [RefreshGlobalStatus](https://github.com/k8ssandra/k8ssandra-operator/blob/main/apis/control/v1alpha1/k8ssandratask_types.go#L101), but not after we [patch the status](https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/control/k8ssandratask_controller.go#L186) in the controller. I suspect there were occasionally conflicting patch operations, and `Active=0` being the zero value of the field was ignored (possibly compounded by the fact that the field is optional).

I switched to an Update operation to persist the status.

Fixes #981

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.9.md
+++ b/CHANGELOG/CHANGELOG-1.9.md
@@ -14,3 +14,5 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
 ## unreleased
+
+* [BUGFIX] [#981](https://github.com/k8ssandra/k8ssandra-operator/issues/981) Fix race condition in K8ssandraTask status update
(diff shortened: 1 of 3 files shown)


adejanovski (approved):
Looks good, nice finding!

PR #1022, opened 2023-07-14
Use controller references to clean up DC components

### LINKED ISSUE #961: Refactor cleanup

### LINKED ISSUE #992: Refactor K8ssandraCluster deletion cleanup operation
In controllers/k8ssandra/cleanup.go we delete all the objects that are tied to a deleted K8ssandraCluster object.
The `deleteDeployments()`, `deleteServices()` and `deleteConfigMaps()` could benefit from the following improvements:

- Combine them into a generic function/method if possible
- Move a few function arguments to the K8ssandraClusterReconciler object in order to reduce their number in the calls
- Evaluate the possibility of returning a more meaningful object than a boolean, which could carry more information (such as an error or a ReconcileResult)
- Write unit tests for it

olim7t:
**What this PR does**:

Fixes #961
Fixes #992

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/cleanup.go
+++ b/controllers/k8ssandra/cleanup.go
@@ -9,7 +9,6 @@ import (
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	k8ssandraapi "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	reaperapi "github.com/k8ssandra/k8ssandra-operator/apis/reaper/v1alpha1"
-	stargateapi "github.com/k8ssandra/k8ssandra-operator/apis/stargate/v1alpha1"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/annotations"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/k8ssandra"
 	k8ssandralabels "github.com/k8ssandra/k8ssandra-operator/pkg/labels"
@@ -38,7 +37,6 @@ func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.
 
 	logger.Info("Starting deletion")
 
-	kcKey := utils.GetKey(kc)
 	hasErrors := false
 
 	for _, dcTemplate := range kc.Spec.Cassandra.Datacenters {
@@ -64,31 +62,6 @@ func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.
 		} else if err = remoteClient.Delete(ctx, dc); err != nil {
 			logger.Error(err, "Failed to delete CassandraDatacenter", "CassandraDatacenter", dcKey, "Context", dcTemplate.K8sContext)
 			hasErrors = true
-		}
-
-		selector := k8ssandralabels.CleanedUpByLabels(kcKey)
-		stargateList := &stargateapi.StargateList{}
-		options := client.ListOptions{
-			Namespace:     namespace,
-			LabelSelector: labels.SelectorFromSet(selector),
-		}
-
-		err = remoteClient.List(ctx, stargateList, &options)
-		if err != nil {
-			logger.Error(err, "Failed to list Stargate objects", "Context", dcTemplate.K8sContext)
-			hasErrors = true
-			continue
-		}
-
-		for _, sg := range stargateList.Items {
-			if err = remoteClient.Delete(ctx, &sg); err != nil {
-				key := client.ObjectKey{Namespace: namespace, Name: sg.Name}
-				if !errors.IsNotFound(err) {
-					logger.Error(err, "Failed to delete Stargate", "Stargate", key,
-						"Context", dcTemplate.K8sContext)
-					hasErrors = true
-				}
-			}
 		}
 
 		if r.deleteReapers(ctx, kc, dcTemplate, namespace, remoteClient, logger) {
@@ -164,19 +137,7 @@ func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *ap
 func (r *K8ssandraClusterReconciler) deleteDc(ctx context.Context, kc *api.K8ssandraCluster, dcName string, logger logr.Logger) result.ReconcileResult {
 	kcKey := utils.GetKey(kc)
 
-	stargate, remoteClient, err := r.findStargateForDeletion(ctx, kcKey, dcName, nil)
-	if err != nil {
-		return result.Error(err)
-	}
-
-	if stargate != nil {
-		if err = remoteClient.Delete(ctx, stargate); err != nil && !errors.IsNotFound(err) {
-			return result.Error(fmt.Errorf("failed to delete Stargate for dc (%s): %v", dcName, err))
-		}
-		logger.Info("Deleted Stargate", "Stargate", utils.GetKey(stargate))
-	}
-
-	reaper, remoteClient, err := r.findReaperForDeletion(ctx, kcKey, dcName, remoteClient)
+	reaper, remoteClient, err := r.findReaperForDeletion(ctx, kcKey, dcName, nil)
 	if err != nil {
 		return result.Error(err)
 	}
@@ -219,45 +180,6 @@ func (r *K8ssandraClusterReconciler) deleteDc(ctx context.Context, kc *api.K8ssa
 	delete(kc.Status.Datacenters, dcName)
 	logger.Info("DC deletion finished", "DC", dcName)
 	return result.Continue()
-}
-
-func (r *K8ssandraClusterReconciler) findStargateForDeletion(
-	ctx context.Context,
-	kcKey client.ObjectKey,
-	dcName string,
-	remoteClient client.Client) (*stargateapi.Stargate, client.Client, error) {
-
-	selector := k8ssandralabels.CleanedUpByLabels(kcKey)
-	options := &client.ListOptions{LabelSelector: labels.SelectorFromSet(selector)}
-	stargateList := &stargateapi.StargateList{}
-	stargateName := kcKey.Name + "-" + dcName + "-stargate"
-
-	if remoteClient == nil {
-		for _, remoteClient := range r.ClientCache.GetAllClients() {
-			err := remoteClient.List(ctx, stargateList, options)
-			if err != nil {
-				return nil, nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
-			}
-			for _, stargate := range stargateList.Items {
-				if stargate.Name == stargateName {
-					return &stargate, remoteClient, nil
-				}
-			}
-		}
-	} else {
-		err := remoteClient.List(ctx, stargateList, options)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to find Stargate (%s) for DC (%s) deletion: %v", stargateName, dcName, err)
-		}
-
-		for _, stargate := range stargateList.Items {
-			if stargate.Name == stargateName {
-				return &stargate, remoteClient, nil
-			}
-		}
-	}
-
-	return nil, nil, nil
 }
 
 func (r *K8ssandraClusterReconciler) findReaperForDeletion(
--- a/controllers/k8ssandra/cleanup.go
+++ b/controllers/k8ssandra/cleanup.go
@@ -8,7 +8,6 @@ import (
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	k8ssandraapi "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
-	reaperapi "github.com/k8ssandra/k8ssandra-operator/apis/reaper/v1alpha1"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/annotations"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/k8ssandra"
 	k8ssandralabels "github.com/k8ssandra/k8ssandra-operator/pkg/labels"
@@ -61,10 +60,6 @@ func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.
 			}
 		} else if err = remoteClient.Delete(ctx, dc); err != nil {
 			logger.Error(err, "Failed to delete CassandraDatacenter", "CassandraDatacenter", dcKey, "Context", dcTemplate.K8sContext)
-			hasErrors = true
-		}
-
-		if r.deleteReapers(ctx, kc, dcTemplate, namespace, remoteClient, logger) {
 			hasErrors = true
 		}
 
@@ -137,19 +132,7 @@ func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *ap
 func (r *K8ssandraClusterReconciler) deleteDc(ctx context.Context, kc *api.K8ssandraCluster, dcName string, logger logr.Logger) result.ReconcileResult {
 	kcKey := utils.GetKey(kc)
 
-	reaper, remoteClient, err := r.findReaperForDeletion(ctx, kcKey, dcName, nil)
-	if err != nil {
-		return result.Error(err)
-	}
-
-	if reaper != nil {
-		if err = remoteClient.Delete(ctx, reaper); err != nil && !errors.IsNotFound(err) {
-			return result.Error(fmt.Errorf("failed to delete Reaper for dc (%s): %v", dcName, err))
-		}
-		logger.Info("Deleted Reaper", "Reaper", utils.GetKey(reaper))
-	}
-
-	dc, remoteClient, err := r.findDcForDeletion(ctx, kcKey, dcName, remoteClient)
+	dc, remoteClient, err := r.findDcForDeletion(ctx, kcKey, dcName)
 	if err != nil {
 		return result.Error(err)
 	}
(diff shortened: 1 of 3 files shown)


controllers/k8ssandra/cleanup.go
```diff
@@ -66,51 +60,6 @@ func (r *K8ssandraClusterReconciler) checkDeletion(ctx context.Context, kc *api.
...
-			hasErrors = true
-		}
-
-		if r.deleteDeployments(ctx, kc, dcTemplate, namespace, remoteClient, logger) {
-			hasErrors = true
-		}
-
-		if r.deleteServices(ctx, kc, dcTemplate, namespace, remoteClient, logger) {
-			hasErrors = true
-		}
```
olim7t, line 105: The deployments and services cleanup was added for #980, but we changed the way Medusa is deployed since then, so this was already stale code.

adejanovski:
@olim7t , is this ready for review?

olim7t:
> is this ready for review?

It is now. I've tested manually to confirm that all impacted resources are correctly cleaned up:
- `Stargate`
- `Reaper`
- Medusa: ConfigMap and CronJob
- `-metrics-agent` ConfigMap
- `-per-node-config` ConfigMap
- `-cass-vector` ConfigMap

I'm still seeing some secrets left behind after I delete a DC, for example `<cluster-name>-medusa`, `<cluster-name>-reaper` and `<cluster-name>-reaper-ui`.
But those are cluster-level secrets that get replicated, not DC-level. So they weren't cleaned up by the code that I removed. I think we should assign them an owner reference to the `K8ssandraCluster` (assuming the replicated-secret controller cleans up the replicated copies when they get deleted).

vincentcreusot (approved):
Pretty good
Been able to run unit tests, all pass, and test removing the `CassandraDatacenter` which removes the `Stargate` resource correctly.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -346,3 +347,34 @@ func (r *K8ssandraClusterReconciler) getOperatorNamespace() string {
 	}
 	return operatorNamespace
 }
+
+// setupMedusaCleanup adds owner references to ensure that the remote resources created by reconcileMedusa are correctly
+// cleaned up when the CassandraDatacenter is deleted. We do that in a second pass because the CassandraDatacenter did
+// not exist yet at the time those resources were created.
+func (r *K8ssandraClusterReconciler) setupMedusaCleanup(
```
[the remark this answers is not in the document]
Miles-Garnsey: How much effort would it be to ensure that Reaper and Medusa are created only after the CassDC? That should already be the case I thought, but maybe I'm wrong here.

If it isn't the case, there are workarounds for this that aren't too arduous, but they require a new type (like CassandraDatacenterRequest) to exist, which take ownership until the actual cassDC is created.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -346,3 +347,34 @@ func (r *K8ssandraClusterReconciler) getOperatorNamespace() string {
...
+	// namespace, they would share the same ConfigMap, so setting one of them as the owner is wrong.
+	// However, this is an unlikely scenario (DCs are usually in different contexts, let alone namespaces). Also, if we
+	// delete one DC and the other still needs the ConfigMap, it will be recreated.
+	configMapKey := client.ObjectKey{
+		Namespace: dc.Namespace,
+		// see pgk/medusa/reconcile.go
+		Name: fmt.Sprintf("%s-medusa", kc.SanitizedName()),
+	}
+	result := setDcOwnership(ctx, dc, configMapKey, &corev1.ConfigMap{}, remoteClient, r.Scheme, logger)
+	if result.Completed() {
```
Miles-Garnsey, line 371: Issue: This doesn't seem like the right use of `Completed()`. Actually, I try not to use `Completed()` at all - it isn't clear what it means because the `callBackSoon` implementation of `ReconcileResult` also returns `Completed() == true`. 

I would use one of these:

```
	IsError() bool
	IsRequeue() bool
	IsDone() bool
```

If you have an `IsError()` or `IsRequeue()` then you should return out to bail on the reconcile loop and allow the controller to requeue it. Conversely, actually if it is `IsDone()` then you can continue the reconciliation. 

Right now, I'm pretty sure you'll end up bailing early on requeues which isn't what you want.

controllers/k8ssandra/per_node_config.go
```diff
@@ -172,3 +172,20 @@ func (r *K8ssandraClusterReconciler) reconcileUserProvidedPerNodeConfiguration(
 
 	return result.Continue()
 }
+
+// setupMedusaCleanup adds an owner reference to ensure that the remote ConfigMap created by
```
Miles-Garnsey, line 176: Nitpick: comment doesn't match the function name.

controllers/k8ssandra/reaper.go
```diff
@@ -127,7 +127,10 @@ func (r *K8ssandraClusterReconciler) reconcileReaper(
 		if err := remoteClient.Get(ctx, reaperKey, actualReaper); err != nil {
 			if errors.IsNotFound(err) {
 				logger.Info("Creating Reaper resource")
-				if err := remoteClient.Create(ctx, desiredReaper); err != nil {
+				if err := controllerutil.SetControllerReference(actualDc, desiredReaper, r.Scheme); err != nil {
```
Miles-Garnsey, line 130: Question: why do we do this inside reconcileReaperSchema? Seems like an odd place to reconcile k8s resources.

controllers/k8ssandra/reaper_test.go
```diff
@@ -254,7 +254,10 @@ func createMultiDcClusterWithControlPlaneReaper(t *testing.T, ctx context.Contex
 			Name:      "reaper"},
 	}
 	t.Log("check that control plane reaper is created")
-	require.Eventually(f.ReaperExists(ctx, cpReaperKey), timeout, interval)
+	withReaper := f.NewWithReaper(ctx, cpReaperKey)
```
Miles-Garnsey, line 257: Issue: Can we have an assertion here to ensure that the right ownerReferences are present? Can we also ensure that all the other resources you're modifying with the ownerReference have a test somewhere in the envtests to ensure that's getting placed?

controllers/k8ssandra/stargate.go
```diff
@@ -52,7 +53,10 @@ func (r *K8ssandraClusterReconciler) reconcileStargate(
 		if err := remoteClient.Get(ctx, stargateKey, actualStargate); err != nil {
 			if errors.IsNotFound(err) {
 				logger.Info("Creating Stargate resource")
-				if err := remoteClient.Create(ctx, desiredStargate); err != nil {
+				if err := controllerutil.SetControllerReference(actualDc, desiredStargate, r.Scheme); err != nil {
+					logger.Error(err, "Failed to set controller reference on Stargate resource")
+					return result.Error(err)
+				} else if err := remoteClient.Create(ctx, desiredStargate); err != nil {
```
Miles-Garnsey, line 59: Nitpick: do we need to use an `else if` here? Style-wise, I prefer a series of basic if statements which return early rather than more complex control flow, but this is obviously a personal preference thing.

controllers/k8ssandra/vector.go
```diff
@@ -57,3 +58,21 @@ func (r *K8ssandraClusterReconciler) reconcileVector(
 	dcLogger.Info("Vector Agent ConfigMap successfully reconciled")
 	return result.Continue()
 }
+
+// setupVectorCleanup adds owner references to ensure that the remote resources created by reconcileVector are correctly
+// cleaned up when the CassandraDatacenter is deleted. We do that in a second pass because the CassandraDatacenter did
+// not exist yet at the time those resources were created.
+func (r *K8ssandraClusterReconciler) setupVectorCleanup(
+	ctx context.Context,
+	kc *k8ssandraapi.K8ssandraCluster,
```
Miles-Garnsey, line 67: Suggestion: with these functions, it would be nice not to pass the whole chunk of `kc` state, and just pass a string instead. Since I think we're just grabbing the kc.SanitizedName() in all cases.

Miles-Garnsey (changes requested):
I started doing a review on this before I noticed Vincent had already done some work on it. Had a few comments and suggestions, also one big architecture question around whether this is going to orphan resources at the worst possible time for users.

I'm still testing manually, had issues doing so on a multi-cluster setup (no log output to indicate an error but the k8ssandraCluster never reconciled to a CassDC), so I'm going to see if a single cluster setup works better.

Miles-Garnsey:
I'm having trouble manually testing this, have tried a few times now. 

I'm running:

```
make single-up && kubectl apply -f build/k8ssandra-cluster.yaml
```

But this doesn't reconcile to a cassDC - it just gives me:

```
Every 2.0s: kubectl --context kind-k8ssandra-0 get cassandradatacenters --all-namespaces                                                                                  Unknown-CJR0QR29NM: Tue Sep 24 12:37:16 2024

No resources found
```

and 

```
2024-09-24T10:28:40.825Z	INFO	k8ssandracluster-webhook	validate K8ssandraCluster create	{"K8ssandraCluster": "test"}
W0924 10:34:24.982370       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
W0924 10:34:44.960130       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
W0924 10:35:04.961546       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
W0924 10:35:30.957564       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
W0924 10:36:23.959466       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
W0924 10:36:38.953898       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
2024-09-24T10:36:39.866Z	INFO	k8ssandracluster-webhook	validate K8ssandraCluster update	{"K8ssandraCluster": "test"}
W0924 10:36:53.953320       1 warnings.go:70] Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
```

from the logs.

I'll try again with the same procedure on the latest release branch and see if it also has the same behaviour, my k8ssandraCluster might have some broken since I'm trying to turn on Reaper and Medusa both configured with `medusa: {}` and `reaper: {}` and that might not be supported.

adejanovski:
@Miles-Garnsey, you shouldn't be able to enable medusa by creating an empty object. It's going to require some properties to be defined.
Reaper will deploy fine with `reaper: {}` though.

controllers/k8ssandra/auth_test.go
```diff
@@ -127,8 +127,6 @@ func createSingleDcClusterNoAuth(t *testing.T, ctx context.Context, f *framework
 	err = f.DeleteK8ssandraCluster(ctx, client.ObjectKey{Namespace: kc.Namespace, Name: kc.Name}, timeout, interval)
 	require.NoError(t, err, "failed to delete K8ssandraCluster")
 	f.AssertObjectDoesNotExist(ctx, t, dcKey, &cassdcapi.CassandraDatacenter{}, timeout, interval)
-	f.AssertObjectDoesNotExist(ctx, t, stargateKey, &stargateapi.Stargate{}, timeout, interval)
```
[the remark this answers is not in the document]
olim7t: Owner references are not actually cleaned up in env tests, so we verify the presence of the reference instead of the deletion. See also the comment on `IsOwnedByCassandraDatacenter`.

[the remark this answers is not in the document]
Miles-Garnsey: Ah, I see. OK, so a test for this functionality wouldn't work anyway because OwnerReferences don't result in a cleanup? But because this was previously manually cleaned up it used to work? 

I'm OK to resolve this comment if so - it makes sense to me. I'd forgotten how limited the envests were.

controllers/k8ssandra/cleanup.go
```diff
@@ -350,129 +184,35 @@ func (r *K8ssandraClusterReconciler) findDcForDeletion(
...
-			}
-		}
-	}
-	return
-}
-
-func (r *K8ssandraClusterReconciler) deleteServices(
+// setDcOwnership loads the remote resource identified by controlledKey, sets dc as its owner, and writes it back. If
+// the remote resource does not exist, this is a no-op.
+func setDcOwnership[T client.Object](
```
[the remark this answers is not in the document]
olim7t: It is a no-op:
```go
	if err := remoteClient.Get(ctx, controlledKey, controlled); err != nil {
		if errors.IsNotFound(err) {
			return result.Continue()
		}
```
I did it that way because it's not always trivial to determine if the resource is supposed to exist (based on the K8ssandraClusterSpec); we'd have to duplicate logic from the creation code.

All those resources are required for the CassandraDatacenter to start correctly, so I think we can safely assume that they exist.

controllers/k8ssandra/medusa_reconciler.go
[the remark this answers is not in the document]
olim7t: > If the cassDC creation fails, then doesn't the running order here mean that deletion of the k8ssandraCluster will not result in cleanup of the descendant resources?

Correct, that is an issue.

> How much effort would it be to ensure that Reaper and Medusa are created only after the CassDC?

That would help, but I don't think it can be done for all resources, some must exist before the CassandraDatacenter (e.g. ConfigMaps that we mount as volumes on the pod).

Miles-Garnsey:
I tried a variety of things to test this, including removing the `medusa: {}` block, and also reverting to the simplest manifest I could think of (on `main` no less!)

I still see no progress towards rollout and no logs from the operator. Will continue trying to test tomorrow.

```
kind: K8ssandraCluster
metadata:
  name: test
spec:
  cassandra:
    serverVersion: 4.0.4
    datacenters:
      - metadata:
          name: dc1
        size: 1
        storageConfig:
          cassandraDataVolumeClaimSpec:
            storageClassName: standard
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 5Gi
        config:
          jvmOptions:
            heapSize: 384Mi
```

controllers/k8ssandra/medusa_reconciler.go
olim7t: > Conversely, actually if it is `IsDone()` then you can continue the reconciliation

No, I think what `IsDone()` means is "the current reconciliation loop is done", in other words: drop everything we're doing and exit the main reconciliation function without any requeues.

When we want to continue the reconciliation, we use `Continue()`,  and `Continue().Completed()` is false. In fact it's the only implementation that returns false, so `Completed()` means "_is not_ `Continue()`".

So I think the use of `Completed()` here is correct: we set the ownership of the ConfigMap;
- if that succeeded, it returns `Continue()` and we proceed with the ownership of the CronJob
- if that didn't succeed (it errored out, was requeued, or we decided the whole reconciliation loop was fulfilled), then we return without addressing the CronJob

controllers/k8ssandra/reaper.go
olim7t: It is inside `reconcileReaper()`, which handles the creation of the `Reaper` k8s resource.

controllers/k8ssandra/reaper_test.go
olim7t: This one is actually a bit of a special case, because the control plane Reaper is not owned by a DC.

All other cases go through `f.ReaperExists()`, which now has the ownership check baked in.

controllers/k8ssandra/stargate.go
olim7t: I don't have a strong preference either way, and indeed "no else" seems to be the most prevalent in the codebase. However this particular file uses the "else" style all over, so I'm going to stick to it for consistency...

controllers/k8ssandra/cleanup.go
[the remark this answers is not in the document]
Miles-Garnsey: I have reservations about this because if the resource name changes (e.g. due to cluster name overrides or some other weird piece of logic) we could easily end up in a situation where the resources aren't marked with an owner correctly, and this failure would happen silently.

controllers/k8ssandra/medusa_reconciler.go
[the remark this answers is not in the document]
Miles-Garnsey: If this is a great deal of effort (and it sounds like it might be...) can I suggest that we raise a refactoring ticket to implement a new kind: CassandraDatacenterRequest which is used as the owner until the real CassDC comes online? Then the Request object can be cleaned up and trigger the cleanups we need, ensuring that this happens even if the CassDC creation fails.

Miles-Garnsey:
I can't figure out what I'm doing wrong here and am giving up on manually testing this. I'm running:

```
make single-up && \
kubectl apply -f test/testdata/fixtures/single-dc/k8ssandra.yaml
```

This is on `main` and involves a cluster definition directly used in the e2e tests. Somehow I still can't get a cluster up. The last log line I see from k8ssandra-operator is `2024-09-25T14:28:29.609Z	INFO	k8ssandracluster-webhook	validate K8ssandraCluster create	{"K8ssandraCluster": "test"}`

No errors are logged, so I'm not sure what's causing the blockage. Obviously whatever test scripts I had working previously no longer work, or something is broken in the makefile on `main`. 

I'll leave my comments as they are, since I think there are some questions that we need to answer that entail either some changes in this PR or some follow up refactoring tickets, since as things stand I think we're going to end up orphaning resources quite often.

controllers/k8ssandra/medusa_reconciler.go
burmanm: > The issue I have with Completed() is that it is unclear what is completed. It doesn't seem to match the use of Done(), so that's problematic IMO.

Done() and Completed() are different. Done() is indeed Completed also, but Error() or CallbackSoon() are not Done(), but are Completed(). 

Completed() means the current reconcile call is done as far as we can right now, but does not mean our entire reconciliation process is done. We might requeue through error or requeue request to process it again. Done() is exactly that, we will not return to that reconcile part without any external changes and this process will not requeue.

This process hasn't changed since 2020 from where it was copied from. The tests indicate that as well.

https://github.com/datastax/cass-operator/blob/v1.0.0/operator/internal/result/result_helper_test.go

Miles-Garnsey: I don't think those tests are sufficient documentation to indicate that the reconcile.Result type is meant to work that way @burmanm. They are talking about "reconciling being complete" but that clearly isn't what `Completed()` is actually intended to mean. As you say, it is probably intended that "completed" here means that the current reconciliation run is completed, but reconciliation as a whole is not complete. 

Again, **what particular reconciliation action** is done or completed, or needs to be requeued is not clear from any of this logic, nor the tests. And for ReconcileObject, `Done()` is used to indicate that the action was successful. So that's why I suggest we make things more explicit by avoiding the use of `Completed()` and instead returning when something explicitly needs to be requeued or has errored.

olim7t: > what particular reconciliation action is done or completed [...] is not clear

As I understand it, the action under consideration is always the top-level reconciliation loop, for example `K8ssandraClusterReconciler.Reconcile()`. So if a component returns `Done()`, it should "bubble up" the stack: every caller stops what it was doing and immediately returns, until we reach `Reconcile()` and return `ctrl.Result{}` to the runtime.

@burmanm would you agree with that assessment?

olim7t: #1419

controllers/k8ssandra/medusa_reconciler.go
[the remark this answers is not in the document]
adejanovski: I'd rather not introduce a new bogus object to deal with cleanups on failed CassDC creations.
The assumption is that whatever prevents the creation of the cassdc will get resolved through another apply operation to patch the original K8ssandraCluster.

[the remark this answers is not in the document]
Miles-Garnsey: I don't think I'd describe this as a "bogus object", this is similar to the pattern followed by cert manager and k8s native APIs for cert issuance, where things might fail without any fault on the part of the object creator.

> I'd rather not introduce a new bogus object to deal with cleanups on failed CassDC creations.
The assumption is that whatever prevents the creation of the cassdc will get resolved through another apply operation to patch the original K8ssandraCluster.

I don't understand why we'd make that assumption. There are too many things that could fail here (sometimes in non-recoverable ways). E.g. storage class can't be changed in an existing DC from memory - if the user then wants to delete the DC to re-create with a new storage class, that will always result in orphaned objects under the current design.

If we're fine with orphaning objects then we can leave it as is, but I think this is going to cause some level of pain down the track.

PR #1028, opened 2023-07-26
Fix system logger image config

### LINKED ISSUE #1027: ImageConfig is overriding the system-logger tag to be latest
cass-operator defines that the server-system-logger image tag to use is `v1.16.0`, but in k8ssandra-operator we override that configMap and put `latest` instead.
The newly pushed `latest` tag is currently breaking if we don't use the `latest` tag as well for cass-operator.

We need to use a fixed version instead of `latest` to ensure we don't break clusters when updating the latest tag (and probably stop creating `latest` tags altogether).

## Definition of Done
- [ ] the ImageConfig overrides in our kustomizations point the system-logger image to the v1.16.0 tag instead of latest

adejanovski:
**What this PR does**:
Point the system-logger to v1.16.0 instead of latest in the ImageConfig override.

Fixes #1027 

**Checklist**
- [x] Changes manually tested
- [ ] ~Documentation added/updated~

### DIFF
--- a/CHANGELOG/CHANGELOG-1.9.md
+++ b/CHANGELOG/CHANGELOG-1.9.md
@@ -15,6 +15,7 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+* [BUGFIX] [#1027](https://github.com/k8ssandra/k8ssandra-operator/issues/1027) Point system-logger image to use the v1.16.0 tag instead of latest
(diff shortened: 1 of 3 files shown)


bradfordcp:
Could we pull in the new registry as part of this (cr.k8ssandra.io)?

burmanm:
> Could we pull in the new registry as part of this (cr.k8ssandra.io)?

I thought we don't want to do that in the tests?

PR #1030, opened 2023-07-26
Fix replication dcname overrides

### LINKED ISSUE #1026: System replication annotations aren't using the dc name overrides
CassandraDatacenter can have dc names overrides, in which case the cassdc custom resource will be named differently than the datacenter in Cassandra's rack-dc properties.
Currently k8ssandra-operator uses the CR name to set the replication of system keyspaces through an annotation, instead of using the override. This override is what Cassandra knows, not the CR name.

We need to modify these annotations so that the override is used instead of the technical name of the cassdc resource.

```[tasklist]
### Definition of Done
- [x] The `k8ssandra.io/initial-system-replication` uses the dc name override instead of the dc resource name when it exists
- [x] All E2E tests pass with DC name overrides
```

adejanovski:
**What this PR does**:
There are several places in the code where we don't account for the DC name overrides, notably when placing annotations that include dc names, and when creating restore mappings.
This PR addresses these problems and adds testing to avoid regressions.
Most E2E tests are now using a DC name override (at the notable exception of the upgrade one because older versions of the CRD didn't have the DatacenterName field).
This led to a lot of changes in the tests to replace many hardcoded references to the dc names with some future proof name discovery logic.

Fixes #1026

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/workflows/kind_multicluster_e2e_tests.yaml
+++ b/.github/workflows/kind_multicluster_e2e_tests.yaml
@@ -59,7 +59,7 @@ jobs:
         e2e_test:
           - CreateMultiDatacenterCluster
           - CreateMixedMultiDataCenterCluster
-          - AddDcToCluster
+          - AddDcToClusterDiffDataplane
           - RemoveDcFromCluster
           - CheckStargateApisWithMultiDcCluster
           - CreateMultiStargateAndDatacenter
--- a/controllers/k8ssandra/cleanup.go
+++ b/controllers/k8ssandra/cleanup.go
@@ -139,7 +139,7 @@ func (r *K8ssandraClusterReconciler) checkFinalizer(ctx context.Context, kc *api
 }
 
 func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *api.K8ssandraCluster, logger logr.Logger) result.ReconcileResult {
-	dcName := k8ssandra.GetDatacenterForDecommission(kc)
+	dcName, dcNameOverride := k8ssandra.GetDatacenterForDecommission(kc)
 	if dcName == "" {
 		return result.Continue()
 	}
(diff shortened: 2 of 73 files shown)


.github/workflows/kind_multicluster_e2e_tests.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
         e2e_test:
           - CreateMultiDatacenterCluster
           - CreateMixedMultiDataCenterCluster
-          - AddDcToCluster
+          - AddDcToClusterDiffDataplane
```
adejanovski, line 62: This test was renamed because we have another test with the `AddDcToCluster` prefix (`AddDcToClusterSameDataplane`), which was executed each time `AddDcToCluster` was invoked.
This is because subtests in Go seem to execute every test that starts with the configured value.

controllers/k8ssandra/cleanup.go
```diff
@@ -157,33 +157,38 @@ func (r *K8ssandraClusterReconciler) checkDcDeletion(ctx context.Context, kc *ap
 		return result.Continue()
 	default:
 		logger.Info("Proceeding with DC deletion", "DC", dcName)
-		return r.deleteDc(ctx, kc, dcName, logger)
+
+		cassDcName := dcName
+		if dcNameOverride != nil && *dcNameOverride != "" {
+			cassDcName = *dcNameOverride
+		}
```
adejanovski, line 164: This is covered in the e2e tests, which codecov cannot detect.

pkg/reaper/resource.go
```diff
@@ -24,7 +24,7 @@ const (
 // DefaultResourceName generates a name for a new Reaper resource that is derived from the Cassandra cluster and DC
 // names.
 func DefaultResourceName(dc *cassdcapi.CassandraDatacenter) string {
-	return cassdcapi.CleanupForKubernetes(dc.Spec.ClusterName + "-" + dc.Name + "-reaper")
+	return cassdcapi.CleanupForKubernetes(dc.Spec.ClusterName + "-" + dc.DatacenterName() + "-reaper")
```
adejanovski, line 27: Tested in the e2e tests.

test/e2e/auth_test.go
```diff
@@ -29,11 +29,11 @@ func multiDcAuthOnOff(t *testing.T, ctx context.Context, namespace string, f *fr
 	dc1Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
 	dc2Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[1], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc2"}}
 
-	reaper1Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "cluster1-dc1-reaper"}}
-	reaper2Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[1], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "cluster1-dc2-reaper"}}
+	reaper1Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "cluster1-real-dc1-reaper"}}
+	reaper2Key := framework.ClusterKey{K8sContext: f.DataPlaneContexts[1], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "cluster1-real-dc2-reaper"}}
```
adejanovski, line 33: I couldn't use the `DcPrefix()` function here because the DC object doesn't exist yet.

burmanm:
TODO: The last commit is not right and shouldn't be done. We need to have tests without dc name override as default. Otherwise the entire purpose of "metadata/name" goes away and should be removed and we create uuids for all datacenters. 

This PR however does not do that, but now we've removed all the normal behavior testing, thus user could be using the default settings and end up with non-working setup.

Special cases should only have their special tests, not be part of every test. Otherwise we risk the possibility that we start to depend on every possible extra setting being available (which should then be added to all e2e tests) and cases where there's less settings set things will no longer work.

pkg/k8ssandra/util.go
```diff
@@ -5,7 +5,7 @@ import (
 	"k8s.io/utils/strings/slices"
 )
 
-func GetDatacenterForDecommission(kc *api.K8ssandraCluster) string {
+func GetDatacenterForDecommission(kc *api.K8ssandraCluster) (string, *string) {
```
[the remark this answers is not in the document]
adejanovski: sure thing 👍

burmanm (approved):

PR #1047, opened 2023-09-13
Add detailed backup information in the MedusaBackup CRD status

### LINKED ISSUE #1046: Add backup information in the MedusaBackup status
The MedusaBackup object doesn't contain much currently that we can leverage.
We should add more information there which could be leveraged by the operator for pre-checks, as well as human inspection:

- Number of nodes that completed the backup
- Topology at the time of the backup

```[tasklist]
### Definition of Done
- [x] A new `.status.total_nodes` field is added to indicate the total number of nodes in the cluster at the time of the backup
- [x] A new `.status.finished_nodes` field is added to indicate the number of nodes that have completed this backup
- [x] A new `.status.nodes` field is added to store the `Nodes` struct from the backup summary that contain the list of nodes at the time of the backup with their tokens and placement
```

adejanovski:
**What this PR does**:
Adds backup summary information in the MedusaBackup CRD status.
These additional field are:
- Total number of nodes
- Number of nodes with successful backup
- List of nodes with their token ranges and topology placement
- Backup status

It also adds kubebuilder printcolumn directives to enrich the kubectl get outputs for MedusaBackupJob and MedusaBackup objects.

Fixes #1046

**Checklist**
- [x] Changes manually tested


controllers/medusa/medusabackupjob_controller.go
```diff
@@ -210,7 +215,25 @@ func (r *MedusaBackupJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
 }
 
-func (r *MedusaBackupJobReconciler) createMedusaBackup(ctx context.Context, backup *medusav1alpha1.MedusaBackupJob, logger logr.Logger) error {
+func (r *MedusaBackupJobReconciler) getBackupSummary(ctx context.Context, backup *medusav1alpha1.MedusaBackupJob, pods []corev1.Pod, logger logr.Logger) (*medusa.BackupSummary, error) {
+	for _, pod := range pods {
+		if remoteBackups, err := GetBackups(ctx, &pod, r.ClientFactory); err != nil {
```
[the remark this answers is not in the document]
adejanovski: I think it's because the BackupSummary contains information from all nodes, while BackupStatus is for an individual node. And yeah, it's definitely overkill to read all backups just to get one (it can be quite expensive when you have a lot of backups in storage).
It would be good to be able to filter server side and get a single backup info though. The `medusa status` command already implements reading a [single backup](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/status.py#L30) to get completion information. We could enrich it with the topology info and expose it in the GRPC server for use here.
I'll create a refactoring ticket for this.

controllers/medusa/medusabackupjob_controller.go
```diff
@@ -128,7 +128,12 @@ func (r *MedusaBackupJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		logger.Info("backup complete")
 
 		// The MedusaBackupJob is finished and we now need to create the MedusaBackup object.
-		if err := r.createMedusaBackup(ctx, backup, logger); err != nil {
+		backupSummary, err := r.getBackupSummary(ctx, backup, pods, logger)
+		if err != nil {
+			logger.Error(err, "Failed to get backup summary")
+			return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
burmanm, line 134: suggestion: If err != nil, the controller-runtime will requeue in any case. Setting Requeue time when we know requeue is going to happen in any case is confusing. 

The RequeueAfter value is actually ignored, since the err != nil causes the controller-runtime to process that first and then do a RateLimited queueing:

https://github.com/kubernetes-sigs/controller-runtime/blob/8117577334339e46bc27a46dbf98a0112aef6bb4/pkg/internal/controller/controller.go#L322

burmanm (approved):

controllers/medusa/medusabackupjob_controller.go
adejanovski: sounds good. I thought about that but tried to stay consistent with the rest of the code in this controller.
I'll create a refactoring ticket to make this change everywhere in the medusa controllers.

PR #1070, opened 2023-09-29
Link prominently to docs site in README

Miles-Garnsey:
I get very sad when docs don't link prominently to repos, or repos don't link prominently to docs.

Currently, a Google search for k8ssandra-operator returns this repo as the first result, but the README is not a good source for documentation. Sadly, the docs link is at the very bottom of the page. This PR fixes that.


adejanovski (approved):
Approved, so you can rejoice Miles ;)

PR #1091, opened 2023-10-19
Make HttpManagement.Enabled a pointer

Miles-Garnsey:
**What this PR does**:

The HttpManagement.Enabled in the Reaper package should be a pointer, to allow for easier distinction between unset and false values. This is not important for k8ssandra-operator (where a default value of false is fine) but is relevant for downstream consumers of the types, in particular MC, where we would like to have a default value of true.

Hotfix.

**Checklist**
- [ ] Changes manually tested


pkg/reaper/deployment.go
```diff
@@ -152,7 +152,7 @@ func NewDeployment(reaper *api.Reaper, dc *cassdcapi.CassandraDatacenter, keysto
 			Value: fmt.Sprintf("%d", reaper.Spec.HeapSize.Value()),
 		})
 	}
-	if reaper.Spec.HttpManagement.Enabled {
+	if reaper.Spec.HttpManagement.Enabled != nil && *reaper.Spec.HttpManagement.Enabled {
```
olim7t, line 155: [suggestion] not a hard requirement, but there is a library method for this kind of expression
```suggestion
	if pointer.BoolDeref(reaper.Spec.HttpManagement.Enabled, false) {
```

olim7t (approved):
LGTM, don't forget the changelog entry.

PR #1120, opened 2023-11-24
Add validation check for projected pod names length

### LINKED ISSUE #1115: Add a validation check for the projected pod names length
Pod names are limited in length and we should make sure to fail fast when we project an overflow.
The following elements must be accounted for:
- cluster name
- dc name
- rack name
- `-sts-???` part

The validation webhook should reject any attempt to create a cluster that would create pod names which are too long.

```[tasklist]
### Definition of Done
- [ ] Creating a cluster/dc/rack that generates pod names too long
```

adejanovski:
**What this PR does**:
Checks if the projected pod names (for each dc and rack) could overflow the 63 characters limit. Fails the validation if it's the case.

Fixes #1115 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.12.md
+++ b/CHANGELOG/CHANGELOG-1.12.md
@@ -15,3 +15,4 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+- [ENHANCEMENT] [#1115](https://github.com/k8ssandra/k8ssandra-operator/issues/1115) Add a validation check for the projected pod names length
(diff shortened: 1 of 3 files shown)


Miles-Garnsey:
Issue: Have we seen this become an issue in practice? I've been reading [this](https://pauldally.medium.com/why-you-try-to-keep-your-deployment-names-to-47-characters-or-less-1f93a848d34c#:~:text=Most%20Kubernetes%20objects%2C%20including%20Deployment,%E2%89%A4%20253%20characters%20in%20length.) article and it indicates that for deployments, truncation of overlength names happens automatically. 

But based on [this](https://github.com/kubernetes/kubernetes/issues/64023#issuecomment-1148334755) comment I'm not sure if the same applies to statefulsets (particularly because they appear to have other issues with labels), and indeed it may be the case that an STS name over 52 characters will cause issues.

Just trying to figure out what we're doing here.

PR #1180, opened 2024-01-19
Create the MedusaConfiguration API

### LINKED ISSUE #1157: Create a new MedusaConfiguration API
The MedusaConfiguration API schema will use the same type as [Medusa's storage properties](https://docs.k8ssandra.io/reference/crd/releases/k8ssandra-operator-releases/k8ssandra-operator-crds-1.10/#medusaspecstorageproperties).

Name | Type | Description | Required
-- | -- | -- | --
apiProfile | string | AWS Profile to use for authentication. | false
backupGracePeriodInDays | integer | Age after which orphan sstables can be deleted from the storage backend. Protects from race conditions between purge and ongoing backups. Defaults to 10 days. | false
bucketName | string | The name of the bucket to use for the backups. | false
concurrentTransfers | integer | Number of concurrent uploads. Helps maximizing the speed of uploads but puts more pressure on the network. Defaults to 1.Default: 1 | false
host | string | Host to connect to for the storage backend. | false
maxBackupAge | integer | Maximum backup age that the purge process should observe.Default: 0 | false
maxBackupCount | integer | Maximum number of backups to keep (used by the purge process). Default is unlimited.Default: 0 | false
multiPartUploadThreshold | integer | File size over which cloud specific cli tools are used for transfer. Defaults to 100 MB.Default: 104857600 | false
port | integer | Port to connect to for the storage backend. | false
prefix | string | Name of the top level folder in the backup bucket. If empty, the cluster name will be used. | false
region | string | Region of the storage bucket. Defaults to "default". | false
secure | boolean | Whether to use SSL for the storage backend. | false
storageProvider | enum | The storage backend to use for the backups.Enum: local, google_storage, azure_blobs, s3, s3_compatible, s3_rgw, ibm_storage | false
storageSecretRef | string | Kubernetes Secret that stores the key file for the storage provider's API. If using 'local' storage, this value is ignored. | false
transferMaxBandwidth | string | Max upload bandwidth in MB/s. Defaults to 50 MB/s.Default: 50MB/s | false

Initially, the controller for this API would only verify that the referenced secret is available and update the status accordingly.
In the future it could attempt a connection to the bucket and update the status of the object to indicate if the settings are valid and allow a successful connection.

The referenced secret must exist in the same namespace as the MedusaConfiguration object.

```[tasklist]
### Definition of Done
- [ ] A new MedusaConfiguration API is added with using medusa's storage properties as schema
```

adejanovski:
**What this PR does**:
Adds a MedusaConfiguration API.

Fixes #1157 

**Checklist**
- [x] Changes manually tested


controllers/medusa/medusaconfiguration_controller.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to get MedusaConfiguration")
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	configuration := instance.DeepCopy()
+	patch := client.MergeFrom(configuration.DeepCopy())
```
olim7t, line 73: [question] why call `DeepCopy` twice?

olim7t (approved):

controllers/medusa/medusaconfiguration_controller.go
adejanovski: The patch and the original object we're patching need to be in separate addresses for when we invoke `.Status().Patch()`.
I'm not sure why we initially do a DeepCopy() on the object we get from the kube api, but I'm just following the convention we use elsewhere, which could be unnecessary but at least works fine :)

apis/medusa/v1alpha1/medusaconfiguration_types.go
```diff
@@ -0,0 +1,121 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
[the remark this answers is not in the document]
adejanovski: sounds good 👍

PR #1182, opened 2024-01-22
Add the ability to configure Reaper with truststore/keystore

### LINKED ISSUE #1181: Add ability to configure keystore and truststore for Reaper http-api
**What is missing?**
Current CRD does not support Reaper's http-api to configure keystore and truststores.

**Why do we need it?**

**Environment**

* K8ssandra Operator version:

burmanm:
**What this PR does**:
Adds keystore and truststore properties and their configuration and mounting to the Reaper Spec

Fixes #1181

**Checklist**
- [ ] Changes manually tested


apis/reaper/v1alpha1/reaper_types.go
```diff
@@ -378,7 +378,9 @@ type HttpManagement struct {
 	// When enabled, HTTP will be used instead of JMX for management connectivity between Cassandra
 	// and Reaper. In future, this will be true by default
 	// +kubebuilder:default=false
-	Enabled *bool `json:"enabled"`
+	Enabled bool `json:"enabled,omitempty"`
```
olim7t, line 381: [thought] I would be careful changing that, since the comment says it might default to `true` in the future.

The problem with a raw boolean is that in some case it becomes impossible to differentiate between "unset" and "explicitly set to false".

olim7t (approved):
LGTM, but we should test to validate the boolean change.

PR #1202, opened 2024-02-07
Replicate medusa bucket secrets to each Cass DC

### LINKED ISSUE #1159: Replicate bucket key secrets to namespaces hosting clusters that reference the MedusaConfiguration

rzvoncek:
**What this PR does**:

This PR adds a bucket secret reconciliation process. If Medusa is configured globally (meaning the cluster's medusa config features a `medusaConfigurationRef`, then this process will make sure the bucket key secrets are properly replicated into the namespaces (and clusters) of each of the DCs. It will also alter the k8ssandra cluster to use the copied secret, instead of the original one.

This PR does not implement the actual replication, it only makes a copy of the secret and places labels on it. We make a copy so that we can clean up after ourselves (the original secret is user-created and we should not touch it). The labels are then picked up by the secret reconciliation process which actually copies them over.

Fixes #1159.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/medusa_reconciler.go
+++ b/controllers/k8ssandra/medusa_reconciler.go
@@ -4,11 +4,9 @@ import (
 	"context"
 	"fmt"
 	"github.com/adutra/goalesce"
-	medusaapi "github.com/k8ssandra/k8ssandra-operator/apis/medusa/v1alpha1"
-	"k8s.io/apimachinery/pkg/types"
-
 	"github.com/go-logr/logr"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	medusaapi "github.com/k8ssandra/k8ssandra-operator/apis/medusa/v1alpha1"
 	cassandra "github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/labels"
 	medusa "github.com/k8ssandra/k8ssandra-operator/pkg/medusa"
@@ -18,6 +16,8 @@ import (
 	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -32,11 +32,11 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 	kc := desiredKc.DeepCopy()
 	namespace := utils.FirstNonEmptyString(dcConfig.Meta.Namespace, kc.Namespace)
 	logger.Info("Medusa reconcile for " + dcConfig.CassDcName() + " on namespace " + namespace)
-	medusaSpec := kc.Spec.Medusa
-	if medusaSpec != nil {
+	if kc.Spec.Medusa != nil {
 		logger.Info("Medusa is enabled")
 
-		mergeResult := mergeStorageProperties(ctx, remoteClient, namespace, medusaSpec, logger, kc)
+		mergeResult := r.mergeStorageProperties(ctx, r.Client, namespace, kc.Spec.Medusa, logger, kc)
+		medusaSpec := kc.Spec.Medusa
 		if mergeResult.IsError() {
 			return result.Error(mergeResult.GetError())
 		}
@@ -49,6 +49,7 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 				return result.Error(fmt.Errorf("medusa encryption certificates were not provided despite client encryption being enabled"))
 			}
 		}
+
 		if medusaSpec.StorageProperties.StorageSecretRef.Name == "" {
 			medusaSpec.StorageProperties.StorageSecretRef = corev1.LocalObjectReference{Name: ""}
 			if medusaSpec.StorageProperties.CredentialsType == "role-based" && medusaSpec.StorageProperties.StorageProvider == "s3" {
@@ -187,6 +188,11 @@ func (r *K8ssandraClusterReconciler) reconcileMedusaSecrets(
 			logger.Error(err, "Failed to reconcile Medusa CQL user secret")
 			return result.Error(err)
 		}
+
+		if err := r.reconcileBucketSecrets(ctx, r.ClientCache.GetLocalClient(), kc, logger); err != nil {
+			logger.Error(err, "Failed to reconcile Medusa bucket secrets")
+			return result.Error(err)
+		}
 	}
 
 	logger.Info("Medusa user secrets successfully reconciled")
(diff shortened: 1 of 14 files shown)


controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -229,7 +235,7 @@ func mergeStorageProperties(
 	configKey := types.NamespacedName{Namespace: namespace, Name: medusaSpec.MedusaConfigurationRef.Name}
 	if err := remoteClient.Get(ctx, configKey, storageProperties); err != nil {
 		logger.Error(err, fmt.Sprintf("failed to get MedusaConfiguration %s", configKey))
-		return result.Error(err)
+		return result.RequeueSoon(r.DefaultDelay)
```
burmanm, line 238: question: Why the change? The only difference I can see is the change of delay when reconcile happens (error will do a reconcile also - but with a backoff timer to prevent flooding the server which this doesn't).

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -247,3 +253,63 @@ func mergeStorageProperties(
...
+	}
+
+	// fetch the referenced configuration
+	medusaConfigNamespace := utils.FirstNonEmptyString(medusaSpec.MedusaConfigurationRef.Namespace, namespace)
+	medusaConfigName := medusaSpec.MedusaConfigurationRef.Name
+	medusaConfig := &medusaapi.MedusaConfiguration{}
+	medusaConfigKey := types.NamespacedName{Namespace: medusaConfigNamespace, Name: medusaConfigName}
+	if err := remoteClient.Get(ctx, medusaConfigKey, medusaConfig); err != nil {
+		logger.Error(err, "failed to get MedusaConfiguration", "key", medusaConfigKey)
+		return result.RequeueSoon(r.DefaultDelay)
```
burmanm, line 273: question: Why not result.Error(err) ? Now we don't get any proper processing of the error. Only if you wish to stop reconcile should this be result.Done() or something similar, but for requeue, always let the controller-runtime do that one for errors.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -247,3 +253,63 @@ func mergeStorageProperties(
...
+	}
+
+	// fetch the referenced medusa configuration's bucket secret
+	bucketSecretName := medusaConfig.Spec.StorageProperties.StorageSecretRef.Name
+	bucketSecret := &corev1.Secret{}
+	bucketSecretKey := types.NamespacedName{Namespace: medusaConfigNamespace, Name: bucketSecretName}
+	err := remoteClient.Get(ctx, bucketSecretKey, bucketSecret)
+	if err != nil {
+		logger.Error(err, "failed to get bucket secret", "key", bucketSecretKey)
+		return result.RequeueSoon(r.DefaultDelay)
```
burmanm, line 283: question: Same here, result.Error(err) is better.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -247,3 +253,63 @@ func mergeStorageProperties(
...
+
+	// finally, update the storage properties to point to a newly created secret
+	if kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name == clusterBucketSecret.Name {
+		logger.Info("The cluster bucket secret is already set")
+		return result.Continue()
+	} else {
+		newSecretRef := corev1.LocalObjectReference{Name: clusterBucketSecret.Name}
+		kc.Spec.Medusa.StorageProperties.StorageSecretRef = newSecretRef
+		if err = remoteClient.Update(ctx, kc); err != nil {
+			logger.Error(err, "failed to update K8ssandraCluster, re-queueing")
```
burmanm, line 309: .

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -247,3 +253,63 @@ func mergeStorageProperties(
...
+			return result.RequeueSoon(r.DefaultDelay)
```
burmanm, line 310: .

burmanm:
issue: I see no changes in the envtests, would you be able to add those as well? That provides so much faster response to changes so it helps other developers more than the e2e tests (which realistically, no one is running on their machine and wait until GHA throws something and they're not as easily debuggable).

controllers/k8ssandra/medusa_reconciler.go
rzvoncek: I had no good reason for this. Saw some weird behaviour and though requeueing is going to help. Now I know better.
Reverted to returing the error.

controllers/k8ssandra/medusa_reconciler.go
rzvoncek: Also, changed to return the error.

rzvoncek:
Addressed the issue. Added env tests and updated e2e tests. 
While doing this, also changed when we actually do the bucket secret reconciliation. Now it's not happening for each DC, but only once, at the place where we deal with all Medusa-related secrets.

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -119,7 +120,7 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 		}
 		if !ready {
 			logger.Info("Medusa standalone deployment is not ready yet")
-			return result.RequeueSoon(r.DefaultDelay)
+			return result.Error(err)
```
burmanm, line 123: There's no err necessarily here, the method can return a `false, nil` in which case err is nil here.

controllers/k8ssandra/medusa_reconciler.go
burmanm: issue: This error handling is still incorrect, in case of error, err should be returned for proper requeue process (especially in this case when it can be Kubernetes API server overload or similar and we don't want to make it worse)

controllers/k8ssandra/medusa_reconciler.go
rzvoncek: Great spot, fixed.

rzvoncek:
Did another round of improvements. The return error vs. requeue was indeed a problem. All green now (aside from codecov).

docs/content/en/tasks/backup-restore/_index.md
```diff
@@ -106,9 +106,11 @@ stringData:
 
 The file should always specify `credentials` as shown in the example above; in that section, provide the expected format and credential values that are expected by Medusa for the chosen storage backend. For more, refer to the [Medusa documentation](https://github.com/thelastpickle/cassandra-medusa/blob/master/docs/Installation.md) to know which file format should used for each supported storage backend.
 
-A successful deployment should inject a new init container named `medusa-restore` and a new container named `medusa` in the Cassandra StatefulSet pods.  
+If using a shared Medusa configuration (see below), this secret can be created in the same namespace as the `MedusaConfiguration` object. The K8ssandra operator will then make sure the secret is replicated to the namespaces hosting the Cassandra clusters.
```
adejanovski, line 109:
```suggestion
If using a shared Medusa configuration (see below), this secret has to be created in the same namespace as the `MedusaConfiguration` object. k8ssandra-operator will then make sure the secret is replicated to the namespaces hosting the Cassandra clusters.
```

adejanovski (approved):
Brilliant!
Everything worked perfectly in my tests, with both namespace and cluster scope.

I put a suggestion change for some wording in the docs.

PR #1204, opened 2024-02-08
Add new ConcurrencyPolicy setting to the MedusaBackupSchedule

### LINKED ISSUE #1203: Add ConcurrencyPolicy to MedusaBackupSchedule
**What is missing?**
Adding ConcurrencyPolicy that would either allow or prevent concurrent backups happening in the cluster. I think the default should be Forbid (just like with CassandraTasks), but lets allow user to modify the behavior if they so wish.

**Why do we need it?**
In many cases, having multiple backups running simultaneously in the same cluster might cause issues with overload or disk usage. 

**Environment**

* K8ssandra Operator version:

burmanm:
…ent multiple active backups working on the same Datacenter

**What this PR does**:
Prevents multiple backup jobs from running concurrently when targeting the same Datacenter. This works with the internal MedusaBackupScheduler and can be disabled using ConcurrencyPolicy Allow.

Fixes #1203 

**Checklist**
- [ ] Changes manually tested


controllers/medusa/medusabackupschedule_controller.go
```diff
@@ -83,6 +95,16 @@ func (r *MedusaBackupScheduleReconciler) Reconcile(ctx context.Context, req ctrl
 	createBackup := false
 
 	if nextExecution.Before(now) {
+		if backupSchedule.Spec.ConcurrencyPolicy == batchv1.ForbidConcurrent {
+			if activeTasks, err := r.activeTasks(backupSchedule, dc); err != nil {
+				return ctrl.Result{}, err
+			} else {
+				if len(activeTasks) > 0 {
+					logger.V(1).Info("Postponing backup schedule due to existing active backups", "MedusaBackupSchedule", req.NamespacedName)
```
rzvoncek, line 103: [Question]: I've not come across `logger.V(1)` yet. What does it do?

charts/k8ssandra-operator/templates/clusterrole.yaml
rzvoncek, line 1: [Question] We're adding this (and also the `charts/k8ssandra-operator/templates/role.yaml`) file. Why did this happen?

rzvoncek (approved):
The changes look good to me, but there are a few thing I did not understand (see the questions).

controllers/medusa/medusabackupschedule_controller.go
burmanm: It's the same as what in Java you would call for "logger.debug". There's multiple levels (not sure if there's even any limit) of logging you can set, but usually V(1) is for debugging in Go projects and rest are not used.

PR #1238, opened 2024-03-12
Use ReplicatedSecrets to control Medusa bucket secrets replication into K8ssandraCluster namespaces

### LINKED ISSUE #1217: Updating a medusaConfiguration referenced secret should propagate
When using a medusaConfigurationRef, the original secret is copied and then replicated to the contexts and namespaces involved.
But in case the secret is updated due to credentials rotation for example, the copy doesn't get updated, which prevents from replicating the changes.

We need to make sure such secret updates get picked up by the operator and that the secret copy gets refreshed so that all medusa containers can get the new credentials.

```[tasklist]
### Definition of Done
- [ ] Rotating a medusaConfiguration secret propagates to all clusters using the configuration
```

Miles-Garnsey:
**What this PR does**:

Switch to using ReplicatedSecrets to control replication of the medusa bucket secrets.

Fixes #1217 
#1260 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.14.md
+++ b/CHANGELOG/CHANGELOG-1.14.md
@@ -14,4 +14,5 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
 ## unreleased
+* [FEATURE] [#1242](https://github.com/k8ssandra/k8ssandra-operator/issues/1242) Allow for creation of replicated secrets with a prefix, so that we can distinguish between multiple secrets with the same origin but targeting different clusters.
 * [BUGFIX] [#1226](https://github.com/k8ssandra/k8ssandra-operator/issues/1226) Medusa purge cronjob should be created in the operator namespace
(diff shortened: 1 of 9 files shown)


Miles-Garnsey:
I've marked out the places where I think these changes should take place, and commented with the specifics of the changes I intend to make.

There is a slight issue to work around here as I've currently drafted the design. It relates to the fact that (in my current round of comments) I talk about creating a single ReplicatedSecret for a single MedusaConfig, but this probably won't work for the following reason;

* When a cluster is created, it would presumably go and create a replicatedSecret if one didn't already exist in the MedusaConfig namespace.
* But if a replicatedSecret did exist, then it would go and add an entry to the ReplicationTargets in the ReplicatedSecret.
* The problem arises on cluster deletion, where the finalizer might go and either remove the replicationTarget entry or remove the secret (if no more targets exist).
* However, in the case that multiple K8ssandraClusters existed in the same context-namespace, this might cause the deletion of one cluster to delete replicated secrets relied on by another. 

As a result, I think we're going to have to 

1. Create multiple independant ReplicatedSecrets per cluster in the MedusaConfig's namespace.
2. Somehow ensure that (if two clusters reference the same MedusaConfig and are in the same namespace) the secrets are renamed while being replicated so that we don't end up with a conflict in the K8ssandraCluster's namespace instead.

@adejanovski let me know if you agree with this approach, and with the locations in the code where I've proposed changes. I haven't worked on the Medusa Controller extensively, so your input is appreciated.

Miles-Garnsey:
@burmanm thinks there might be a way to create replicas of the Medusa StorageSecretRef without needing to add a new name-namespace type selector to identify the secret. 

We are going to catch up about this next week to see if we can progress this work.

Miles-Garnsey:
Closing as I no longer think this is required.

PR #1243, opened 2024-03-13
ReplicatedSecrets can have a prefix per target

### LINKED ISSUE #1242: Allow ReplicatedSecrets's targets to have a prefix
**What is missing?**

We should allow the target secret's name for a replicated secret to have a prefix which distinguishes it from other replicatedSecrets with the same name that may target the same namespace.

This is so that multiple K8ssandraCluster's residing in the same namespace do not create multiple ReplicatedSecrets which then conflict with one another due to having the same name.

**Why do we need it?**

[This](https://github.com/k8ssandra/k8ssandra-operator/issues/1217) issue will be easier to solve if we can create multiple ReplicatedSecrets at the MedusaConfig's origin which then map to multiple secrets prefixed with the cluster name in the target. In the case that multiple clusters exist in the target context-namespace this could cause conflicts and competition between the reconcile loops for the clusters unless we can distinguish the secrets via their names with a prefix.

**Environment**

* K8ssandra Operator version:

Miles-Garnsey:
**What this PR does**:
Allow a ReplicatedSecret replication target to define a prefix to de-conflict instances where the same source may target the same target multiple times due to presence of several clusters in the target namespace.

Fixes #1242 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/replication/secret_controller.go
+++ b/controllers/replication/secret_controller.go
@@ -225,13 +225,8 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request) 
 			} else {
 				namespace = target.Namespace
 			}
-			if remoteClient == localClient && target.Namespace == sec.Namespace {
-				//TODO: @Michael does this allow us to remove the UID check in requiresUpdate or any other logic?
-				// If if origin and source context-namespace are the same we want to bail immediately.
-				continue TargetSecrets
-			}
 			fetchedSecret := &corev1.Secret{}
-			if err = remoteClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s%s", target.TargetPrefix, sec.Name), Namespace: namespace}, fetchedSecret); err != nil {
+			if err = remoteClient.Get(ctx, types.NamespacedName{Name: getPrefixedSecretName(target.TargetPrefix, sec.Name), Namespace: namespace}, fetchedSecret); err != nil {
 				if errors.IsNotFound(err) {
 					logger.Info("Copying secret to target cluster", "Secret", sec.Name, "TargetContext", target)
 					// Create it
@@ -239,7 +234,7 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request) 
 					copiedSecret.Namespace = namespace
 					copiedSecret.ResourceVersion = ""
 					copiedSecret.OwnerReferences = []metav1.OwnerReference{}
-					copiedSecret.Name = fmt.Sprintf("%s%s", target.TargetPrefix, sec.Name)
+					copiedSecret.Name = getPrefixedSecretName(target.TargetPrefix, sec.Name)
 					if err = remoteClient.Create(ctx, copiedSecret); err != nil {
 						logger.Error(err, "Failed to sync secret to target cluster", "Secret", copiedSecret.Name, "TargetContext", target)
 						break TargetSecrets
@@ -260,7 +255,7 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request) 
 				logger.Info("Modifying secret in target cluster", "Secret", sec.Name, "TargetContext", target)
 				syncSecrets(sec, fetchedSecret)
 				copiedSecret := fetchedSecret.DeepCopy()
-				copiedSecret.Name = fmt.Sprintf("%s%s", target.TargetPrefix, sec.Name)
+				copiedSecret.Name = getPrefixedSecretName(target.TargetPrefix, sec.Name)
 				if err = remoteClient.Update(ctx, copiedSecret); err != nil {
 					logger.Error(err, "Failed to sync target secret for matching payloads", "Secret", fetchedSecret.Name, "TargetContext", target)
 					break TargetSecrets
@@ -295,9 +290,6 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request) 
 }
 
 func requiresUpdate(source, dest client.Object) bool {
-	// TODO: we need to revisit this logic since I think that the change of name (due to adding a prefix) might cause a new UID
-	// to be generated even in the case where the origin and destination cluster-namespace is the same.
-
 	// In case we target the same cluster
 	if source.GetUID() == dest.GetUID() {
 		return false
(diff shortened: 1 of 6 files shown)


controllers/replication/secret_controller.go
```diff
@@ -286,6 +295,9 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request)
 }
 
 func requiresUpdate(source, dest client.Object) bool {
+	// TODO: we need to revisit this logic since I think that the change of name (due to adding a prefix) might cause a new UID
```
burmanm, line 298: Name itself is not important, but if the source and target have different names, then they're not the same secret, even when replicated inside the same namespace. Thus they would have different UID as they're different objects.

controllers/replication/secret_controller.go
```diff
@@ -312,6 +324,9 @@ func requiresUpdate(source, dest client.Object) bool {
 }
 
 func syncSecrets(src, dest *corev1.Secret) {
+	// TODO: @Michael what does this do, and do I need to worry about differences in the names between the source and destination?
```
burmanm, line 327: It enforces that the destination ObjectMeta is the same as the source ObjectMeta, as DeepCopyInto would not remove additional stuff from the modified destination object.

Miles-Garnsey:
This ticket needs one additional thing. At present the final env test that I've added is failing, as the target secret (which has a prefix in this test) is not being properly deleted after it's source secret is deleted. I don't know quite where this logic is supposed to happen, so I've reached out to @burmanm for some help.

Everything else looks ok however.

burmanm:
> I don't know quite where this logic is supposed to happen

https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/replication/secret_controller.go#L70

controllers/replication/secret_controller.go
Miles-Garnsey: That's what I figured, I guess my question is about this line/comment:

```
// In case we target the same cluster
	if source.GetUID() == dest.GetUID() {
```

Can we remove this if we check whether the origin and destination is the same elsewhere?

controllers/replication/secret_controller.go
Miles-Garnsey: Can I rename this as "removeDCSpecificMetadata" or something similar? Just so future implementors have a better idea what's happening.

Miles-Garnsey:
@burmanm the link you provided to the cleanup logic only seems to deal with the ReplicatedSecret resource:

> I don't know quite where this logic is supposed to happen

> https://github.com/k8ssandra/k8ssandra-operator/blob/main/controllers/replication/secret_controller.go#L70

I'm trying to figure out where the cleanup is that is triggered when the _source secret_ is deleted.

Miles-Garnsey:
I've discussed this with @burmanm and it appears that if a source secret is deleted without the replicatedSecret that references it being deleted, then actually the downstream secret that is produced by the ReplicatedSecret **is not deleted**.

At the moment we do have orphaned resources in a sense, so I will work with Micke to create a new ticket to address that. But this PR is good to go.

controllers/replication/secret_controller.go
```diff
@@ -224,15 +225,21 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request)
 			} else {
 				namespace = target.Namespace
 			}
+			if remoteClient == localClient && target.Namespace == sec.Namespace {
+				//TODO: @Michael does this allow us to remove the UID check in requiresUpdate or any other logic?
```
burmanm, line 229: No, this type of check shouldn't be here. If there's a prefix feature, then the copy could target the same namespace also in the same cluster.

Also, in some cases localClient could be different from remoteClient even if they target the same cluster. At least if the localClient is set to be non-caching client which the remoteClient isn't.

Lets say we have "my-user-supplied-A" and we want to copy it always for reason X to "cluster1-my-user-supplied-A". With this check in place, that won't be possible.

controllers/replication/secret_controller.go
```diff
@@ -224,15 +225,21 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request)
...
+				// If if origin and source context-namespace are the same we want to bail immediately.
+				continue TargetSecrets
+			}
 			fetchedSecret := &corev1.Secret{}
-			if err = remoteClient.Get(ctx, types.NamespacedName{Name: sec.Name, Namespace: namespace}, fetchedSecret); err != nil {
+			if err = remoteClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s%s", target.TargetPrefix, sec.Name), Namespace: namespace}, fetchedSecret); err != nil {
```
burmanm, line 234: suggestion: I think this prefix logic naming secrets should be in a function, or otherwise some change will forget it..

controllers/replication/secret_controller.go
burmanm: But with a prefix, they're not. They should be replicated. 

What we want to avoid is modifying the same object we read currently. A replicated one, even if located on the same place, is not the same object.

controllers/replication/secret_controller.go
burmanm: This function isn't about removing anything Datacenter specific. Sure, we filter out cassandra.datastax.com annotations (also cluster, rack, etc) from the target since these are restricted annotations / labels, but that's not all it does.

It also ensures that the annotation and label values set in the original one are having higher priority than the ones in the target secret.

So lets say the original one has labels A=B, C=D, but someone has modified the target one to have A=E, C=D, F=J. We will overwrite the targets to A=B, C=D, F=J set. We're not removing some existing ones from the target (there are some important ones), just ensuring that the same label can't have different value.

Thus, it's really about syncing metadata and that's where the name of the function comes from. The testSyncSecrets explains the syncing from one target to another.

controllers/replication/secret_controller.go
Miles-Garnsey: That actually makes a good deal of sense now you present it  that way. So you'd like to retain the check on the UIDs (to avoid conflicts where the original tries to overwrite itself) but remove this check to enable a case where a secret is replicating into the same context-namespace but with a different name due to the prefix? I'm ok with this if you can confirm this is what you're thinking.

controllers/replication/secret_controller.go
Miles-Garnsey: I'm removing this check as advised.

controllers/replication/secret_controller.go
Miles-Garnsey: Done as requested.

burmanm (approved):

burmanm (approved):

PR #1244, opened 2024-03-13
Properly use concurrent_transfers from MedusaConfig object

### LINKED ISSUE #1239: Concurrent transfers from a MedusaConfiguration is always overridden by 1
The Medusa storage properties defined in a K8ssandraCluster spec use 1 as kubebuilder default for concurrent_transfers. As this is not the "zero value" for integers, it always wins against the value defined in a MedusaConfiguration.

In order to fix this, while retaining the priority of the storage properties over the MedusaConfiguration, we need to modify the default and set it to 0, and modify the code that generates the medusa.ini configmap entry to swap 0 for 1 in case it wasn't defined in either the MedusaConfiguration nor the storage properties.

```[tasklist]
### Definition of Done
- [ ] `K8ssandra.spec.medusa.storageProperties.concurrentTransfers` defaults to 0 instead of 1 (see [here](https://github.com/k8ssandra/k8ssandra-operator/blob/main/apis/medusa/v1alpha1/medusa_types.go#L77-L78))
- [ ] Before persisting the medusa.ini to the configmap in the K8ssandraCluster reconcile, swap 0 for 1 if that is the value we get out of the Goalesce merge
```

rzvoncek:
**What this PR does**:
This PR changes the default value to 0 (instead of 1). This way, the merging storage properties, the correct value wins. 
This PR also makes sure that if we do end up with a 0, the medusa.ini ends up with a 1 because Medusa can't handle 0 concurrent transfers.

Fixes #1239

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.13.md
+++ b/CHANGELOG/CHANGELOG-1.13.md
@@ -16,6 +16,7 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 ## unreleased
 
 * [BUGFIX] [#1235](https://github.com/k8ssandra/k8ssandra-operator/issues/1235) Prevent operator from modifying the spec when superUserRef is not set. Also, remove the injection to mount secrets to cassandra container.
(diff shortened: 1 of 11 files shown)


adejanovski (approved):
Perfect, thanks!

PR #1246, opened 2024-03-16
Reject cluster creation if a DC has an invalid name (fixes #1141)

### LINKED ISSUE #1141: DC names are not properly sanitized
If a cluster declares a datacenter with invalid characters:
```yaml
kind: K8ssandraCluster
metadata:
  namespace: k8ssandra-operator
  name: demo
spec:
  cassandra:
    serverVersion: 4.1.0
    datacenters:
      - metadata:
          name: dc_1   # <<< invalid Kubernetes name
        size: 1
        racks:
          - name: rack1
        storageConfig:
          cassandraDataVolumeClaimSpec:
            storageClassName: standard
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 100Mi
```

The name is not properly sanitized, which leads to this error:
```
ConfigMap "demo-dc_1-metrics-agent-config" is invalid: metadata.name: Invalid value: "demo-dc_1-metrics-agent-config": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

The problem is not limited to that ConfigMap: even after fixing it ([here](https://github.com/k8ssandra/k8ssandra-operator/blob/20398713516abf74912f6d4226d683a3f8c3def5/pkg/telemetry/cassandra_agent/cassandra_agent_config.go#L138)), the invalid name causes another error when the operator tries to create the `CassandraDatacenter`.

We should either reject invalid DC names in the validation webhook, or sanitize them and maybe use the original value as the `datacenterName`.

olim7t:
**What this PR does**:
1\) Handle the case where a DC's `metadata.name` cannot be used directly as a Kubernetes resource name:
```yaml
datacenters:
   - metadata:
      name: Data_Center_1
```
I changed the validation webhook to reject this. I think this is the best solution because we already provide the `datacenterName` override for cases where the CQL name couldn't be used here. We could sanitize `metadata.name` when it's not valid, but that would bring even more confusion IMO.

2\) Fix the naming of some secondary resources (ConfigMaps, etc.) that were incorrectly using the unsanitized override. For example, given:
```yaml
datacenters:
  - metadata:
    name: dc1
  datacenterName: Data_Center_1
```
The operator would generate resource names such as `...-Data_Center_1-metrics-agent-config` which is obviously wrong.

But this gets tricky because there are other places where we were using a sanitized version of the CQL name, which is another way to produce a valid name:
```go
return cassdcapi.CleanupForKubernetes(dc.Spec.ClusterName + "-" + dc.DatacenterName() + "-reaper")
```
(which would produce `...-data-center-1-reaper`)
I've changed those as well, but maybe I was too heavy-handed. I personally think that we should use the "Kubernetes" name (`dc1`) in all resource names, but in practice I think we tend to use the sanitized version of the CQL name.

Finally, there are a number of remaining references to `DatacenterName()`, for example in Medusa or Reaper-related code. I'm not sure if these are correct or not.

Fixes #1141

**Checklist**
- [x] Changes manually tested


olim7t:
I don't think the CI issue is related. I didn't touch anything related to backup/restore, and it's also timing out on main.

emerkle826 (approved):
:+1:

emerkle826:
@olim7t Need to add a Changelog entry too I guess

PR #1262, opened 2024-03-27
Medusa bucket secret label

### LINKED ISSUE #1253: Add a unique label to the Medusa bucket secret
**What is missing?**

The Medusa bucket secret (whether user supplied or operator-generated) should be labelled with a unique label so that it can be picked up by the ReplicatedSecrets controller for replication into individual K8ssandraCluster namespaces.

**Why do we need it?**

Currently, a ReplicatedSecret takes a LabelSelector, which is fine when we are marking it out as belonging to a particular cluster/DC. 

However, the Medusa bucket secret is a 1:many replication into many potential clusters/DCs, not into a single cluster, namespace or DC. 

As such, the design calls for the creation of a new label which will be unique per secret, so that the existing label selectors can be used to pick up the single secret for replication into the K8ssandraCluster's namespace.

**Note:** This ticket has been edited from an original which called for the creation of a name-based selector. This design should not be widely used as we have not evaluated the impacts of the increased label value cardinality on etcd's indexing mechanisms.

**Environment**

* K8ssandra Operator version:

Miles-Garnsey:
**What this PR does**:

Fixes #1253

**Checklist**
- [ ] Changes manually tested


Miles-Garnsey:
A bunch of unit/integration test failures here, these don't happen locally so I'm not sure what the problem is. I'm assuming there's something wrong with main which is getting picked up in the pre-test rebase.

CHANGELOG/CHANGELOG-1.14.md
```diff
@@ -14,6 +14,8 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
 ## unreleased
+* [BUGFIX] [#1253](https://github.com/k8ssandra/k8ssandra-operator/issues/1253) Medusa storage secrets are now labelled with a unique label.
+* [FEATURE] [#1260](https://github.com/k8ssandra/k8ssandra-operator/issues/1260) Update controller-gen to version 0.12.0.
```
burmanm, line 18: 0.14.0

controllers/medusa/medusaconfiguration_controller.go
```diff
@@ -74,15 +80,23 @@ func (r *MedusaConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.
...
-			configuration.Status.SetCondition(medusav1alpha1.ControlStatusSecretAvailable, metav1.ConditionTrue)
+		}
+		configuration.Status.SetCondition(medusav1alpha1.ControlStatusSecretAvailable, metav1.ConditionTrue)
+
+		patch := client.MergeFrom(secret.DeepCopy())
+		if secret.Labels == nil {
+			secret.Labels = make(map[string]string)
+		}
+		secret.Labels[MedusaStorageSecretIdentifierLabel] = utils.HashNameNamespace(secret.Name, secret.Namespace)
+		if err = r.Client.Patch(ctx, secret, patch); err != nil {
```
burmanm, line 98: If the hash already matches the previous one, why would we patch? Also, at least one unit test which guards the implementation?

CHANGELOG/CHANGELOG-1.14.md
Miles-Garnsey: Oop, right you are. Done.

controllers/medusa/medusaconfiguration_controller.go
Miles-Garnsey: I've made this change, however the unit test I've added is now failing since for some reason the patch gets called, fails, and then isn't called again (despite there being a requeue requested).

I'll have to look at this more tomorrow I think.

Miles-Garnsey: Ah, I know what it is. Fixing, this will be ready for another review in  a minute.

controllers/medusa/medusaconfiguration_controller.go
```diff
@@ -74,15 +80,26 @@ func (r *MedusaConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
+		patch := client.MergeFrom(secret.DeepCopy())
+		if secret.Labels == nil {
+			secret.Labels = make(map[string]string)
+		}
+		if secret.Labels[MedusaStorageSecretIdentifierLabel] != utils.HashNameNamespace(secret.Name, secret.Namespace) {
+			secret.Labels[MedusaStorageSecretIdentifierLabel] = utils.HashNameNamespace(secret.Name, secret.Namespace)
+			if err = r.Client.Patch(ctx, secret, patch); err != nil {
+				logger.Error(err, "Failed to patch Medusa Bucket Secret with required label")
+				return ctrl.Result{RequeueAfter: r.DefaultDelay}, err
```
burmanm, line 101: if err != nil, then RequeueAfter shouldn't be set. The err alone will do the correct requeue with a proper backoff strategy.

Miles-Garnsey:
This PR now 

* Also ensures that the system kustomize is not used for e2e tests, instead preferring the one installed in the `./bin` directory. 
* It pins that kustomize back to a 4.x.x version to ensure that we can use namespace replacements as expected.
* It fixes some issues in the makefile where the local bin dir was not being exported to subshells.
* It changes the port used by the docker registry deployed by kind to 5001, which avoids conflicts with other services on recent MacOSes. 
* I've also added patch permissions for secrets, which was necessary to ensure that the update operation could happen.

It should be ready to go now and CI should be 100% green.

Makefile
```diff
@@ -328,8 +328,8 @@ GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint # TODO Add linting to the GHA also
 
 ## Tool Versions
 CERT_MANAGER_VERSION ?= v1.12.2
-KUSTOMIZE_VERSION ?= v5.0.3
-CONTROLLER_TOOLS_VERSION ?= v0.12.0
+KUSTOMIZE_VERSION ?= v4.5.1
```
burmanm, line 331: issue: I have an issue with this, given that 4.5.x is deprecated and 4.5.1 is old version even in 4.x branch and has multiple bugfixes done later. At minimum, this should be 4.5.7.

Users will use 5.x to install our product and now we suddenly don't test it anymore? cass-operator will build structures based on 5.x also. It feels odd if we want to support kustomize installation at all when it doesn't work with the version users would have installed on their machines.

Did you try instead to make it work with 5.x? The current version is 5.4.1, which is quite different from age old 5.0.3.

Miles-Garnsey: It doesn't work with 5.x. I've updated to 4.5.7 as per your request.

Issue #1152, opened 2023-12-27
Remove storageSecretRef requirement for Medusa - support AWS IAM Roles for Service Accounts (IRSA) 


JBOClara:
### Issue Description:

**What is missing?**
Currently, the configuration of Medusa within the K8ssandra Operator requires the use of `storageSecretRef` even though it is marked as optional. This is problematic for setups that utilize AWS IAM Roles for Service Accounts (IRSA) as it necessitates the provision of a `~/.aws/credentials` file to Medusa, which is not ideal or necessary when using IRSA.

**Why do we need it?**
To streamline the integration with AWS services and enhance security, we need the ability to configure Medusa to work with IRSA without the need for a `storageSecretRef`. This would allow Medusa to assume an IAM role and obtain temporary credentials directly from the AWS environment, eliminating the need to manage static credentials.

**Anything else we need to know?**:
This feature would greatly benefit users who are managing their Kubernetes clusters on AWS and are relying on the best practices for authentication by using IAM roles. It would also align with the AWS-recommended way of handling permissions for pods that require access to AWS resources.

Issue #1066, opened 2023-09-27
Do not create Medusa standalone POD when it's not required


adziura-tcloud:
**What is missing?**
Not possible to ignore Medusa standalone deployment creation or configuring existing one (at least I was not able to find it)

**Why do we need it?**
If we do not need/want to do a restore, additional POD is not necessary to be running.

* K8ssandra Operator version:

  Starting from  `1.8.0`

c3-clement:
Today, can restoration be performed without the medusa-standalone pod?

adejanovski:
We still have some calls which go through the standalone pod that should instead go through one of the Cassandra pods.

adziura-tcloud:
@adejanovski , please correct me if I wrong, you want to improve this and it will be possible to omit creation of standalone Medusa deployment in future releases?

another issue, when I upgraded `k8ssandra-operator` from `1.7.0` to `1.15.0` I noticed that Medusa version in Standalone deployment is `0.20.1` while inside the Cassandra POD the container image version was not updated (still `0.13.4`)
is it expected?

PR #1293, opened 2024-04-19
chore: add IRSA usage doc

JBOClara:
**What this PR does**:

Adding doc about IAM role usage.

Complete #1152

**Checklist**
- [ ] Changes manually tested
- [X] Documentation added/updated
- [X] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

### DIFF
--- a/docs/content/en/tasks/backup-restore/_index.md
+++ b/docs/content/en/tasks/backup-restore/_index.md
@@ -155,6 +155,42 @@ spec:
     ...
 ```
 
+## Using IRSA (IAM Roles for Service Accounts) to Attribute Permissions to Medusa (AWS Only)
+
+Simply use `role-based` as the `credentialsType`.
+
+
+```yaml
+apiVersion: k8ssandra.io/v1alpha1
+kind: K8ssandraCluster
+metadata:
+  name: demo
+spec:
+  cassandra:
+    ...
+  medusa:
+    medusaConfigurationRef:
+      name: medusaconfiguration-s3
+    storageProperties:
+      prefix: demo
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
Thanks a lot @JBOClara , much appreciated!

docs/content/en/tasks/backup-restore/_index.md
```diff
@@ -155,6 +155,42 @@ spec:
...
+      prefix: demo
+      credentialsType: role-based
+      storageSecretRef:
+        name: ""
+    ...
+```
+
+To make this work, you must ensure the following steps are completed:
+
+> While Medusa is running in standalone mode, it uses the default service account from the namespace. Make sure this service account has the necessary role annotation.
```
c3-clement, line 184: Is this behavior acceptable ?
That means that the `default` service account needs to be properly annotated and needs to be properly bound with the AWS IAM role. 

Therefore, any pods created without a SA specified will be granted with AWS permissions.
This against the least privilege principle - many of our customers won't accept this.

c3-clement: Following this, just opened this issue https://github.com/k8ssandra/k8ssandra-operator/issues/1294

adejanovski:
FYI folks, [this](https://github.com/k8ssandra/k8ssandra-operator/issues/1066) is in our current dev cycle. We'll probably end up removing this pod altogether since its original use case is no longer valid, and the one we had planned for (bootstrapping a new cluster using a backup) isn't implemented yet.

PR #1296, opened 2024-04-22
Break ReplicatedSecrets infinite loop

### LINKED ISSUE #1295: When CRDs are not upgraded, ReplicatedSecrets may copy infinitely
**What is missing?**

We now use the DropLabels field in the replicatedSecret to ensure that a prefixed secret copied into the same namespace as it's original (and the ReplicatedSecret controlling the replication) does not continue to replicate infinitely.

But if a CRD is not properly upgraded so the DropLabels field exists, or if a user accidentally misconfigures a ReplicatedSecret, then infinite replication can still occur. We should put a safeguard in place to ensure that a secret replicating into it's own namespace will never replicate if doing so would cause an infinite loop.

It is worth noting that we cannot assure this in the case that a ReplicationTarget has been defined with a non zero K8sContext which nonetheless points back to the local cluster. In this case, there is no way for us to tell that the K8sContext is the local one, and infinite replication might still occur.

**Why do we need it?**

We've seen this bug happen once before and it might be dangerous to the cluster given that it continues to create so many copies. It should not happen in ordinary operations, but the extra safety is justified.

**Environment**

* K8ssandra Operator version:

Miles-Garnsey:
**What this PR does**:

Breaks the ReplicatedSecrets infinite loop.

Fixes #1295 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/controllers/replication/secret_controller.go
+++ b/controllers/replication/secret_controller.go
@@ -460,24 +460,13 @@ func getPrefixedSecretName(prefix string, secretName string) string {
 	return fmt.Sprintf("%s%s", prefix, secretName)
 }
 
-func contains(s string, arr []string) bool {
-	for _, i := range arr {
-		if i == s {
-			return true
-		}
-	}
-	return false
-}
-
 func calculateTargetLabels(originalLabels map[string]string, target api.ReplicationTarget) map[string]string {
 	out := originalLabels
 	for k, v := range target.AddLabels {
 		out[k] = v
 	}
-	for key := range out {
-		if contains(key, target.DropLabels) {
-			delete(out, key)
-		}
+	for _, key := range target.DropLabels {
+		delete(out, key)
 	}
 	return out
 }
--- a/controllers/replication/secret_controller.go
+++ b/controllers/replication/secret_controller.go
@@ -461,7 +461,10 @@ func getPrefixedSecretName(prefix string, secretName string) string {
 }
 
 func calculateTargetLabels(originalLabels map[string]string, target api.ReplicationTarget) map[string]string {
-	out := originalLabels
+	out := make(map[string]string)
+	for k, v := range originalLabels {
+		out[k] = v
+	}
 	for k, v := range target.AddLabels {
 		out[k] = v
 	}
--- a/controllers/replication/secret_controller.go
+++ b/controllers/replication/secret_controller.go
@@ -226,7 +226,12 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request) 
 		for i := range secrets {
 			sec := &secrets[i]
 			// If the secret would be created in the same target namespace with the same labels, skip and warn.
-			if wouldBeInfinite(*sec, *rsec, target) {
+			isInf, err := wouldBeInfinite(*sec, *rsec, target)
+			if err != nil {
+				logger.Error(err, "failed to check infinite replication", "Secret", "TargetContext", target)
+				break TargetSecrets
+			}
+			if isInf {
 				logger.Info("warning: secret would be infinite, bailing", "Secret", sec.Name, "TargetContext", target)
 				continue TargetSecrets
 			}
@@ -474,16 +479,17 @@ func calculateTargetLabels(originalLabels map[string]string, target api.Replicat
 	return out
 }
 
-func wouldBeInfinite(origin corev1.Secret, rsec api.ReplicatedSecret, target api.ReplicationTarget) bool {
-	computedLabels := calculateTargetLabels(origin.Labels, target)
-	for k, v := range rsec.Spec.Selector.MatchLabels {
-		if computedLabels[k] != v {
-			return false
+func wouldBeInfinite(origin corev1.Secret, rsec api.ReplicatedSecret, target api.ReplicationTarget) (bool, error) {
+	computedLabels := labels.Set(calculateTargetLabels(origin.Labels, target))
+	selector, err := metav1.LabelSelectorAsSelector(rsec.Spec.Selector)
+	if err != nil {
+		return true, err
+	}
+	if selector.Matches(computedLabels) {
+		if (origin.Namespace == target.Namespace || target.Namespace == "") && target.K8sContextName == "" {
+			// This will still be infinite if the target has a non-empty k8scontext which points back to the origin cluster.
+			return true, nil
 		}
 	}
-	if (origin.Namespace == target.Namespace || target.Namespace == "") && target.K8sContextName == "" {
-		// This will still be infinite if the target has a non-empty k8scontext which points back to the origin cluster.
-		return true
-	}
-	return false
+	return false, nil
 }
--- a/controllers/replication/secret_controller.go
+++ b/controllers/replication/secret_controller.go
@@ -486,7 +486,7 @@ func wouldBeInfinite(origin corev1.Secret, rsec api.ReplicatedSecret, target api
 		return true, err
 	}
 	if selector.Matches(computedLabels) {
-		if (origin.Namespace == target.Namespace || target.Namespace == "") && target.K8sContextName == "" {
+		if (origin.Namespace == target.Namespace || target.Namespace == "") && target.K8sContextName == "" && target.TargetPrefix != "" { // TargetPrefix is included here because the initial .Get in the main reconciliation body will return the original secret if no target prefix is specified, thereby bailing from the secret creation process.
 			// This will still be infinite if the target has a non-empty k8scontext which points back to the origin cluster.
 			return true, nil
 		}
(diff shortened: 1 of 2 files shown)


Miles-Garnsey:
I think this is ready to go, merging ASAP would be good as this both protects against weird cases where secrets replicate infinitely (if your cluster is already badly misconfigured or a user makes a mistake specifying the CRD), but it also solves a subtle issue where labels were being dropped when they shouldn't have been.

Miles-Garnsey:
Right, NOW I think it is working...

controllers/replication/secret_controller.go
```diff
@@ -455,23 +460,30 @@ func getPrefixedSecretName(prefix string, secretName string) string {
...
+	return out
+}
+
+func wouldBeInfinite(origin corev1.Secret, rsec api.ReplicatedSecret, target api.ReplicationTarget) bool {
+	computedLabels := calculateTargetLabels(origin.Labels, target)
+	for k, v := range rsec.Spec.Selector.MatchLabels {
+		if computedLabels[k] != v {
+			return false
 		}
 	}
```
olim7t, line 483: [nitpick] Although I think this is the correct selector logic, I'd rather not reimplement it in our code. Instead of `rsec`, we can pass the `selector` variable that is already computed in the calling method, and then rewrite this as:
```go
if !selector.Matches(labels.Set(computedLabels)) {
  return false
}
```

controllers/replication/secret_controller.go
```diff
@@ -455,23 +460,30 @@ func getPrefixedSecretName(prefix string, secretName string) string {
...
-	return originalLabels
+	if (origin.Namespace == target.Namespace || target.Namespace == "") && target.K8sContextName == "" {
+		// This will still be infinite if the target has a non-empty k8scontext which points back to the origin cluster.
+		return true
+	}
```
olim7t, line 487: [question] Isn't `TargetPrefix != ""` also a precondition for infinite replication?

olim7t (changes requested):
I almost approved directly, but I think the changes I suggest for `wouldBeInfinite` are worth considering.

[thought] As a first-time reader, `Reconcile()` was tough morsel. This is not the object of this PR but we should consider a follow-up ticket to refactor it:
- break down those 250 lines by extracting more functions
- get rid of the loop labels
- do we really need a mutex? (I think the reconciler logic is single-threaded, at least by default)
- I also think the logic to handle overlapping selectors during deletion doesn't consider prefixes (but that's really an edge case)

burmanm:
> do we really need a mutex? (I think the reconciler logic is single-threaded, at least by default)

There's a setting in the reconciler logic to set the amount of threads (to increase scalability for example), so I would ensure all reconcilers are thread safe.

It's just a single line change to get multiple potential reconcilers to run simultaneously.

olim7t:
> There's a setting in the reconciler logic to set the amount of threads (to increase scalability for example)

All right.
We don't override the default or even provide a way to do it right now (that would be `ctrl.NewControllerManagedBy(mgr).WithOptions` in `SetupWithManager` for anyone else following). But if we did I don't see an obvious way to detect that from the reconciler code, so I agree we should be cautious.

controllers/replication/secret_controller.go
```diff
@@ -225,6 +225,11 @@ func (s *SecretSyncController) Reconcile(ctx context.Context, req ctrl.Request)
 		// Iterate all the matching secrets
 		for i := range secrets {
 			sec := &secrets[i]
+			// If the secret would be created in the same target namespace with the same labels, skip and warn.
+			if wouldBeInfinite(*sec, *rsec, target) {
```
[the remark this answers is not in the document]
Miles-Garnsey: I don't like passing pointers when I can pass values. I know that Go makes this hard in some contexts but if the functions can be kept pure it is significantly less cognitive load to reason about them.

controllers/replication/secret_controller.go
Miles-Garnsey: I'd rather avoid this. Delegating to a function here just means you have to context switch more, and this is literally 2 lines of logic.

Miles-Garnsey:
@olim7t thanks for the review, I'm sorry I'm pushing back on most of your comments, I feel like I might be being an ass a little here 😅 They were certainly worth considering though! Let me know if you disagree with my points there, especially if you think I'm wrong about the targetPrefix point.

Miles-Garnsey:
Oh and @olim7t RE your comments about refactoring this function: we could probably use ReconcileObject to make this a bit easier to read. As it is, it is a fair bit of work to get your head around indeed. This is what most of our reconciliation looked like prior to ReconcileObject though, so there's probably a broader refactoring to use that in more places, not just here. 

I'm with Micke when he says that the Mutex needs to stay though.

controllers/replication/secret_controller.go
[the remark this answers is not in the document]
olim7t: OK, I was thinking more in terms of not copying stuff unnecessarily, but I get it.

controllers/replication/secret_controller.go
olim7t: [issue] Actually this is not the correct logic (I missed that in my first review).
A ReplicatedSecret can also use `Selector.MatchExpressions` with operators (`In`, `NotIn`, etc).
`Matches()` will handle that.

controllers/replication/secret_controller.go
olim7t: True, but we call `wouldBeInfinite()` before checking the UID (in `requiresUpdate()`).

As a result, we log the warning `secret would be infinite, bailing` for things that would actually not be infinite, for example:
```yaml
apiVersion: v1
kind: Secret
metadata:
  namespace: k8ssandra-operator
  name: orig-secret
  labels:
    foo: bar
stringData:
---
apiVersion: replication.k8ssandra.io/v1alpha1
kind: ReplicatedSecret
metadata:
  namespace: k8ssandra-operator
  name: replicator
spec:
  selector:
    matchLabels:
      foo: bar
  replicationTargets:
    - namespace: k8ssandra-operator
```
That's not a big deal since that is a fringe case. But also it's a trivial fix.

Miles-Garnsey: > True, but we call wouldBeInfinite() before checking the UID (in requiresUpdate()).

Perhaps I'm misunderstanding you, but I'm pretty sure that UID check will have two different UIDs if there is a target prefix, and you'll then land in this code:

```

	if srcHash, found := source.GetAnnotations()[coreapi.ResourceHashAnnotation]; found {
		// Get dest hash value
		destHash, destFound := dest.GetAnnotations()[coreapi.ResourceHashAnnotation]
		if !destFound {
			return true
		}

		if destSec, valid := dest.(*corev1.Secret); valid {
			hash := utils.DeepHashString(destSec.Data)
			if destHash != hash {
				// Destination data did not match destination hash
				return true
			}
		}

		return srcHash != destHash
```

All of that evaluates true because the names are different (so the hashes are as well, of course). At that point, you'll end up running:

```
logger.Info("Modifying secret in target cluster", "Secret", sec.Name, "TargetContext", target)
				syncSecrets(sec, fetchedSecret, target)
				copiedSecret := fetchedSecret.DeepCopy()
				copiedSecret.Name = getPrefixedSecretName(target.TargetPrefix, sec.Name)
				if err = remoteClient.Update(ctx, copiedSecret); err != nil {
					logger.Error(err, "Failed to sync target secret for matching payloads", "Secret", fetchedSecret.Name, "TargetContext", target)
					break TargetSecrets
				}
```

At that point you're actually in trouble, because you're trying to update a secret which doesn't (or shouldn't) exist - because it would have been infinite.

Basically this error message is there to let you know that you've misspecified your ReplicatedSecret and are risking infinite replication.

controllers/replication/secret_controller.go
Miles-Garnsey: Actually I think you're right, if I look here, the selector is indeed used inside a list call [here](https://github.com/k8ssandra/k8ssandra-operator/blob/feature/finite-replicatedsecrets/controllers/replication/secret_controller.go#L383-L386), so I think there is an argument that `.Matches` indeed will allow the expanded functionality of those operators. I'm not sure if that's used anywhere, but we may as well support it as it is a part of the embedded k8s API.

Thanks for identifying this, I'll make the change.

controllers/replication/secret_controller.go
```diff
@@ -474,16 +479,17 @@ func calculateTargetLabels(originalLabels map[string]string, target api.Replicat
...
-	computedLabels := calculateTargetLabels(origin.Labels, target)
-	for k, v := range rsec.Spec.Selector.MatchLabels {
-		if computedLabels[k] != v {
-			return false
-		}
+func wouldBeInfinite(origin corev1.Secret, rsec api.ReplicatedSecret, target api.ReplicationTarget) (bool, error) {
+	computedLabels := labels.Set(calculateTargetLabels(origin.Labels, target))
+	selector, err := metav1.LabelSelectorAsSelector(rsec.Spec.Selector)
+	if err != nil {
+		return true, err
```
olim7t, line 486: [nitpick] We already have `selector` from line 174 in the main method, we can pass it as an argument to `wouldBeInfinite` instead of rebuilding it here (and then remove the `rsec` argument).

olim7t (approved):

PR #1298, opened 2024-04-24
ObservedGeneration and update block

### LINKED ISSUE #1274: k8ssandra-operator shouldn't modify the cassdc objects on operator upgrade
k8ssandra-operator in spec changes to the cassdc objects which will instantly trigger a rolling restart when the operator is upgraded.
As one operator instance may manage lots of clusters, we don't want upgrades to trigger a rolling restart storm.

```[tasklist]
### Tasks
- [ ] If the Generation of the K8ssandraCluster is the same as the ObservedGeneration we don't update the CassDC objects in Kubernetes (instead, we mark the K8ssandraCluster as requiring refresh). Metadata updates would still be applied.
- [ ] A condition should be placed on the object to indicate that a restart is pending (something as "RestartRequired" condition)
- [ ] Restart can rely on a new annotation `k8ssandra.io/allow-upgrade=once` (using `k8ssandra.io/allow-upgrade=always` would allow permanently to automatically update the K8ssandraCluster on upgrade). This annotation would be placed by either the users or the rolling restart K8ssandraTask. If this annotation is detected, the CassDc Spec update will be unblocked, the "RestartRequired" condition will be set to false and the annotation will be removed at the end of the reconcile if its value is set to `once`
- [ ] We need to allow reconcile to trigger if this annotation is placed on the objects (k8c). This requires an additional predicate for annotation listeners, but then one more filter. This could be useful outside of this feature also.
```

burmanm:
**What this PR does**:
Adds new status field: ObservedGeneration and ensure it is updated at the end of the reconciliation

Fixes #1274 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/constants.go
+++ b/apis/k8ssandra/v1alpha1/constants.go
@@ -67,7 +67,13 @@ const (
 
 	// AutomatedUpdateAnnotation is an annotation that allows the Datacenters to be updated even if no changes were done to the K8ssandraCluster spec
 	AutomatedUpdateAnnotation = "k8ssandra.io/autoupdate-spec"
+
+	AllowUpdateAlways AllowUpdateType = "always"
+	AllowUpdateOnce   AllowUpdateType = "once"
 )
+
+// TODO Use the accepted values from cass-operator's api instead to prevent drift, once Kubernetes dependencies are updated in k8ssandra-operator
+type AllowUpdateType string
 
 var (
 	SystemKeyspaces = []string{"system_traces", "system_distributed", "system_auth"}
--- a/apis/k8ssandra/v1alpha1/constants.go
+++ b/apis/k8ssandra/v1alpha1/constants.go
@@ -64,6 +64,9 @@ const (
 
 	// Annotation to indicate the purpose of a given resource.
 	PurposeAnnotation = "k8ssandra.io/purpose"
+
+	// AutomatedUpdateAnnotation is an annotation that allows the Datacenters to be updated even if no changes were done to the K8ssandraCluster spec
+	AutomatedUpdateAnnotation = "k8ssandra.io/autoupdate-spec"
 )
 
 var (
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -566,5 +566,5 @@ func (kc *K8ssandraCluster) GetClusterIdHash() string {
 }
 
 func (k *K8ssandraCluster) GenerationChanged() bool {
-	return k.Status.ObservedGeneration < k.Generation
+	return k.Status.ObservedGeneration != 0 && k.Status.ObservedGeneration < k.Generation
 }
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -106,9 +106,16 @@ type K8ssandraClusterStatus struct {
 
 	// +kubebuilder:default=None
 	Error string `json:"error,omitempty"`
+
+	// ObservedGeneration is the last observed generation of the K8ssandraCluster.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 }
 
 type K8ssandraClusterConditionType string
+
+const (
+	ClusterRequiresUpdate K8ssandraClusterConditionType = "RequiresUpdate"
+)
 
 type DecommissionProgress string
 
@@ -547,4 +554,8 @@ func (sd *ServerDistribution) IsDse() bool {
 // constraints within k8s
 func (kc *K8ssandraCluster) GetClusterIdHash() string {
 	return utils.HashNameNamespace(kc.Name, kc.Namespace)
+}
+
+func (k *K8ssandraCluster) GenerationChanged() bool {
+	return k.Status.ObservedGeneration < k.Generation
 }
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -566,5 +566,5 @@ func (kc *K8ssandraCluster) GetClusterIdHash() string {
 }
 
 func (k *K8ssandraCluster) GenerationChanged() bool {
-	return k.Status.ObservedGeneration != 0 && k.Status.ObservedGeneration < k.Generation
+	return k.Status.ObservedGeneration < k.Generation
 }
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -518,6 +518,15 @@ func (s *K8ssandraClusterStatus) GetConditionStatus(conditionType K8ssandraClust
 	return corev1.ConditionUnknown
 }
 
+func (s *K8ssandraClusterStatus) SetConditionStatus(conditionType K8ssandraClusterConditionType, status corev1.ConditionStatus) {
+	now := metav1.Now()
+	s.SetCondition(K8ssandraClusterCondition{
+		Type:               conditionType,
+		Status:             status,
+		LastTransitionTime: &now,
+	})
+}
+
 func (s *K8ssandraClusterStatus) SetCondition(condition K8ssandraClusterCondition) {
 	for i, c := range s.Conditions {
 		if c.Type == condition.Type {
--- a/controllers/control/k8ssandratask_controller.go
+++ b/controllers/control/k8ssandratask_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	cassapi "github.com/k8ssandra/cass-operator/apis/control/v1alpha1"
 	k8capi "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/clientcache"
@@ -54,6 +55,8 @@ import (
 
 const (
 	k8ssandraTaskFinalizer = "control.k8ssandra.io/finalizer"
+	InternalTaskAnnotation = "control.k8ssandra.io/internal-command"
+	internalRefreshCommand = "refresh"
 	defaultTTL             = time.Duration(86400) * time.Second
 )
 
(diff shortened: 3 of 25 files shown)


burmanm:
@olim7t

controllers/control/k8ssandratask_controller.go
```diff
@@ -200,6 +210,92 @@ func (r *K8ssandraTaskReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 }
 
+func (r *K8ssandraTaskReconciler) executeRefreshTask(ctx context.Context, kTask *api.K8ssandraTask, kc *k8capi.K8ssandraCluster) (ctrl.Result, error) {
```
olim7t, line 213: [thought] It really bothers me to add an entirely new way to execute a task just for this.
- do we really need a task just to set an annotation on the cluster?
- is there any way we could handle DCs as part of the reconcile loop? That is, if we process a K8ssandraCluster with auto-update=once, then we propagate it to any DC with RequiresUpdate.

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2615,3 +2644,87 @@ func createMultiDcDseCluster(t *testing.T, ctx context.Context, f *framework.Fra
...
+	dcKey := framework.ClusterKey{NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}, K8sContext: f.DataPlaneContexts[1]}
+	require.Eventually(f.DatacenterExists(ctx, dcKey), timeout, interval)
+
+	t.Log("update datacenter status to ready")
+	err = f.SetDatacenterStatusReady(ctx, dcKey)
+	require.NoError(err, "failed to set datacenter status ready")
+
+	verifyClusterReconcileFinished(ctx, t, f, kc)
+
+	// Modify the CassandraDatacenter hash to be some gibberish
```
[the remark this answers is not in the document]
burmanm: Yes, this is just a hack for the testing purpose. The idea is that the new version of k8ssandra-operator would create a different hash (since it wants to create different kind of CassandraDatacenter) which then wouldn't match the old one.

There's no other easy way to cause this in the test without causing Generation to change also (which we don't want in the test).

apis/k8ssandra/v1alpha1/constants.go
```diff
@@ -64,8 +64,17 @@ const (
...
+
+	// AutomatedUpdateAnnotation is an annotation that allows the Datacenters to be updated even if no changes were done to the K8ssandraCluster spec
+	AutomatedUpdateAnnotation = "k8ssandra.io/autoupdate-spec"
+
+	AllowUpdateAlways AllowUpdateType = "always"
+	AllowUpdateOnce   AllowUpdateType = "once"
 )
 
+// TODO Use the accepted values from cass-operator's api instead to prevent drift, once Kubernetes dependencies are updated in k8ssandra-operator
+type AllowUpdateType string
```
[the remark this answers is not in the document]
burmanm: That would blow up the PR, since updating cass-operator is not the issue, but the old dependencies in k8ssandra-operator.

So updating cass-operator requires updating Kubernetes dependencies, controller-runtime dependencies and those are not backwards compatible with the current code, so there would be more changes.

To give a scope of the required changes, here are the changes that cass-operator alone required: https://github.com/k8ssandra/cass-operator/commit/c273e274310b0b007178ab214b2e32774f3cf819

adejanovski:
@burmanm , is this ready for another round of review?

burmanm:
Yes

controllers/k8ssandra/datacenters.go
```diff
@@ -143,9 +151,15 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
-			if !annotations.CompareHashAnnotations(actualDc, desiredDc) {
-				dcLogger.Info("Updating datacenter")
-
+			if !annotations.CompareHashAnnotations(actualDc, desiredDc) && !AllowUpdate(kc) {
+				logger.Info("Datacenter requires an update, but we're not allowed to do it", "CassandraDatacenter", dcKey)
+				// We're not allowed to update, but need to
+				patch := client.MergeFrom(kc.DeepCopy())
+				kc.Status.SetConditionStatus(api.ClusterRequiresUpdate, corev1.ConditionTrue)
+				if err := r.Client.Status().Patch(ctx, kc, patch); err != nil {
+					return result.Error(fmt.Errorf("failed to set %s annotation: %v", api.AutomatedUpdateAnnotation, err)), actualDcs
```
olim7t, line 160: [nitpick] the message is wrong, we're trying to set the condition, not the annotation.

controllers/k8ssandra/datacenters.go
```diff
@@ -233,17 +247,60 @@ func (r *K8ssandraClusterReconciler) reconcileDatacenters(ctx context.Context, k
...
+			if dc.Status.GetConditionStatus("RequiresUpdate") == corev1.ConditionTrue { // TODO Update this to cassdcapi const when cass-operator is updatedß
+				dcsRequiringUpdate = append(dcsRequiringUpdate, dc.DatacenterName())
+			}
+		}
+
+		if len(dcsRequiringUpdate) > 0 {
+			generatedName := fmt.Sprintf("refresh-%d", time.Now().Unix())
+			internalTask := &ktaskapi.K8ssandraTask{}
+			err := r.Get(ctx, types.NamespacedName{Namespace: kc.Namespace, Name: generatedName}, internalTask)
+			// If task wasn't found, create it and if task is still running, requeue
```
olim7t, line 262: [issue] If I'm following correctly, we might requeue for a few reconciliation loops while we wait for the task to finish. However, each loop creates its own `generatedName`, which might be different because the UNIX timestamp part has a resolution of one second. With bad timing, we might end up creating multiple tasks.

olim7t (changes requested):
I still have to test this manually, but publishing this review early because I think the task naming issue needs attention.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -556,3 +572,7 @@ func (kc *K8ssandraCluster) DefaultNumTokens(serverVersion *semver.Version) floa
 func (kc *K8ssandraCluster) GetClusterIdHash() string {
 	return utils.HashNameNamespace(kc.Name, kc.Namespace)
 }
+
+func (k *K8ssandraCluster) GenerationChanged() bool {
+	return k.Status.ObservedGeneration < k.Generation
```
olim7t, line 577: [thought] Since this PR also introduces the new field, maybe we should add this to correctly detect the first operator upgrade:
```suggestion
	return k.Status.ObservedGeneration != 0 && k.Status.ObservedGeneration < k.Generation
```

controllers/k8ssandra/datacenters.go
olim7t: Just to confirm, manual testing confirms that the operator keeps creating refresh tasks in a loop.

```yaml
kind: K8ssandraCluster
metadata:
  namespace: k8ssandra-operator
  name: demo
spec:
  cassandra:
    serverVersion: 4.1.0
    datacenters:
      - metadata:
          name: dc1
        size: 1
        racks:
          - name: rack1
        storageConfig:
          cassandraDataVolumeClaimSpec:
            storageClassName: standard
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 100Mi
```

Simulate a cass-operator upgrade by setting the condition on the `CassandraDatacenter` manually:
```zsh
# Find out the index of the condition in the array. If this is different than 10,
# you need to change it in the next command
kubectl get CassandraDatacenter dc1 -n k8ssandra-operator -o json \
  | jq '.status.conditions | map(.type == "RequiresUpdate") | index(true)'

kubectl patch CassandraDatacenter dc1 -n k8ssandra-operator \
  --subresource status \
  --type='json' \
  -p='[{"op": "replace", "path": "/status/conditions/10/status", "value": "True"}]'
```

Set the annotation on the cluster:
```zsh
kubectl annotate K8ssandraCluster demo -nk8ssandra-operator \
  k8ssandra.io/autoupdate-spec=once
```

Expected: a single `K8ssandraTask` is created.
Actual: a new one is created every ~20s.

olim7t (approved):
I pushed commits to fix a few remaining issues. This is good to go as far as k8ssandra-operator is concerned.

However I'm noticing an issue with the refresh `CassandraTask`: it doesn't seem to ever be marked as finished (and as a consequence neither does the `K8ssandraTask`). It doesn't impact the new behavior we're introducing here, the conditions and annotations are correctly reset.

I'll test on cass-operator alone and file an issue if necessary.

olim7t:
> However I'm noticing an issue with the refresh `CassandraTask`: it doesn't seem to ever be marked as finished (and as a consequence neither does the `K8ssandraTask`).

OK, this might be a side-effect of my testing method: I was artificially setting the `RequiresUpdate=true` condition on the `CassandraDatacenter` to simulate a cass-operator upgrade, but refreshing the DC did not produce any actual changes to the stateful set. It looks like something in the way the refresh is implemented causes the task to not be marked as completed in that case. This is not something that will happen in real life, so I think we can safely ignore it.

adejanovski:
Given we found some bugs (or side effects) in cass-operator's implementation of this feature, I think we should first solve those and re-evaluate how it's done in this PR.
Also, and that's probably something we should have done from the beginning, I think we should make this an opt in feature rather than the default behavior.

adejanovski:
@burmanm, as discussed together, could you add some docs before we merge this? We'd need to explain users what the new behavior looks like and how to unblock a spec update on operator upgrade.
Thanks!

docs/content/en/install/upgrade/_index.md
```diff
@@ -0,0 +1,29 @@
...
+no_list: true
+weight: 2
+description: "Notes on upgrading existing installation"
+---
+
+Upgrading the operators is usually a straight-forward operation based on the standard installation method's upgrade procedure. In certain cases however, updates will require certain manual processing to avoid disruptions to the running clusters.
+
+## Updates after operator upgrade to running Cassandra clusters
+
+Sometimes the updates to operators might bring new features or improvements to existing running Cassandra clusters. However, starting from release 1.18 we will no longer update them automatically when the operators are upgraded to prevent a rolling restart on inconvenient time. If there are changes to be applied after upgrading, the ``K8ssandraCluster`` instances are marked with a Status Condition ``RequiresUpdate`` set to True.
```
rzvoncek, line 13: `on inconvenient time` -> `at inconvenient time`

rzvoncek (approved):
Reviewed the docs update (`docs/content/en/install/upgrade/_index.md`).
There's one small improvement I found, but I couldn't get GH to let me propose a commit.

docs/content/en/install/upgrade/_index.md
burmanm: I think the correct form is "at an inconvenient time" ;)

olim7t: "at an inconvenient time" sounds like the best wording to me.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
burmanm: Turns out, this causes a bug. ObservedGeneration is 0 in new clusters also and that's perfectly valid - so we have to allow update in that case, otherwise we'll never update the CassandraDatacenters. I think we'll just have to accept that older clusters won't benefit from restart protection the first time.

PR #1304, opened 2024-05-02
Remove the medusa standalone pod

### LINKED ISSUE #1066: Do not create Medusa standalone POD when it's not required

adejanovski:
**What this PR does**:
Remove the medusa standalone pod which serves no real purpose today and can be traded for the medusa containers running in the Cassandra pods

Fixes #1066

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/medusa/controllers_test.go
+++ b/controllers/medusa/controllers_test.go
@@ -162,7 +162,15 @@ func setupMedusaRestoreJobTestEnv(t *testing.T, ctx context.Context) *testutils.
 		}
 
 		for _, env := range testEnv.GetDataPlaneEnvTests() {
-			dataPlaneMgr, err := ctrl.NewManager(env.Config, ctrl.Options{Scheme: scheme.Scheme})
+			dataPlaneMgr, err := ctrl.NewManager(
+				env.Config,
+				ctrl.Options{
+					Scheme:  scheme.Scheme,
+					Host:    env.WebhookInstallOptions.LocalServingHost,
+					Port:    env.WebhookInstallOptions.LocalServingPort,
+					CertDir: env.WebhookInstallOptions.LocalServingCertDir,
+				},
+			)
 			if err != nil {
 				return err
 			}
@@ -173,6 +181,7 @@ func setupMedusaRestoreJobTestEnv(t *testing.T, ctx context.Context) *testutils.
 				Scheme:           scheme.Scheme,
 				ClientFactory:    medusaRestoreClientFactory,
 			}).SetupWithManager(dataPlaneMgr)
+			secretswebhook.SetupSecretsInjectorWebhook(dataPlaneMgr)
 			if err != nil {
 				return err
 			}
--- a/controllers/medusa/medusarestorejob_controller.go
+++ b/controllers/medusa/medusarestorejob_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/shared"
 	"net"
 	"time"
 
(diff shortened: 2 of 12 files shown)


controllers/medusa/controllers_test.go
```diff
@@ -173,6 +181,7 @@ func setupMedusaRestoreJobTestEnv(t *testing.T, ctx context.Context) *testutils.
 				Scheme:           scheme.Scheme,
 				ClientFactory:    medusaRestoreClientFactory,
 			}).SetupWithManager(dataPlaneMgr)
+			secretswebhook.SetupSecretsInjectorWebhook(dataPlaneMgr)
```
Miles-Garnsey, line 184: Question: I'm not sure what the changes in this file are doing? They don't look related to the purpose of this PR...

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -262,26 +263,36 @@ func (r *MedusaRestoreJobReconciler) podTemplateSpecUpdateComplete(ctx context.C
...
-	medusaRestoreMapping, err := restoreHostMap.ToMedusaRestoreMapping()
-	if err != nil {
-		return nil, err
-	}
+	for _, pod := range pods {
+		addr := net.JoinHostPort(pod.Status.PodIP, fmt.Sprint(shared.BackupSidecarPort))
+		if medusaClient, err := r.ClientFactory.NewClient(ctx, addr); err != nil {
+			request.Log.Error(err, "Failed to create Medusa client", "address", addr)
+		} else {
+			if err != nil {
```
Miles-Garnsey, line 279: Issue: This logic doesn't appear to make very much sense. You're checking for an error, then logging the error, then checking for the error again and returning it?? What am I missing here?

Miles-Garnsey: Thought: this all really needs a refactor too, e.g. why is this logger sitting on the request struct not the reconciler struct?

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -262,26 +263,36 @@ func (r *MedusaRestoreJobReconciler) podTemplateSpecUpdateComplete(ctx context.C
...
+		request.Log.Error(err, "Failed to get datacenter pods")
 		return nil, err
 	}
 
-	medusaRestoreMapping, err := restoreHostMap.ToMedusaRestoreMapping()
-	if err != nil {
-		return nil, err
-	}
+	for _, pod := range pods {
+		addr := net.JoinHostPort(pod.Status.PodIP, fmt.Sprint(shared.BackupSidecarPort))
```
Miles-Garnsey, line 275: Issue: I'm not sure how this is supposed to work, on the one hand the spec in the ticket calls for there to be logic which sequentially tries different pods until it finds one to whose Medusa container it can connect. 

But that isn't what is happening here. If I'm not mistaken, you're going to error out if you can't connect to the first pod you try. Moreover, if the ordering of the pods is deterministic, then every time you hit this loop you'll obtain the same error, and you'll never progress to trying other sidecar containers in other pods which may be available.

If you have to do it this way (rather than having some real retry logic) at least consider randomising the order of the pods in this slice so that you're trying more than just the first one.

Miles-Garnsey (changes requested):
I've left a few comments where I'm unclear how the logic is supposed to work. I think there are a few problems in there, but I'd like to be mistaken.

More generally, is there a test for this functionality? I can see a lot of stuff removed from the tests, but a straight forward (ideally e2e) test that we can indeed retrieve a restore mapping would be nice to see.

controllers/medusa/controllers_test.go
adejanovski: It's necessary because this envtest now needs to create cassandra pods and get some mock interactions, whereas before it was using the standalone pod which we were faking.
We can't create pods in the envtest if we don't have the webhook present.

Miles-Garnsey: Because they need secrets mounted I presume and those secrets come from the webhook? Roger, sounds fine to me.

controllers/medusa/medusarestorejob_controller.go
adejanovski: The first check of `err != nil` in the `else` clause needs to be removed for this to work.
Not sure how it ended up there, good catch 👍

controllers/medusa/medusarestorejob_controller.go
adejanovski: changed the logger and the method signature 👍

adejanovski (commented):
Done with the required changes, could you do another round @Miles-Garnsey ?

controllers/medusa/medusarestorejob_controller.go
Miles-Garnsey: Sounds about right. Suggestion: I think you should have a test here and have a mock where the first pod fails and the logic has to move on to another pod. Otherwise we aren't checking that this actually meets the specification.

controllers/medusa/medusarestorejob_controller.go
```diff
@@ -262,26 +264,33 @@ func (r *MedusaRestoreJobReconciler) podTemplateSpecUpdateComplete(ctx context.C
...
+		if medusaClient, err := r.ClientFactory.NewClient(ctx, addr); err != nil {
+			logger.Error(err, "Failed to create Medusa client", "address", addr)
+		} else {
+			restoreHostMap, err := medusa.GetHostMap(request.Datacenter, *request.RestoreJob, medusaClient, ctx)
```
Miles-Garnsey, line 280: Probably another issue here. If GetHostMap fails due to network issues, you surely want to continue onto the next pod again don't you? Right now this returns...

Miles-Garnsey (changes requested):
A few more issues here, I think you might want to test more thoroughly, and also think more through when you want to return and when you want to move to the next pod.

Miles-Garnsey (approved):
We're doing some refactoring later, so I'm approving this for now.

PR #1341, opened 2024-06-04
Fix HCD support and add e2e test

adejanovski:
**What this PR does**:
Fixes hcd support by adding the jvm options mapping rules for it. None of the jvm options were mapped at all for HCD before.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
+++ b/apis/k8ssandra/v1alpha1/cassandraconfig_types.go
@@ -51,14 +51,14 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -Xms.
 	// +optional
-	InitialHeapSize *resource.Quantity `json:"heap_initial_size,omitempty" cass-config:"^3.11.x:jvm-options/initial_heap_size;>=4.x,dse@>=6.8.x:jvm-server-options/initial_heap_size"`
+	InitialHeapSize *resource.Quantity `json:"heap_initial_size,omitempty" cass-config:"^3.11.x:jvm-options/initial_heap_size;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/initial_heap_size"`
 
 	// Disabled by default.
 	// Cass Config Builder: supported for Cassandra 3.11 in jvm.options.
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -Xmx.
 	// +optional
-	MaxHeapSize *resource.Quantity `json:"heap_max_size,omitempty" cass-config:"^3.11.x:jvm-options/max_heap_size;>=4.x,dse@>=6.8.x:jvm-server-options/max_heap_size"`
+	MaxHeapSize *resource.Quantity `json:"heap_max_size,omitempty" cass-config:"^3.11.x:jvm-options/max_heap_size;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/max_heap_size"`
 
 	// GENERAL VM OPTIONS
 
@@ -68,7 +68,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -ea.
 	// +optional
-	EnableAssertions *bool `json:"vm_enable_assertions,omitempty" cass-config:"^3.11.x:jvm-options/enable_assertions;>=4.x,dse@>=6.8.x:jvm-server-options/enable_assertions"`
+	EnableAssertions *bool `json:"vm_enable_assertions,omitempty" cass-config:"^3.11.x:jvm-options/enable_assertions;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/enable_assertions"`
 
 	// Enable thread priorities.
 	// Enabled by default.
@@ -76,7 +76,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -XX:+UseThreadPriorities.
 	// +optional
-	EnableThreadPriorities *bool `json:"vm_enable_thread_priorities,omitempty" cass-config:"^3.11.x:jvm-options/use_thread_priorities;>=4.x,dse@>=6.8.x:jvm-server-options/use_thread_priorities"`
+	EnableThreadPriorities *bool `json:"vm_enable_thread_priorities,omitempty" cass-config:"^3.11.x:jvm-options/use_thread_priorities;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/use_thread_priorities"`
 
 	// Enable lowering thread priority without being root on linux.
 	// See CASSANDRA-1181 for details.
@@ -91,7 +91,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -XX:+HeapDumpOnOutOfMemoryError.
 	// +optional
-	HeapDumpOnOutOfMemoryError *bool `json:"vm_heap_dump_on_out_of_memory_error,omitempty" cass-config:"^3.11.x:jvm-options/heap_dump_on_out_of_memory_error;>=4.x,dse@>=6.8.x:jvm-server-options/heap_dump_on_out_of_memory_error"`
+	HeapDumpOnOutOfMemoryError *bool `json:"vm_heap_dump_on_out_of_memory_error,omitempty" cass-config:"^3.11.x:jvm-options/heap_dump_on_out_of_memory_error;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/heap_dump_on_out_of_memory_error"`
 
 	// Per-thread stack size.
 	// Defaults to 256Ki.
@@ -99,7 +99,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -Xss.
 	// +optional
-	PerThreadStackSize *resource.Quantity `json:"vm_per_thread_stack_size,omitempty" cass-config:"^3.11.x:jvm-options/per_thread_stack_size;>=4.x,dse@>=6.8.x:jvm-server-options/per_thread_stack_size"`
+	PerThreadStackSize *resource.Quantity `json:"vm_per_thread_stack_size,omitempty" cass-config:"^3.11.x:jvm-options/per_thread_stack_size;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/per_thread_stack_size"`
 
 	// The size of interned string table. Larger sizes are beneficial to gossip.
 	// Defaults to 1000003.
@@ -107,7 +107,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -XX:StringTableSize.
 	// +optional
-	StringTableSize *resource.Quantity `json:"vm_string_table_size,omitempty" cass-config:"^3.11.x:jvm-options/string_table_size;>=4.x,dse@>=6.8.x:jvm-server-options/string_table_size"`
+	StringTableSize *resource.Quantity `json:"vm_string_table_size,omitempty" cass-config:"^3.11.x:jvm-options/string_table_size;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/string_table_size"`
 
 	// Ensure all memory is faulted and zeroed on startup.
 	// Enabled by default.
@@ -115,7 +115,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -XX:+AlwaysPreTouch.
 	// +optional
-	AlwaysPreTouch *bool `json:"vm_always_pre_touch,omitempty" cass-config:"^3.11.x:jvm-options/always_pre_touch;>=4.x,dse@>=6.8.x:jvm-server-options/always_pre_touch"`
+	AlwaysPreTouch *bool `json:"vm_always_pre_touch,omitempty" cass-config:"^3.11.x:jvm-options/always_pre_touch;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/always_pre_touch"`
 
 	// Disable biased locking to avoid biased lock revocation pauses.
 	// Disabled by default.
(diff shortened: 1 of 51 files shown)


test/e2e/dse_test.go
```diff
@@ -75,6 +75,25 @@ func createSingleDseDatacenterCluster(t *testing.T, ctx context.Context, namespa
...
+	dcKey := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
+	checkDatacenterReady(t, ctx, dcKey, f)
+	checkDatacenterHasHeapSizeSet(t, ctx, dcKey, f)
+	assertCassandraDatacenterK8cStatusReady(ctx, t, f, kcKey, dcKey.Name)
+	dcPrefix := DcPrefix(t, f, dcKey)
+
+	t.Log("Check that we can communicate through CQL with DSE")
+	_, err = f.ExecuteCql(ctx, f.DataPlaneContexts[0], namespace, kc.SanitizedName(), dcPrefix+"-default-sts-0",
+		"SELECT * FROM system.local")
+	require.NoError(t, err, "failed to execute CQL query against DSE")
```
Miles-Garnsey, line 94: You should update the language here to HCD, not DSE.

test/framework/e2e_framework.go
```diff
@@ -56,7 +56,7 @@ var (
 	nodeToolStatusDN = regexp.MustCompile(`DN\s\s`)
 )
 
-func NewE2eFramework(t *testing.T, kubeconfigFile string, useDse bool, controlPlane string, dataPlanes ...string) (*E2eFramework, error) {
+func NewE2eFramework(t *testing.T, kubeconfigFile string, useDse bool, useHcd bool, controlPlane string, dataPlanes ...string) (*E2eFramework, error) {
```
Miles-Garnsey, line 59: issue: This isn't ideal, I'd like to see serverType used directly here since setting both useHcd and useDse shouldn't be set at the same time.

test/framework/e2e_framework.go
```diff
@@ -68,6 +68,10 @@ func NewE2eFramework(t *testing.T, kubeconfigFile string, useDse bool, controlPl
 		cqlshBinLocation = "/opt/dse/bin/cqlsh"
 	}
 
+	if useHcd {
```
Miles-Garnsey, line 71: Issue: I'd rather see this value string interpolated from a serverType field given the expanding number of server types we are supporting.

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -51,14 +51,14 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm-server.options.
 	// Corresponds to: -Xms.
 	// +optional
-	InitialHeapSize *resource.Quantity `json:"heap_initial_size,omitempty" cass-config:"^3.11.x:jvm-options/initial_heap_size;>=4.x,dse@>=6.8.x:jvm-server-options/initial_heap_size"`
+	InitialHeapSize *resource.Quantity `json:"heap_initial_size,omitempty" cass-config:"^3.11.x:jvm-options/initial_heap_size;>=4.x,dse@>=6.8.x,hcd@>=1.x.x:jvm-server-options/initial_heap_size"`
```
Miles-Garnsey, line 54: Question: I haven't gone through this line by line, I assume you've just done a regex find and replace and ensured that anything which is HCD related uses `jvm-server-options` in the same way that DSE does?

Miles-Garnsey (changes requested):
A couple of very minor issues raised here, it took me quite some time to wrap my head around what was happening. 

One big **issue**: I don't think this is finished yet, since I suspect (correct me if I'm wrong) that the new HCD server type will likely no longer use dse.yaml, and instead will have some hcd.yaml - potentially with new options. 

So I think to make this feature complete we will need to add a new HCD.yaml [here](https://github.com/k8ssandra/k8ssandra-operator/blob/fix-hcd-jvm-options/apis/k8ssandra/v1alpha1/cassandraconfig_types.go#L37) to ensure that those options can be configured.

Whether that happens in this PR or (for expediency) in a later one, I'll leave up to you.


I'm still manually testing this too, may have more feedback after that.

Miles-Garnsey:
Manual testing results:

Running this on main using the local e2e test `CreateSingleDatacenterCluster` and swapping out the server type/version/image block for:

```
serverVersion: 1.0.0
serverImage: datastax/hcd:1.0.0-early-preview
serverType: hcd
```

I see that:

The cassDC lacks fields for heap size:

```
config:
    cassandra-env-sh:
      additional-jvm-opts:
      - -Dcassandra.allow_alter_rf_during_range_movement=true
      - -Dcassandra.system_distributed_replication=dc1:3
      - -Dcassandra.jmx.authorizer=org.apache.cassandra.auth.jmx.AuthorizationProxy
      - -Djava.security.auth.login.config=$CASSANDRA_HOME/conf/cassandra-jaas.config
      - -Dcassandra.jmx.remote.login.config=CassandraLogin
      - -Dcom.sun.management.jmxremote.authenticate=true
      - -Dfoo=true
      - -Dbar
    cassandra-yaml:
      authenticator: PasswordAuthenticator
      authorizer: CassandraAuthorizer
      commitlog_directory: /var/lib/cassandra/commitlog2
      data_file_directories:
      - /var/lib/cassandra/data
      - /var/lib/extra/data
      num_tokens: 16
      role_manager: CassandraRoleManager
```

When I look at the jvm-server-options materialised into the image, I see this:

```
cat /opt/hcd/resources/cassandra/conf/jvm-server.options
-ea
-da:net.openhft...
-XX:+UseThreadPriorities
-XX:+HeapDumpOnOutOfMemoryError
-Xss256k
-XX:+AlwaysPreTouch
-XX:+UseTLAB
-XX:+ResizeTLAB
-XX:+UseNUMA
-XX:+PerfDisableSharedMem
-Djava.net.preferIPv4Stack=true
```

However, if I leave this as a Cassandra image with `serverVersion: 4.0.6`, I see:

CassDC:

```
<omitting the ones that existed in the HCD version>
jvm-server-options:
      agent_lib_jdwp: false
      always_pre_touch: false
      cassandra_available_processors: 2
      cassandra_disable_auth_caches_remote_configuration: false
      cassandra_force_default_indexing_page_size: false
      cassandra_ring_delay_ms: 100
      cassandra_write_survey: false
      enable_assertions: false
      flight_recorder: false
      heap_dump_on_out_of_memory_error: false
      initial_heap_size: 536870912
      java_net_prefer_ipv4_stack: true
      jmx-connection-type: local-no-auth
      jmx-port: 7199
      jmx-remote-ssl: false
      jmx-remote-ssl-opts: abc
      max_heap_size: 536870912
      per_thread_stack_size: 262144
      perf_disable_shared_mem: false
      resize_tlb: false
      string_table_size: 1048576
      unlock_commercial_features: false
      use-biased-locking: false
      use_thread_priorities: false
      use_tlb: false
    jvm11-server-options:
      conc_gc_threads: 2
      g1r_set_updating_pause_time_percent: 5
      garbage_collector: G1GC
      initiating_heap_occupancy_percent: 75
      max_gc_pause_millis: 500
      parallel_gc_threads: 2
```

So it does make its way into the cassDC, if I then run `cat /opt/cassandra/conf/jvm-server.options` inside a container I get:

```
-XX:+UnlockDiagnosticVMOptions

-Dcassandra.available_processors=2
-Dcassandra.disable_auth_caches_remote_configuration=false
-Dcassandra.force_default_indexing_page_size=false
-Dcassandra.join_ring=true
-Dcassandra.load_ring_state=true
-Dcassandra.ring_delay_ms=100
-Dcassandra.write_survey=false
-XX:+DebugNonSafepoints
-XX:GuaranteedSafepointInterval=300000
-Xms536870912
-Dio.netty.eventLoop.maxPendingTasks=65536
-Djava.net.preferIPv4Stack=true
-Djdk.nio.maxCachedBufferSize=1048576
-Xmx536870912 <THIS ONE!)>
-Dsun.nio.PageAlignDirectMemory=true
-Xss262144 <THIS ONE!)>
-XX:+PreserveFramePointer
-Dcassandra.printHeapHistogramOnOutOfMemoryError=false
-XX:-RestrictContended
-XX:StringTableSize=1048576
-XX:+UseNUMA


-Dcom.sun.management.jmxremote.authenticate=false
-Dcassandra.jmx.local.port=7199

# The newline in the end of file is intentional
```

And similarly `cat /opt/cassandra/conf/jvm.options`

```
...
-Xss256k
...
```

I'm not sure that we want -Xss set on both files, but I assume that jvm11-server.options takes precedence there. Overall, I think this is a bit of a mess, because things which are defined in the `jvm-server-options` seem to land in `jvm11-server.options`. I recall that this was because of backwards compatibility concerns, however I wonder if this situation could e improved in the interests of maintainability.

Running the same test with the HCD blocks on this branch I see:

The cassDC has 

```
jvm-server-options:
        agent_lib_jdwp: false
        always_pre_touch: false
        cassandra_available_processors: 2
        cassandra_disable_auth_caches_remote_configuration: false
        cassandra_force_default_indexing_page_size: false
        cassandra_ring_delay_ms: 100
        cassandra_write_survey: false
        enable_assertions: false
        flight_recorder: false
        heap_dump_on_out_of_memory_error: false
        initial_heap_size: 536870912
        java_net_prefer_ipv4_stack: true
        jmx-connection-type: local-no-auth
        jmx-port: 7199
        jmx-remote-ssl: false
        jmx-remote-ssl-opts: abc
        max_heap_size: 536870912
        per_thread_stack_size: 262144
        perf_disable_shared_mem: false
        resize_tlb: false
        string_table_size: 1048576
        unlock_commercial_features: false
        use-biased-locking: false
        use_thread_priorities: false
        use_tlb: false
      jvm8-server-options:
        conc_gc_threads: 2
        g1r_set_updating_pause_time_percent: 5
        garbage_collector: G1GC
        initiating_heap_occupancy_percent: 75
        max_gc_pause_millis: 500
        parallel_gc_threads: 2
```

So that looks a bit odd, because we're landing some JVM options in the jvm8-server-options. Validating which java version is running I get:

```
java -version
openjdk version "11.0.23" 2024-04-16
```

**So I think this is wrong, because I validated the Java version on Cassandra 4 as well and I'm pretty sure that was 11 too - so these should match AFAICT.**

Checking the underlying files, there is actually nothing in the directory `/opt/hcd/conf`, it doesn't exist. Instead, it seems that the conf files may be landed in either `/opt/hcd/resources/cassandra/conf/` or `/opt/hcd/resources/hcd/conf/`.

Examining those files:

```
cat /opt/hcd/resources/hcd/conf/jvm-server.options | grep Xmx
# If that's the case, uncomment the -Xmx and Xms options below to override the
# It is recommended to set min (-Xms) and max (-Xmx) heap sizes to
#-Xmx4G
```

This seems unset.

```
cat /opt/hcd/resources/cassandra/conf/jvm11-server.options | grep Xmx
```

Also unset.

```
cat /opt/hcd/resources/cassandra/conf/jvm-server.options | grep Xmx
-Xmx536870912
```

^^ That looks right, but with so many config files I can't tell where this should come from, so I looked at:

```
find / -name cassandra-env.sh
find: ‘/proc/tty/driver’: Permission denied
find: ‘/root’: Permission denied
/opt/hcd/resources/cassandra/conf/cassandra-env.sh
find: ‘/etc/ssl/private’: Permission denied
find: ‘/var/cache/ldconfig’: Permission denied
find: ‘/var/cache/apt/archives/partial’: Permission denied
/config/cassandra-env.sh
```

So it seems that there are actually about 3 different config directories here. So I ps'd the process itself to try to determine what was going on. I do see `-Xmx536870912 -Xss262144 -Xms536870912` in the startup options for the JVM there, so this does appear to be working. 

I'd like to challenge a few things though that I don't really understand:

1. Are we sure this should be in the jvm8-server-options in the cassDC? This doesn't match Cassandra's behaviour and isn't in accordance with the fact that this is a JVM11 container.
2. I don't fully understand where the `/config` directory is coming from. I've had a look at cass-operator and config-builder, but I can't see any PRs there which would allow things to map to these new HCD config directories. Maybe this is in an entrypoint somewhere?
3. This whole structure is really quite abstruse, I think we should organise a refactor at some point in the future (potentially we also need to release a new API so we're not bound by backward compatibility) to make it clearer.

Miles-Garnsey:
My conclusion on this is that this PR does work to make HCD run, and it appears that we get the startup arguments into the Cassandra process that we want.

However, given the wholesale confusion as to which configs are being picked up and where they come from, I think **ISSUE**: please add an e2e test which runs a `ps` command inside the container and (with some non-default heap sizes) ensures that the heap configurations make their way into the Cassandra process. It might be good to add some other jvm-options and such in there as well, if they are commonly used.

Miles-Garnsey (approved):
One question for you to think about in the comments. Otherwise (subject to this bringing in the nodetool related changes to k8ssandra-operator and those then getting CI passing - quite a few caveats there...) approved.

test/e2e/dse_test.go
```diff
@@ -75,6 +76,32 @@ func createSingleDseDatacenterCluster(t *testing.T, ctx context.Context, namespa
...
+	dcKey := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
+	checkDatacenterReady(t, ctx, dcKey, f)
+	checkDatacenterHasHeapSizeSet(t, ctx, dcKey, f)
+	assertCassandraDatacenterK8cStatusReady(ctx, t, f, kcKey, dcKey.Name)
+	dcPrefix := DcPrefix(t, f, dcKey)
+
+	t.Log("Check that we can communicate through CQL with DSE")
+	_, err = f.ExecuteCql(ctx, f.DataPlaneContexts[0], namespace, kc.SanitizedName(), dcPrefix+"-default-sts-0", string(kc.Spec.Cassandra.ServerType),
+		"SELECT * FROM system.local")
+	require.NoError(t, err, "failed to execute CQL query against DSE", err)
```
Miles-Garnsey, line 95: Oh, one more thing - you still need to change DSE references here to HCD...

test/e2e/dse_test.go
```diff
@@ -75,6 +76,32 @@ func createSingleDseDatacenterCluster(t *testing.T, ctx context.Context, namespa
...
+	opts := kubectl.Options{
+		Namespace: namespace,
+		Context:   f.DataPlaneContexts[0],
+	}
+	output, err := kubectl.Exec(opts, dcPrefix+"-default-sts-0", "ps", "-aux")
+	require.NoError(t, err, "failed to execute ps command")
+	assert.Contains(t, output, "java -Dhcd.server_process -Xms536870912 -Xmx536870912", "expected heap size to be set to 512M")
```
Miles-Garnsey, line 102: I'm not sure if this will pass, because the substring `java -Dhcd.server_process -Xms536870912 -Xmx536870912` may not appear continguously in the output, and I'm not sure if that behaviour is always stable. Perhaps assert on each of the substrings independently to avoid flakes?

apis/k8ssandra/v1alpha1/cassandraconfig_types.go
```diff
@@ -200,7 +200,7 @@ type JvmOptions struct {
 	// Cass Config Builder: supported for Cassandra 4.0 in jvm11-server.options.
 	// Corresponds to: -Dio.netty.tryReflectionSetAccessible=true.
 	// +optional
-	NettyTryReflectionSetAccessible *bool `json:"netty_try_reflection_set_accessible,omitempty" cass-config:">=4.x:jvm11-server-options/io_netty_try_reflection_set_accessible;dse@>=6.8.x,hcd@>=1.x.x:jvm8-server-options/io_netty_try_reflection_set_accessible"`
```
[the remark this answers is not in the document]
adejanovski: the heap size isn't affected by this since it ends up in jvm-server.options, and not a jdk specific file.

We won't e2e test all of the possible jvm options, that would be overkill, so I'd rather keep the check for the heap size specifically as it breaks things in a big way.
I reckon we may have bugs in how some of them are processed.

test/e2e/dse_test.go
adejanovski: The command line should be built in a predictable way, as long as we don't change the jvm options used in the manifest.

PR #1358, opened 2024-06-18
Consider DC-level config when validating numToken updates in webhook (fixes #1222)

### LINKED ISSUE #1222: Validating webhook rejects updates if the server version is declared at the DC level
The fix for #1119 parses the Cassandra version during a cluster update: 
```golang
cassVersion, err := semver.NewVersion(oldCluster.Spec.Cassandra.ServerVersion)
```
([source](https://github.com/k8ssandra/k8ssandra-operator/commit/554225264dd328aaf70b55666f71fd5187ba20ad#diff-b45565c4bbd47954036865bbbb5d48b9fb5adba73a07d68945196335fe65c3f0R123))

However, our CRD allows that version to be omitted, if it is declared at the DC level:
```yaml
spec:
  cassandra:
    # no serverVersion here
    datacenters:
      - metadata:
          name: dc1
        serverVersion: 4.1.0
```

If a cluster is created in that manner, and one later tries to update it, the validating webhook rejects the update:
```
DEBUG controller-runtime.webhook.webhooks wrote response {
  "webhook": "/validate-k8ssandra-io-v1alpha1-k8ssandracluster",
  "code": 403,
  "reason": "Invalid Semantic Version",
  [...]
```

The code should consider the DC-level version if necessary. I'm not sure of the exact semantics used by the operator (can both be present? can DCs declare different versions?), so this will require a preliminary analysis.

olim7t:
**What this PR does**:
Instead of only comparing the top-level config (which can fail if fields are missing), compute and compare the numToken of each individual DC, using the inheritance and defaulting rules.
This also covers cases that were not handled before, like config being nil on one side but not the other.

Fixes #1222

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
@@ -18,11 +18,11 @@ package v1alpha1
 
 import (
 	"fmt"
+	"github.com/Masterminds/semver/v3"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/clientcache"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -150,26 +150,8 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) (admission.Warning
 		}
 	}
 
-	oldCassConfig := oldCluster.Spec.Cassandra.DatacenterOptions.CassandraConfig
-	newCassConfig := r.Spec.Cassandra.DatacenterOptions.CassandraConfig
-	if oldCassConfig != nil && newCassConfig != nil {
-		oldNumTokens, oldNumTokensExists := oldCassConfig.CassandraYaml["num_tokens"]
-		newNumTokens, newNumTokensExists := newCassConfig.CassandraYaml["num_tokens"]
-
-		if !oldNumTokensExists {
-			cassVersion, err := semver.NewVersion(oldCluster.Spec.Cassandra.ServerVersion)
-			if err != nil {
-				return nil, err
-			}
-			defaultNumTokens := oldCluster.DefaultNumTokens(cassVersion)
-			if newNumTokensExists && newNumTokens.(float64) != defaultNumTokens {
-				return nil, ErrNumTokens
-			}
-		} else {
-			if oldNumTokens != newNumTokens {
-				return nil, ErrNumTokens
-			}
-		}
+	if err := validateUpdateNumTokens(oldCluster.Spec.Cassandra, r.Spec.Cassandra); err != nil {
+		return nil, err
 	}
 
 	// Verify that the cluster name override was not changed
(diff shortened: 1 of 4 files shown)


apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -192,6 +174,71 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) (admission.Warning
...
+	}
+	newNumTokensPerDc, err := numTokensPerDc(newCassandra)
+	if err != nil {
+		return err
+	}
+
+	for dcName, newNumTokens := range newNumTokensPerDc {
+		oldNumTokens, oldExists := oldNumTokensPerDc[dcName]
+		if oldExists && oldNumTokens != newNumTokens {
+			return ErrNumTokens
```
olim7t, line 193: We could add more details and keep iterating to get an exhaustive list, but I'm not sure if it's worth it.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -192,6 +174,71 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) (admission.Warning
...
+		}
+	}
+	return nil
+}
+
+func numTokensPerDc(cassandra *CassandraClusterTemplate) (map[string]interface{}, error) {
```
burmanm, line 199: question: Why interface{} and not map[string]float64. Or actually just map[string]uint / int, since we don't use floats for anything in reality (it's a YAML/JSON problem) and we should throw an error if we can't map to uint value.

olim7t: Because it's good enough for the task at hand, which is only about comparing values.

We could indeed check that the values are integers, but then why stop at `num_tokens`, and not have the webhook validate every other config value? It's not the approach we have taken so far, so I assume we want to keep things as light as possible here and do more advanced checks later (in this case `computeDcNumTokens()`).

As to float64 for the defaults, it's because that's what unmarshaling into `Unstructured` produces for any numeric value:
```go
	u := new(Unstructured)
	u.UnmarshalJSON([]byte(`{"num_token": 16}`))
	numToken, _ := u.Get("num_token")
	i16 := 16
	f16 := float64(i16)
	fmt.Println("numToken==i16", numToken == i16) // false
	fmt.Println("numToken==f16", numToken == f16) // true
```

burmanm: > As to float64 for the defaults, it's because that's what unmarshaling into Unstructured produces for any numeric value

It is, because JSON & YAML do not accept any other numeric value. Problem itself is more interesting if you compare:

"num_tokens": "16" and "num_tokens": 16. Both are the same value, but they could be unmarshalled differently. We do not force users to use : 16 or : "16" and most users wouldn't even able to tell what the difference is. From your example, we end up with a difference however:

```go
func TestUnmarshalNumbers(t *testing.T) {
	dest := make(map[string]interface{}, 1)
	assert.NoError(t, json.Unmarshal([]byte(`{"num_token": "16"}`), &dest))
	numToken := dest["num_token"]
	i16 := 16
	f16 := float64(i16)

	assert.NotEqual(t, numToken, i16)
	assert.NotEqual(t, numToken, f16)
}
```

Neither compare will work. One needs to parse it first as number (using strconv) and only then compare the resulting int.

olim7t: True, this PR assumes the user will provide a YAML number (as does the current implementation on the main branch btw).
If supporting YAML strings is also a requirement, then it would make sense to parse.

adejanovski:
@olim7t @burmanm, what's the status on this PR?

burmanm (approved):
suggestion: I'm going to approve since current main branch behavior is "wrong" also, but we really shouldn't expect users to have their YAML as numbers always, since we also have situations where we require user to input the number as string instead of number (to avoid string/uint64 -> float64 issue, like we had the bug with weird commit number begin parsed as a number).

PR #1390, opened 2024-08-22
Add service to expose DC nodes in the control plane (fixes #1382)

### LINKED ISSUE #1382: Create a new ClusterIP service in the CP per CassDc
In order to allow Reaper to discover the IPs of all pods in all the managed clusters, which could be hosted in remote dataplanes, we'd need one additional service to be created in the CP for each CassDc.
This service would be equivalent to the additional-seed-service, where k8ssandra-operator explicitly sets the endpoints to the IP addresses of the seed nodes, but we would get the IPs of all pods in the DC instead of the pods for other DCs.
k8ssandra-operator would be responsible for reconciling those endpoints in case of IP changes, and Reaper would use the service URL as contact point for the cluster.
A new watcher must be added on the Endpoints object named `<cluster>-<dc>-all-pods-services` across all dataplanes (and the control plane), then in the `reconcileDatacenters()` method, a call must be made to a newly created method which will reconcile our new service along with the corresponding endpoints.
A new label must be added on the all pods service to be able to filter services in the watcher efficiently.

```[tasklist]
### Definition of Done
- [ ] One headless ClusterIP service gets created in the control plane for each DC
- [ ] this service endpoints contains the IPs of all nodes across the DC
- [ ] k8ssandra-operator explicitly sets the endpoints for these services like it does for additional-seed services
- [ ] The list of endpoints is copied directly from the watched service
- [ ] The **-**-all-pods-service services in cass-operator get an additional label such as `cassandra.datastax.com/all-pods: true` 
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-6) by [Unito](https://www.unito.io)

olim7t:
**What this PR does**:

Fixes #1382

**Checklist**
- [x] Changes manually tested


apis/k8ssandra/v1alpha1/constants.go
```diff
@@ -58,6 +58,11 @@ const (
 	K8ssandraClusterNameLabel      = "k8ssandra.io/cluster-name"
 	K8ssandraClusterNamespaceLabel = "k8ssandra.io/cluster-namespace"
 
+	// TODO this is a workaround because cass-operator does not let us set k8ssandra.io/* labels in
+	//  AdditionalServiceConfig. Remove and replace by the two original constants once this is fixed.
+	TmpK8ssandraClusterNameLabel      = "not.k8ssandra.io/cluster-name"
+	TmpK8ssandraClusterNamespaceLabel = "not.k8ssandra.io/cluster-namespace"
```
olim7t, line 64: This requires a fix in cass-operator.
I will mark the PR ready for review once that is done and we can revert to the original labels.

controllers/k8ssandra/contact_points_service.go
```diff
@@ -0,0 +1,254 @@
...
+				api.ComponentLabel:                 api.ComponentLabelValueCassandra,
+				api.K8ssandraClusterNameLabel:      kc.Name,
+				api.K8ssandraClusterNamespaceLabel: kc.Namespace,
+				api.DatacenterLabel:                dc.Name,
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Type:  corev1.ServiceTypeClusterIP,
+			Ports: ports,
+			// We don't provide a selector since the operator manages the Endpoints directly
```
olim7t, line 153: I'm not sure if this service should be headless (currently it is not), it seems convenient to use the service name the contact point if a component wants to connect to "any node" in the DC.

apis/k8ssandra/v1alpha1/constants.go
olim7t: k8ssandra/cass-operator#689

controllers/k8ssandra/contact_points_service.go
olim7t: Actually, I think the service _must_ be headless if we want to be able to resolve individual pod addresses.

rzvoncek (approved):
I could not get my 2-dc multi-context kind cluster change the IP of a pod, but adding and removing pods worked as intended.

PR #1421, opened 2024-09-30
MC-1296 Deploy Reaper capable of interaction with encrypted mgmt-api

rzvoncek:
**What this PR does**:
This PR changes the reconciliation process of k8ssandra cluster and reaper objects so that it becomes possible for one a single reaper insance to interact with several k8ssandra clusters with their mgmt-api communication encrypted (each using a different certificate).

Fixes 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/reaper/reaper_controller_test.go
+++ b/controllers/reaper/reaper_controller_test.go
@@ -627,18 +627,6 @@ func testCreateReaperWithHttpAuthEnabled(t *testing.T, ctx context.Context, k8sC
 		return errors.IsNotFound(err)
 	}, timeout, interval, "reaper stateful set deletion check failed")
 
-	assert.Eventually(t, func() bool {
-		err = k8sClient.Get(ctx, stsKey, sts)
-		// we'd expect errors.IsNotFound(err) here, except for some reason this is not happening in env tests
-		return err == nil
-	}, timeout, interval, "reaper stateful set deletion check failed")
-
-	assert.Eventually(t, func() bool {
-		err = k8sClient.Get(ctx, secretKey, truststoresSecret)
-		// again, we'd expect errors.IsNotFound(err) ...
-		return err == nil
-	}, timeout, interval, "reaper truststore secret deletion check failed")
-
 	// so we delete stuff manually
 	err = k8sClient.Delete(ctx, sts)
 	require.NoError(t, err)


Miles-Garnsey (changes requested):
@rzvoncek when I run `CreateControlPlaneReaperAndDataCenter` from vs code, I'm seeing the test fail - 

```
--- FAIL: TestOperator (926.60s)
    --- FAIL: TestOperator/CreateControlPlaneReaperAndDataCenter (926.60s)
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/e2e_framework.go:69: Using config file: /Users/milesgarnsey/projects/k8ssandra-operator/build/kind-kubeconfig
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/e2e_framework.go:77: Using control plane: kind-k8ssandra-0
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/e2e_framework.go:94: Using data planes: [kind-k8ssandra-0 kind-k8ssandra-1]
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/reaper_test.go:350: check that Reaper reaper1 in cluster kind-k8ssandra-0 is ready
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:1805: check that datacenter enc-mgmt-dc1 in cluster kind-k8ssandra-0 is ready
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:1805: check that datacenter enc-mgmt-2-dc1 in cluster kind-k8ssandra-0 is ready
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/reaper_test.go:321: Verify Reaper received k8ssandra-cluster secrets
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:1951: 
            	Error Trace:	/Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:1951
            	            				/Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/reaper_test.go:322
            	            				/Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:481
            	Error:      	Received unexpected error:
            	            	secrets "reaper1-truststores" not found
            	Test:       	TestOperator/CreateControlPlaneReaperAndDataCenter
            	Messages:   	failed to get reaper's truststore secret
        /Users/milesgarnsey/projects/k8ssandra-operator/test/e2e/suite_test.go:776: failed to delete namespace: context deadline exceeded
FAIL
FAIL	github.com/k8ssandra/k8ssandra-operator/test/e2e	927.122s
```

Is there something extra I need to do to make this work?

rzvoncek:
This led me to notice the test I added did not actually ran with the CI ... I've added it and it now ran [here](https://github.com/k8ssandra/k8ssandra-operator/actions/runs/11498842640/job/32005589775?pr=1421). It seems to have passed. The PR adds a bunch of stuff, but it seems to be all present in the changed files.

So I'm not sure why this would fail ... when running from an IDE. I always run the e2e tests from shell, with:

```
e2e_test="TestOperator/CreateControlPlaneReaperAndDataCenter"
make single-up
make E2E_TEST="$e2e_test" e2e-test
```

Could you please try that way?

controllers/k8ssandra/reaper.go
```diff
@@ -267,6 +275,22 @@ func getSingleReaperDcName(kc *api.K8ssandraCluster) string {
 	return ""
 }
 
+func usesHttpAuth(kc *api.K8ssandraCluster, actualDc *cassdcapi.CassandraDatacenter) (bool, *cassdcapi.ManagementApiAuthManualConfig) {
```
Miles-Garnsey, line 278: Suggestion: this looks like a getter with some defaulting behaviour. Can I suggest that it should be a method on the `ManagementApiAuth` type.

controllers/k8ssandra/reaper.go
```diff
@@ -290,3 +314,69 @@ func (r *K8ssandraClusterReconciler) addClusterToExternalReaper(
...
 	return result.Continue()
 }
+
+func addManagementApiSecretsToReaper(
+	ctx context.Context,
+	remoteClient client.Client,
+	kc *api.K8ssandraCluster,
+	actualDc *cassdcapi.CassandraDatacenter,
+	logger logr.Logger,
+	authConfig *cassdcapi.ManagementApiAuthManualConfig,
```
Miles-Garnsey, line 324: Suggestion: could you get this from the kc object to avoid the additional parameter? This function has a lot already.

controllers/k8ssandra/reaper.go
```diff
@@ -290,3 +314,69 @@ func (r *K8ssandraClusterReconciler) addClusterToExternalReaper(
...
+) error {
+
+	reaperName := kc.Spec.Reaper.ReaperRef.Name
+
+	reaperNamespace := kc.Spec.Reaper.ReaperRef.Namespace
+	if reaperNamespace == "" {
+		reaperNamespace = kc.Namespace
```
Miles-Garnsey, line 331: Question: Is this the right default? It seems to me like by default Reaper will deploy to the mission control namespace as a static resource. But I kind of prefer this logic because it mirrors what we do elsewhere more closely...

pkg/reaper/secrets.go
```diff
@@ -92,3 +92,7 @@ func secretToEnvVars(secret *corev1.Secret, envUsernameParam, envPasswordParam s
 
 	return &usernameEnvVar, &passwordEnvVar, nil
 }
+
+func GetTruststoresSecretName(reaperName string) string {
```
Miles-Garnsey, line 96: Suggestion: consider abbreviating this to -ts (like we do -ks for keystores) because otherwise the name will be too long for the API server.

controllers/k8ssandra/reaper.go
```diff
@@ -290,3 +314,69 @@ func (r *K8ssandraClusterReconciler) addClusterToExternalReaper(
...
+	tssKey := client.ObjectKey{Namespace: reaperNamespace, Name: tssName}
+
+	tss := &corev1.Secret{}
+	if err := remoteClient.Get(ctx, tssKey, tss); err != nil {
+		logger.Error(err, "failed to get Reaper's truststore secret")
+		return err
+	}
+
+	cs := &corev1.Secret{}
+	csName := authConfig.ClientSecretName + "-ks"
```
Miles-Garnsey, line 344: Issue: It looks like maybe you're re-using the same client secret currently used by cass-operator. Do you think we should maybe issue new secrets for Reaper for the sake of security?

controllers/k8ssandra/reaper.go
```diff
@@ -290,3 +314,69 @@ func (r *K8ssandraClusterReconciler) addClusterToExternalReaper(
...
+	csKey := client.ObjectKey{Namespace: kc.Namespace, Name: csName}
+	if err := remoteClient.Get(ctx, csKey, cs); err != nil {
+		logger.Error(err, "failed to get k8ssandra cluster client secret", "secretName", csName)
+		return err
+	}
+
+	clusterName := cassdcapi.CleanupForKubernetes(actualDc.Spec.ClusterName)
```
Miles-Garnsey, line 351: Issue: I'm not sure this is the right logic - in particular clusterName is not always populated. Is there a function you can re-use in cass operator to figure out the resource names?

controllers/k8ssandra/reaper.go
```diff
@@ -290,3 +314,69 @@ func (r *K8ssandraClusterReconciler) addClusterToExternalReaper(
...
+	clustersKeystore := clusterName + "-keystore.jks"
+
+	// check if the reaper's big secret already has entry for this cluster
+	if tss.Data == nil {
+		tss.Data = make(map[string][]byte)
+	}
+	_, hasTruststore := tss.Data[clustersTruststore]
+	_, hasKeystore := tss.Data[clustersKeystore]
+
+	if hasTruststore && hasKeystore {
```
Miles-Garnsey, line 362: Issue: Do you need to ensure that the content is correct? What if the secrets have rotated?

controllers/reaper/reaper_controller.go
```diff
@@ -484,3 +492,31 @@ func (r *ReaperReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	sKey := types.NamespacedName{Namespace: actualReaper.Namespace, Name: sName}
+	s := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sName,
+			Namespace: actualReaper.Namespace,
+		},
+	}
+	if err := r.Client.Get(ctx, sKey, s); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Creating Reaper's truststore ConfigMap", "ConfigMap", sKey)
```
Miles-Garnsey, line 507: Issue: this should say secret, not configmap.

controllers/reaper/reaper_controller.go
```diff
@@ -484,3 +492,31 @@ func (r *ReaperReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if err = controllerutil.SetControllerReference(actualReaper, s, r.Scheme); err != nil {
```
Miles-Garnsey, line 508: Issue: I think you need `controllerutil.SetOwnerReference` as well.

controllers/reaper/reaper_controller_test.go
```diff
@@ -570,6 +572,72 @@ func testCreateReaperWithLocalStorageType(t *testing.T, ctx context.Context, k8s
...
+	secretKey := types.NamespacedName{Namespace: testNamespace, Name: reaper.GetTruststoresSecretName(r.Name)}
+	truststoresSecret := &corev1.Secret{}
+	require.Eventually(t, func() bool {
+		return k8sClient.Get(ctx, secretKey, truststoresSecret) == nil
+	}, timeout, interval, "truststore secret creation check failed")
+	assert.True(t, truststoresSecret.Data == nil)
+	assert.Equal(t, 0, len(truststoresSecret.Data))
+
+	// In this configuration, we expect Reaper to also have a mount for the http auth secrets
+	assert.Len(t, sts.Spec.Template.Spec.Containers[0].VolumeMounts, 3)
```
Miles-Garnsey, line 603: Issue: why do we need a deepcopy here? Also, I think you might be better off writing something more abstract, as right now any changes to the ordering of the mounts etc. will cause this test to fail.

controllers/reaper/reaper_controller_test.go
```diff
@@ -570,6 +572,72 @@ func testCreateReaperWithLocalStorageType(t *testing.T, ctx context.Context, k8s
...
+
+	assert.Eventually(t, func() bool {
+		err = k8sClient.Get(ctx, stsKey, sts)
+		// we'd expect errors.IsNotFound(err) here, except for some reason this is not happening in env tests
+		return err == nil
+	}, timeout, interval, "reaper stateful set deletion check failed")
+
+	assert.Eventually(t, func() bool {
+		err = k8sClient.Get(ctx, secretKey, truststoresSecret)
+		// again, we'd expect errors.IsNotFound(err) ...
```
Miles-Garnsey, line 630: Issue: can you double check this comment in light of cleanups not happening in envtests. And the one above too...

test/e2e/suite_test.go
```diff
@@ -1933,24 +1934,47 @@ func createKeyspaceAndTable(
 ) {
 	_, err := f.ExecuteCql(ctx, k8sContext, namespace, clusterName, pod, fmt.Sprintf("CREATE KEYSPACE %s WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : %d};", keyspace, replicationFactor))
 	require.NoError(t, err, "failed to create keyspace")
-
-	_, err = f.ExecuteCql(ctx, k8sContext, namespace, clusterName, pod, fmt.Sprintf("CREATE TABLE %s.%s (id int PRIMARY KEY);", keyspace, table))
+	_, err = f.ExecuteCql(ctx, k8sContext, namespace, clusterName, pod, fmt.Sprintf("CREATE TABLE %s.%s (id INT PRIMARY KEY, v TEXT);", keyspace, table))
 	require.NoError(t, err, "failed to create table")
 }
 
+func verifyReaperSecrets(
```
Miles-Garnsey, line 1941: Suggestion: it would be good to check the fingerprints on the certs to make sure they're what you expect.

test/testdata/fixtures/reaper-control-plane/reaper.yaml
```diff
@@ -17,3 +17,5 @@ spec:
     enabled: false
   uiUserSecretRef:
     name: reaper-ui-secret
+  containerImage:
```
Miles-Garnsey, line 20: Question: Do we still need this or have we now done a release? I suggest you put a TODO comment here so that it gets (hopefully) picked up before release.

Miles-Garnsey (changes requested):
A few things:

1. I think you need to create new secrets for the sake of security, not re-use existing client secrets.
2. Unit/integration tests are currently failing.
3. I'm still not able to run `CreateControlPlaneReaperAndDataCenter` successfully. I also checked that other tests work. `AddDcToClusterSameDataplane` definitely works when run this way so it isn't just because we're running from the IDE.

If I'm understanding this correctly there are two sets of changes. One set in the Reaper controller creates mounts and an empty secret. Another set in the k8ssandra cluster controller adds the existing client encryption materials to Reaper's secret. 

I don't fully understand the API on the Reaper side yet, but this looks OK generally to me, just have a look at some of my comments throughout and the three main things I've highlighted above.

controllers/reaper/reaper_controller_test.go
burmanm: I'm not sure what you meant here, but there's no garbage collection in envtests. So no reference deletions happen as there's no such controller running in envtest. So the comment on line 612 is correct.

controllers/reaper/reaper_controller.go
burmanm: You do not, SetControllerReference sets OwnerReference. It just sets the "controller" part in the OwnerReference. 

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/controllerutil.go#L85

burmanm:
> I think you need to create new secrets for the sake of security, not re-use existing client secrets.

This is not the correct way, as mgmt-api supports a single TLS cert (combination of tls.crt/ca.crt/tls.key files, not Java keystores). Thus, if Reaper doesn't use the cass-operator one, then Reaper will not be able to connect to mgmt-api.

controllers/k8ssandra/reaper.go
rzvoncek: I'm sorry but I don't think removing the `authConfig` function parameter leads to simpler `addManagementApiSecretsToReaper` function. 
The reason is that the `authConfig` passed into this function comes from the `usesHttpAuth` function, which picks it from one of the two possible palces in the `K8ssandraCluster` spec. I'd have to

controllers/k8ssandra/reaper.go
rzvoncek: It is indeed very likely we'll be deploying Reaper in the `mission-control` namespace. But I also think we should not be hardcoing the defaults of that namespace anywhere. I think it once bit me already, althought I don't remember where.
So I went for waht we do elsewhere. We seem to be in agreement, so I'll go ahead and resolve this bit.

pkg/reaper/secrets.go
rzvoncek: Pushed a commit doing this.

controllers/k8ssandra/reaper.go
rzvoncek: I agree with Micke - and not only becausae that means I don't have to change anything 😃 
Generating and managing another secret for Reaper <-> mgmt-api seems like a big headache. Taking this ad absurdum, would we then need each client of mgmt-api have its own secret - and mgmt-api accept all of them. That does not sound correct.

controllers/k8ssandra/reaper.go
rzvoncek: The `actualDc.Spec.ClusterName` is a field of `CassandraDatacenterSepc` in cass-operator. It's declared like this:
```
	// +kubebuilder:validation:MinLength=2
	ClusterName string `json:"clusterName"`
```
This looks like the field must be a string of at least two characters. Otherwise, I think, the cass-operator's webhook will prevent creating the DC. In other words, if a `CassandraDatacenter` object shows up, this field must be at least 2 chars long. Please correct me if I'm missing something.

Anyhow, we could also take the K8ssandraCluster object name, or `kc.Spec.Cassandra.ClusterName`. I'd say all these should be set to the same thing, but I don't really know.

Could someone else please chime in?

controllers/k8ssandra/reaper.go
rzvoncek: This is a good catch. The rotation is not currently handled. I've made [MC-1367](https://datastax.jira.com/browse/MC-1367) to add that. 
What other correctnes checks could you think of ?

controllers/reaper/reaper_controller.go
rzvoncek: Pushed a commit doing this.

controllers/reaper/reaper_controller_test.go
rzvoncek: I've removed the `DeepCopy()`. No idea why it was there.

rzvoncek: I've also refactored the access by index. Now it finds the container by name first.

test/testdata/fixtures/reaper-control-plane/reaper.yaml
rzvoncek: We only merged REAP-10 yesterday, we do not have a Reaper release with this just yet.

controllers/k8ssandra/reaper.go
Miles-Garnsey: Understood, apologies I didn't realise it was in cass-operator.

controllers/k8ssandra/reaper.go
Miles-Garnsey: OK, closing as this was only a suggestion

controllers/k8ssandra/reaper.go
Miles-Garnsey: The principle we have followed elsewhere is that each application gets its own set of encryption materials. So one set for Cassandra, one set for management API, one set for Reaper, etc. 

@rzvoncek suggest you check this with Chris, I'm not personally fussy either way.

And @burmanm - can you please elaborate on your perspectives here so we can understand your reasoning.

controllers/reaper/reaper_controller_test.go
Miles-Garnsey: I wasn't talking about 612, I was talking abot 624.

controllers/k8ssandra/reaper.go
Miles-Garnsey: > This is not the correct way, as mgmt-api supports a single TLS cert (combination of tls.crt/ca.crt/tls.key files, not Java keystores). Thus, if Reaper doesn't use the cass-operator one, then Reaper will not be able to connect to mgmt-api.

@burmanm I've found your reasoning in your additional comment further below. I think you've misunderstood what I'm suggesting, I am suggesting issuing a new Reaper cert using the same issuer used for the management API cert. Because they will both share the same issuer, the existing management API truststore will trust the new cert issued for Reaper.

Whether it uses a Java keystore or pem encoded stuff isn't relevant here.

controllers/k8ssandra/reaper.go
Miles-Garnsey: Re-reading this... You're only using it for setting names for the truststore/keystore files that get materialised in the secret right? So this will just affect the keys in the aggregated secret that contains all of the encryption materials for the different Cassandra clusters? 

It may not matter to much if that's right, although I'd still suggest you use something like the namespace-name of the k8ssandra cluster since otherwise multiple clusters with the same name in different namespaces can overwrite each other...

controllers/k8ssandra/reaper.go
Miles-Garnsey: Fingerprints are probably sufficient until you need to think about rotation. We can think more about rotating the certs and CAs down the track (although CA related work is on hold at present, so I'm not sure when that will happen).

controllers/reaper/reaper_controller.go
Miles-Garnsey: OK, all good then. We'd set both in some other cases in the codebase, maybe that was wrong.

controllers/k8ssandra/reaper.go
rzvoncek: You both have valid points and it's not immediately clear how to proceed.

To move us forward, I'd like to point out that this PR is dealing with _using_ a certificate. It's not concerned what other uses that certificates have, nor where it came from. Genereating certs is mission-control, not k8ssandra, operator's job anyway.

If we decide we want dedicated certs for Reaper, we will need to track this with another ticket.

controllers/reaper/reaper_controller_test.go
rzvoncek: I went back and double checked this.

In the test, I go ahead and delete a Reaper object. This object is pointed to from a stateful set and from a secret via owner reference. The two are owned by the Reaper object. 

When writing this test, I did not know what precisely the _no gargage collection in envtests_ means. The Reaper object got deleted, so I was expecting the other things to disapear too - because that's what they do in an actual cluster. However, without the garbage collection, they won't disappear and I have to delete them manually. 

As a step forward, I propse removing the two `assert.Eventually` checks because they are confusing.

controllers/k8ssandra/reaper.go
Miles-Garnsey: The issue here is that you're not allowing for two different secrets to be used, I think this code assumes that the `authConfig.ClientSecretName` (which is the same secret used by management API) will always also be used by Reaper.

controllers/reaper/reaper_controller_test.go
rzvoncek: I've removed the `assert.Eventually`s.

controllers/k8ssandra/reaper.go
rzvoncek: This is correct. I'll make a fix.

rzvoncek: After some investigation, and a chat with Alex, we concluded that this is, in fact, a blocking issue.
We need a proper way of making Reaper (and its truststore) capable to tell two equally named clusters from different namespaces apart. It's not just for encryption, but for monitoring also.

We'll be putting this ticket on hold until we fix this in Reaper.

PR #1424, opened 2024-10-02
Fix empty log lines parsing in parse_cassandra_log, add automated tes…

### LINKED ISSUE #1409: Vector error when parsing Cassandra logs
A user is observing the following error:
```
thread 'vector-worker' panicked at /cargo/registry/src/index.crates.io-6f17d22bba15001f/onig-6.4.0/src/lib.rs:648:23:
Onig: Regex search error: retry-limit-in-match over
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'vector-worker' panicked at src/topology/builder.rs:920:30:
internal error: entered unreachable code: join error or bad poll
2024-09-13T16:31:52.484789Z ERROR transform{component_kind="transform" component_id=parse_cassandra_log component_type=remap}: vector::topology: An error occurred that Vector couldn't handle: the task panicked and was aborted.
2024-09-13T16:31:52.484869Z  INFO vector: Vector has stopped.
```
Which points to the [regex](https://github.com/k8ssandra/k8ssandra-operator/blob/3d3d735bc72056afee9a54d96f5c4d05159cd6a4/pkg/telemetry/vector.go#L144) that we use to parse Cassandra system logs. It looks like we ran into an unexpected log format that the regex fails on.

Unfortunately this causes a panic and a restart of the container, see vectordotdev/vector#15199.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-255) by [Unito](https://www.unito.io)

### LINKED ISSUE #1425: make manifests requires kustomize as pre-step
**What happened?**

Running tests runs ``manifests`` first, which runs our prepare-helm-release.sh and that requires Kustomize to be present. Yet, we don't install kustomize until later. So manifests target requires kustomize as its pre-step.

```console
 /Users/michael.burman/projects/datastax/k8ssandra-operator/bin/controller-gen "crd" rbac:roleName=k8ssandra-operator webhook paths="./..." output:crd:artifacts:config=config/crd/bases
| ./scripts/prepare-helm-release.sh
| ./scripts/prepare-helm-release.sh: line 13: kustomize: command not found
| make: *** [Makefile:135: manifests] Error 127
```

**Environment**

* K8ssandra Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* K8ssandra Operator Logs:

```
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-264) by [Unito](https://www.unito.io)

burmanm:
…ts for Vector

**What this PR does**:
Changes the rules to strip whitespace from logs, this removes empty loglines also (which cause a crash of Vector grok parsing).

Also, add tests that generate Vector files from our Telemetry rules and then run vector test against them.

Fixes #1409
Fixes #1425 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,15 @@ else
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(GO_FLAGS) ./apis/... ./pkg/... ./test/yq/... ./controllers/... -covermode=atomic -coverprofile coverage.out
 endif
 
+.PHONY: vector-test
+vector-test: ## Run vector tests
+	@echo Generating test files for Vector tests
+	$(eval TMP := $(shell mktemp -d))
+	VECTOR_TEST_FILES=true OUTPUT_PATH=$(TMP) go test -v ./pkg/telemetry -run=TestGenerateTomlTestFiles
+	@echo Running vector test files
+	OUTPUT_PATH=$(TMP) VECTOR=$(VECTOR) scripts/run-vector-tests.sh
+	rm -rf $(TMP)
+
 E2E_TEST_TIMEOUT ?= 3600s
 
 PHONY: e2e-test
@@ -326,6 +335,7 @@ KUSTOMIZE ?= $(LOCALBIN)/kustomize
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint # TODO Add linting to the GHA also
+VECTOR ?= $(LOCALBIN)/vector
 
 ## Tool Versions
 CERT_MANAGER_VERSION ?= v1.12.2
@@ -345,6 +355,10 @@ cert-manager-multi: ## Install cert-manager to the clusters
 		kubectl config use-context kind-k8ssandra-$$i; \
 		make cert-manager;  \
 	done
+
+.PHONY: vector-install
+vector-install:
+	curl --proto '=https' --tlsv1.2 -sSfL https://sh.vector.dev | bash -s -- --prefix $(LOCALBIN) -y
 
 # Install NGINX in the current Kind cluster using Helm and a values file that is suitable for
 # running e2e tests locally with a cluster created with setup-kind-multicluster.sh. Helm must be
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -159,6 +159,15 @@ else
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(GO_FLAGS) ./apis/... ./pkg/... ./test/yq/... ./controllers/... -covermode=atomic -coverprofile coverage.out
 endif
 
+.PHONY: vector-test
+vector-test: ## Run vector tests
+	@echo Generating test files for Vector tests
+	$(eval TMP := $(shell mktemp -d))
+	VECTOR_TEST_FILES=true OUTPUT_PATH=$(TMP) go test -v ./pkg/telemetry -run=TestGenerateTomlTestFiles
+	@echo Running vector test files
+	OUTPUT_PATH=$(TMP) VECTOR=$(VECTOR) scripts/run-vector-tests.sh
```
olim7t, line 168: [thought] Couldn't we just run the `vector` executable (and assert its return value) from within the Go test?
That way it would be a regular test, we wouldn't need a dedicated make target / script file / CI step to run it.

olim7t (approved):
Great that we could address that quickly.
Approved (pending the install issue, but you're already working on that).

Makefile
```diff
@@ -159,6 +159,15 @@ else
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(GO_FLAGS) ./apis/... ./pkg/... ./test/yq/... ./controllers/... -covermode=atomic -coverprofile coverage.out
 endif
 
+.PHONY: vector-test
+vector-test: ## Run vector tests
```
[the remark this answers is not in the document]
burmanm: I didn't want it to depend on the vector-install if some person (well, maybe just me?) has it installed through brew or other means. Also, since it was broken on some platforms like MacOS (the fix isn't merged yet), it would make running the tests difficult.

Makefile
burmanm: We could do that also, but I don't think we've done that too often in our tests. It's an external program, so not really running go test code. 

But would we really want Vector installation to be a dependency on running Go tests? ``make test`` would require then vector to be always installed.

PR #1426, opened 2024-10-04
Prevent changes of DC name or both adding and removing a DC at the same time. K8OP-263

### LINKED ISSUE #1422: Changing a DC name should be forbidden
There are no guardrails currently against changing the datacenter[].meta.name field in a K8ssandraCluster.

This could cause data loss and we should prevent this from happening.
Same applies to rack names (although cass-operator could prevent this already).

In any case, modifying datacenterName or datacenter[].meta.name should be forbidden by both the validation webhook and fail during reconcile.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-263) by [Unito](https://www.unito.io)
┆Fix Versions: 2024-10

Miles-Garnsey:
**What this PR does**:

Fixes #1422 

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
@@ -573,3 +573,34 @@ func (kc *K8ssandraCluster) GetClusterIdHash() string {
 func (k *K8ssandraCluster) GenerationChanged() bool {
 	return k.Status.ObservedGeneration < k.Generation
 }
+
+func DcAdded(oldSpec K8ssandraClusterSpec, newSpec K8ssandraClusterSpec) bool {
+	oldDcs := make([]string, 0)
+	for _, dc := range oldSpec.Cassandra.Datacenters {
+		oldDcs = append(oldDcs, dc.Meta.Name)
+	}
+	wasAdded := false
+	for _, dc := range newSpec.Cassandra.Datacenters {
+		if !utils.SliceContains(oldDcs, dc.Meta.Name) {
+			wasAdded = true
+			break
+		}
+	}
+	return wasAdded
+
+}
+
+func DcRemoved(oldSpec K8ssandraClusterSpec, newSpec K8ssandraClusterSpec) bool {
+	newDcs := make([]string, 0)
+	for _, dc := range newSpec.Cassandra.Datacenters {
+		newDcs = append(newDcs, dc.Meta.Name)
+	}
+	wasRemoved := false
+	for _, dc := range oldSpec.Cassandra.Datacenters {
+		if !utils.SliceContains(newDcs, dc.Meta.Name) {
+			wasRemoved = true
+			break
+		}
+	}
+	return wasRemoved
+}
--- a/apis/k8ssandra/v1alpha1/k8ssandracluster_types_test.go
+++ b/apis/k8ssandra/v1alpha1/k8ssandracluster_types_test.go
@@ -10,6 +10,7 @@ import (
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	stargateapi "github.com/k8ssandra/k8ssandra-operator/apis/stargate/v1alpha1"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestK8ssandraCluster(t *testing.T) {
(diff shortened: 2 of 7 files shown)


apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -172,7 +173,9 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) (admission.Warning
 	if err := validateUpdateNumTokens(oldCluster.Spec.Cassandra, r.Spec.Cassandra); err != nil {
 		return nil, err
 	}
-
+	if DcRemoved(oldCluster.Spec, r.Spec) && DcAdded(oldCluster.Spec, r.Spec) {
```
burmanm, line 176: issue: This should be tested in the webhook_test.go also.

apis/k8ssandra/v1alpha1/k8ssandracluster_types.go
```diff
@@ -573,3 +573,34 @@ func (kc *K8ssandraCluster) GetClusterIdHash() string {
 func (k *K8ssandraCluster) GenerationChanged() bool {
 	return k.Status.ObservedGeneration < k.Generation
 }
+
+func DcAdded(oldSpec K8ssandraClusterSpec, newSpec K8ssandraClusterSpec) bool {
```
burmanm, line 577: issue: No unit tests in the _types_test.go

controllers/k8ssandra/k8ssandracluster_controller_test.go
```diff
@@ -2398,6 +2405,10 @@ func changeClusterNameFails(t *testing.T, ctx context.Context, f *framework.Fram
 	err = f.Client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: clusterName}, k8c)
 	require.NoError(err, "failed to get K8ssandraCluster")
 
+	k8c.Spec.Cassandra.Datacenters[0].Meta.Name = "newDC1"
```
burmanm, line 2408: issue: This breaks the clusterName test, since the error text isn't tested. As you don't reset the status, the next test will fail, even if the clusterName check was removed since the DC name check would be checked.

controllers/k8ssandra/validate_cluster_update.go
```diff
@@ -0,0 +1,37 @@
...
+	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	"github.com/k8ssandra/k8ssandra-operator/pkg/utils"
+)
+
+func validateK8ssandraCluster(kc api.K8ssandraCluster) error {
+	oldDCs := make([]string, 0)
+	for dcName, _ := range kc.Status.Datacenters {
+		oldDCs = append(oldDCs, dcName)
+	}
+	newDcs := make([]string, 0)
```
burmanm, line 15: question: Why aren't these using the DcAdded and DcRemoved, when they were made exported methods?

controllers/k8ssandra/k8ssandracluster_controller_test.go
Miles-Garnsey: Ah, OK that's true. It will fail because of the DC name change, even if the newClusterName was what caused the failure. Nice catch, thank you.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
Miles-Garnsey: Fair enough, done.

controllers/k8ssandra/validate_cluster_update.go
Miles-Garnsey: Because the logic is different.

apis/k8ssandra/v1alpha1/k8ssandracluster_types_test.go
```diff
@@ -215,3 +216,25 @@ func TestGenerationChanged(t *testing.T) {
...
+
+func TestDcRemoved(t *testing.T) {
+	kcOld := createClusterObjWithCassandraConfig("testcluster", "testns")
+	kcNew := kcOld.DeepCopy()
+	kcOld.Spec.Cassandra.Datacenters = append(kcOld.Spec.Cassandra.Datacenters, CassandraDatacenterTemplate{
+		Meta: EmbeddedObjectMeta{
+			Name: "dc2",
+		},
+	})
+	require.True(t, DcRemoved(kcOld.Spec, kcNew.Spec))
```
burmanm, line 228: The following functionality (even if entirely broken) will pass this test:

```go
func DcRemoved(oldSpec K8ssandraClusterSpec, newSpec K8ssandraClusterSpec) bool {
  return true
}
```

This is a filler test which doesn't test any actual functionality. Least it could do is verify that before the append one would get false, like adding after line 222:

``require.False(t, DcRemoved(kcOld.Spec, kcNew.Spec))`` 

Or, to avoid faking that implementation with just:

```go
func DcRemoved(oldSpec K8ssandraClusterSpec, newSpec K8ssandraClusterSpec) bool {
  return len(oldSpec.Cassandra.Datacenters) == len(newSpec.Cassandra.Datacenters)
}
```

For example, just renaming some Datacenter name to verify that DcRemoved returned true even if the lengths match. And same sort for the DcAdded. 

Because right now it would be pretty easy to "refactor" and make tests pass while actually breaking the entire functionality and we do want to avoid that.

Miles-Garnsey: I think you're taking this a bit too far - you can almost always write an implementation that will pass a test but still be wrong SOMEHOW. 

But I've done as you suggested even so.

adejanovski:
@burmanm, could you do another round of review now that @Miles-Garnsey addressed your comments? Thanks!

controllers/k8ssandra/validate_cluster_update.go
burmanm: Well, then that means this isn't tested at all?

burmanm (approved):

burmanm:
I've approved due to time constraint, even if the reconcile process isn't really unit tested, but hopefully that is addressed next week.

apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go
```diff
@@ -172,7 +173,9 @@ func (r *K8ssandraCluster) ValidateUpdate(old runtime.Object) (admission.Warning
...
+		return nil, fmt.Errorf("renaming, as well as adding and removing DCs at the same time is prohibited as it can cause data loss")
```
olim7t, line 177: [nitpick] It would be nice to also include the names of the removed and added DCs.

PR #1432, opened 2024-10-17
Add support for priorityClassName (fixes #1034)

### LINKED ISSUE #1034: priorityClassName support
### What is missing?
priorityClassName support is surprisingly missing from the CRD.

Maybe I'm missing something but it doesn't seem possible right now to set this value.

### Why is this needed?
It's a standard (and highly used) field in Kubernetes.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-85) by [Unito](https://www.unito.io)
┆Fix Versions: 2024-10,2024-11,2025-01
┆Issue Number: K8OP-85

olim7t:
**What this PR does**:

Fixes #1034

**Checklist**
- [x] Changes manually tested


burmanm (approved):
suggestion: This was not part of the original ticket, so these are not mandatory changes, but if you feel like these would worth doing doing later (but not part of this PR), we could create a ticket for it:

That is, we don't actually check if the ``PriorityClass`` exists. Right now that could cause the Pod to have a priorityClassName that points to nothing and this probably prevents the pod from being scheduled correctly. I think this check could be implemented in cass-operator also, I have no preference to that right now. 

Second is actually creating PriorityClasses in k8ssandra-operator, but I would have to think about this feature more. Most users should probably set something like ``preemptionPolicy`` to ``Never`` to prevent new Cassandra pods from killing existing pods. Right now, that would be up to the user to understand what they're doing (like the entire feature, which I think we could consider "experienced users only"). 

If we use this feature in downstream products, then tracking PodStatus' nominatedNodeName might make sense also. And our ``PodDisruptionBudget`` policy is probably not very compatible with this feature either. The scheduling is quite complex on some of these and the documentation is a bit hand waving in some cases "we might or might not do this and maybe something else".

adejanovski:
@olim7t, could you rebase and merge the PR? It's been approved for a couple months now 😅

PR #1461, opened 2024-11-26
cass-operator v1.23.0

### LINKED ISSUE #1460: medusa calculates podNames incorrectly when unbalanced racks are used
**What happened?**
medusa's hostmap.go assumes evenly balanced racks, but this is not necessarily true. Given the size of 8 and 3 racks, the StS would end up with sizes 3,3,2 and this is fully functional. Yet, Medusa would believe this is actually 3,3,3 since it only divides and rounds. 

**Environment**

* K8ssandra Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* K8ssandra Operator Logs:

```
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-298) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-298

### LINKED ISSUE #1466: Lost Status updates, add datacenter contextName
**What happened?**
ContextName is missing from our Datacenter status, so if the user deletes the datacenters[] part of the DC, we can no longer necessarily pick up the correct target context. 

As for missing status updates, there are a lot.

In the datacenters.go, we don't actually use the updated value of kdcStatus, we simply copy stuff to it but do not reassign that new value (since a new reference was created with := earlier). As such that status update is entirely lost.

In reaper.go and stargate.go, we update the entire Status part and then miss any fields that might have been set previously. This was clear when I modified the tests to check status fields and they were not there when 10 seconds earlier those fields were there (such as the context). Thus, we need to copy the Stargate info to existing Status, not overwrite the entire Status with just what Stargate had in mind. Also, some updates in the stargate.go were also missing due to the same issue as in datacenters.go, where the new reference was never actually set back to the kcObject.Status.

**Environment**

* K8ssandra Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* K8ssandra Operator Logs:

```
```

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-300) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-300

burmanm:
**What this PR does**:
Updates to cass-operator v1.23.0 and fixes couple of bugs that I  encountered while updating.

Fixes #1466 
Fixes #1460 

**Checklist**
- [ ] Changes manually tested


rzvoncek (approved):
I've reviewed the part about #1460. I've seen it properly restore a DC with unbalanced racks, which is something the current main branch cannot do.

PR #1463, opened 2024-11-28
Document ReconcileResult type (fixes #1419)

### LINKED ISSUE #1419: Document ReconcileResult type
The [ReconcileResult](https://github.com/k8ssandra/k8ssandra-operator/blob/3a84cad7ff3b7d665acfb7273a6b1670476ba46c/pkg/result/result_helper.go#L12) interface is widely used throughout our codebase, yet from [this conversation](https://github.com/k8ssandra/k8ssandra-operator/pull/1022#discussion_r1773050446) it appears that there is some confusion around its usage.

Suggested tasks:

* document the API: what each implementation means, and how a caller should typically handle a result coming back from a sub-component (I believe it should always be the same pattern outside of the main reconciliation method).
* check that existing code conforms with that documentation. In particular, `reconciliation.ReconcileObject()` returns `result.Done()` when the reconciliation is successful, arguably it should be `result.Continue()`.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-261) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-261

olim7t:
**What this PR does**:

Fixes #1419

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/medusa/refresh_secrets.go
+++ b/pkg/medusa/refresh_secrets.go
@@ -52,7 +52,7 @@ func RefreshSecrets(dc *cassdcapi.CassandraDatacenter, ctx context.Context, clie
 		case recRes.IsRequeue():
 			requeues++
 			continue
-		case recRes.IsDone():
+		case !recRes.Completed():
 			continue
 		}
 		if requeues > 0 {
--- a/pkg/reconciliation/generic.go
+++ b/pkg/reconciliation/generic.go
@@ -21,6 +21,15 @@ type Reconcileable[T any] interface {
 // ReconcileObject ensures that desiredObject exists in the given state, either by creating it, or updating it if it
 // already exists.
 func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U) result.ReconcileResult {
+	recResult, _ := ReconcileAndGetObject[U, T](ctx, kClient, requeueDelay, desiredObject)
+	return recResult
+}
+
+// ReconcileAndGetObject ensures that desiredObject exists in the given state, either by creating it, or updating it if
+// it already exists. It returns the current state of the object on the server after the reconciliation.
+func ReconcileAndGetObject[U any, T Reconcileable[U]](
+	ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U,
+) (result.ReconcileResult, *U) {
 	objectKey := types.NamespacedName{
 		Name:      T(&desiredObject).GetName(),
 		Namespace: T(&desiredObject).GetNamespace(),
@@ -35,13 +44,13 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 		if errors.IsNotFound(err) {
 			if err := kClient.Create(ctx, T(&desiredObject)); err != nil {
 				if errors.IsAlreadyExists(err) {
-					return result.RequeueSoon(requeueDelay)
+					return result.RequeueSoon(requeueDelay), nil
 				}
-				return result.Error(err)
+				return result.Error(err), nil
 			}
-			return result.Continue()
+			return result.Continue(), &desiredObject
 		}
-		return result.Error(err)
+		return result.Error(err), nil
 	}
 
 	if !annotations.CompareHashAnnotations(T(currentCm), T(&desiredObject)) {
@@ -49,8 +58,9 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 		T(&desiredObject).DeepCopyInto(currentCm)
 		T(currentCm).SetResourceVersion(resourceVersion)
 		if err := kClient.Update(ctx, T(currentCm)); err != nil {
-			return result.Error(err)
+			return result.Error(err), nil
 		}
+		return result.Continue(), currentCm
 	}
-	return result.Continue()
+	return result.Continue(), currentCm
 }
--- a/pkg/reconciliation/generic.go
+++ b/pkg/reconciliation/generic.go
@@ -39,10 +39,9 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 				}
 				return result.Error(err)
 			}
-			return result.RequeueSoon(requeueDelay)
-		} else {
-			return result.Error(err)
+			return result.Continue()
 		}
+		return result.Error(err)
 	}
 
 	if !annotations.CompareHashAnnotations(T(currentCm), T(&desiredObject)) {
@@ -52,7 +51,6 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 		if err := kClient.Update(ctx, T(currentCm)); err != nil {
 			return result.Error(err)
 		}
-		return result.RequeueSoon(requeueDelay)
 	}
 	return result.Continue()
 }
--- a/pkg/reconciliation/generic.go
+++ b/pkg/reconciliation/generic.go
@@ -18,7 +18,8 @@ type Reconcileable[T any] interface {
 	*T
 }
 
-// Try with U, a type of any whose POINTER still fulfils Reoncilable...
+// ReconcileObject ensures that desiredObject exists in the given state, either by creating it, or updating it if it
+// already exists.
 func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U) result.ReconcileResult {
 	objectKey := types.NamespacedName{
 		Name:      T(&desiredObject).GetName(),
@@ -53,5 +54,5 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 		}
 		return result.RequeueSoon(requeueDelay)
 	}
-	return result.Done()
+	return result.Continue()
 }
--- a/pkg/result/result_helper.go
+++ b/pkg/result/result_helper.go
@@ -1,20 +1,54 @@
 package result
 
 import (
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"time"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-// Copyright DataStax, Inc.
-// Please see the included license file for details.
+// This is just so that we can reference TerminalError in the Godoc of [Error]
+var _ error = reconcile.TerminalError(nil)
 
+// ReconcileResult represents the result of a step in the reconciliation process.
+//
+// We typically split the top-level Reconcile() method of a controller into multiple sub-functions. Each of these
+// functions uses ReconcileResult to communicate to its caller how the current iteration should proceed. There are 4
+// possible implementations: [Continue], [Done], [RequeueSoon], and [Error].
+//
+// The idiomatic way to handle a ReconcileResult in an intermediary sub-function is:
+//
+//	if recResult := callStep1(); recResult.Completed() {
+//	   // Global success, requeue or error: stop what we're doing and propagate up
+//	   return recResult
+//	}
+//	// Otherwise, proceed with the next step(s)
+//	if recResult := callStep2(); recResult.Completed() {
+//	   // etc...
+//
+// The idiomatic way to handle a ReconcileResult in the top-level Reconcile() method is:
+//
+//	recResult := callSomeSubmethod()
+//	// Possibly inspect the result (e.g. to set an error field in the status)
+//	return recResult.Output()
 type ReconcileResult interface {
+	// Completed indicates that this result is *NOT* a [Continue].
+	//
+	// In other words, it means that the current iteration of the reconciliation loop is complete, and the top-level
+	// Reconcile() method should return to the controller runtime (either with a success or terminal error that will
+	// stop the entire reconciliation loop, or a requeue or regular error that will trigger a retry).
 	Completed() bool
+	// Output converts this result into a format that the main Reconcile() method can return to the controller runtime.
+	//
+	// Calling this method on a [Continue] will panic.
 	Output() (ctrl.Result, error)
+	// IsError indicates whether this result is an [Error].
 	IsError() bool
+	// IsRequeue indicates whether this result is a [RequeueSoon].
 	IsRequeue() bool
+	// IsDone indicates whether this result is a [Done].
 	IsDone() bool
+	// GetError returns the wrapped error if the result is an [Error], otherwise it returns nil.
 	GetError() error
 }
 
@@ -121,18 +155,31 @@ func (r errorOut) GetError() error {
 	return r.err
 }
 
+// Continue indicates that the current step in the reconciliation is done. The caller should proceed with the next step.
 func Continue() ReconcileResult {
 	return continueReconcile{}
 }
 
+// Done indicates that the entire reconciliation loop was successful.
+// The caller should skip the next steps (if any), and propagate the result up the stack. This will eventually reach the
+// top-level Reconcile() method, which should stop the reconciliation.
 func Done() ReconcileResult {
 	return done{}
 }
 
+// RequeueSoon indicates that the current step in the reconciliation requires a requeue after a certain amount of time.
+// The caller should skip the next steps (if any), and propagate the result up the stack. This will eventually reach the
+// top-level Reconcile() method, which should schedule a requeue with the given delay.
 func RequeueSoon(after time.Duration) ReconcileResult {
 	return callBackSoon{after: after}
 }
 
+// Error indicates that the current step in the reconciliation has failed.
+// The caller should skip the next steps (if any), and propagate the result up the stack. This will eventually reach the
+// top-level Reconcile() method, which should return the error to the controller runtime.
+//
+// If the argument is wrapped with [reconcile.TerminalError], the reconciliation loop will stop; otherwise, it will be
+// retried with exponential backoff.
 func Error(e error) ReconcileResult {
 	return errorOut{err: e}
 }
--- a/pkg/result/result_helper.go
+++ b/pkg/result/result_helper.go
@@ -32,11 +32,13 @@ var _ error = reconcile.TerminalError(nil)
 //	// Possibly inspect the result (e.g. to set an error field in the status)
 //	return recResult.Output()
 type ReconcileResult interface {
-	// Completed indicates that this result is *NOT* a [Continue].
+	// Completed indicates that the current iteration of the reconciliation loop is complete, and the top-level
+	// Reconcile() method should return [ReconcileResult.Output] to the controller runtime.
 	//
-	// In other words, it means that the current iteration of the reconciliation loop is complete, and the top-level
-	// Reconcile() method should return to the controller runtime (either with a success or terminal error that will
-	// stop the entire reconciliation loop, or a requeue or regular error that will trigger a retry).
+	// This returns true for a [Done] or terminal [Error] (where the output will stop the entire reconciliation loop);
+	// and for a [RequeueSoon] or regular [Error] (where the output will trigger a retry).
+	//
+	// This returns false for a [Continue].
 	Completed() bool
 	// Output converts this result into a format that the main Reconcile() method can return to the controller runtime.
 	//
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -842,7 +842,7 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace 
 	require.NoError(err, "failed to patch K8ssandraCluster in namespace %s", namespace)
 	checkStargateReady(t, f, ctx, stargateKey)
 	checkStargateK8cStatusReady(t, f, ctx, kcKey, dcKey)
-	checkContainerPresence(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
+	checkContainerPresenceEventually(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
 	checkVectorAgentConfigMapPresence(t, ctx, f, dcKey, stargate.VectorAgentConfigMapName)
 
 	t.Log("check that if Stargate is deleted directly it gets re-created")
@@ -853,7 +853,7 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace 
 	require.NoError(err, "failed to delete Stargate in namespace %s", namespace)
 	checkStargateReady(t, f, ctx, stargateKey)
 
-	checkContainerPresence(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
+	checkContainerPresenceEventually(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
 	checkVectorAgentConfigMapPresence(t, ctx, f, dcKey, stargate.VectorAgentConfigMapName)
 
 	t.Log("delete Stargate in k8ssandracluster resource")
@@ -2322,6 +2322,15 @@ func checkContainerPresence(t *testing.T, ctx context.Context, f *framework.E2eF
 	podTempSpec := specFunction(t, ctx, f, podKey, kc)
 	_, containerFound := cassandra.FindContainer(podTempSpec, containerName)
 	require.True(t, containerFound, "cannot find Container in pod template spec")
+}
+
+func checkContainerPresenceEventually(t *testing.T, ctx context.Context, f *framework.E2eFramework, podKey framework.ClusterKey, kc *api.K8ssandraCluster, specFunction func(t *testing.T, ctx context.Context, f *framework.E2eFramework, dcKey framework.ClusterKey, kc *api.K8ssandraCluster) *corev1.PodTemplateSpec, containerName string) {
+	t.Logf("check that %s contains Container named %s", podKey.Name, containerName)
+	require.Eventually(t, func() bool {
+		podTempSpec := specFunction(t, ctx, f, podKey, kc)
+		_, containerFound := cassandra.FindContainer(podTempSpec, containerName)
+		return containerFound
+	}, polling.stargateReady.timeout, polling.stargateReady.interval, "cannot find Container in pod template spec")
 }
 
 func checkContainerDeleted(t *testing.T, ctx context.Context, f *framework.E2eFramework, podKey framework.ClusterKey, kc *api.K8ssandraCluster, specFunction func(t *testing.T, ctx context.Context, f *framework.E2eFramework, dcKey framework.ClusterKey, kc *api.K8ssandraCluster) *corev1.PodTemplateSpec, containerName string) {
--- a/controllers/k8ssandra/medusa_reconciler.go
+++ b/controllers/k8ssandra/medusa_reconciler.go
@@ -109,11 +109,7 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 				return result.Error(err)
 			}
 			purgeCronJob.SetLabels(labels.CleanedUpByLabels(kcKey))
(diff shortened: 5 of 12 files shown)


pkg/reconciliation/generic.go
```diff
@@ -53,5 +54,5 @@ func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient cli
 		}
 		return result.RequeueSoon(requeueDelay)
 	}
-	return result.Done()
+	return result.Continue()
```
olim7t, line 57: I know this might be controversial, but given my current understanding of the API that's the right thing to do.

pkg/medusa/refresh_secrets.go
```diff
@@ -52,7 +52,7 @@ func RefreshSecrets(dc *cassdcapi.CassandraDatacenter, ctx context.Context, clie
 		case recRes.IsRequeue():
 			requeues++
 			continue
-		case recRes.IsDone():
+		case !recRes.Completed():
```
olim7t, line 55: Changed this to keep the exact same behavior, but the overall function has issues: the block `if requeues > 0 { ... }` at line 58 is never reached (either with this version or the original ones), so retries are not behaving as expected.

pkg/reconciliation/generic.go
```diff
@@ -18,7 +18,8 @@ type Reconcileable[T any] interface {
 	*T
 }
 
-// Try with U, a type of any whose POINTER still fulfils Reoncilable...
+// ReconcileObject ensures that desiredObject exists in the given state, either by creating it, or updating it if it
+// already exists.
 func ReconcileObject[U any, T Reconcileable[U]](ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U) result.ReconcileResult {
```
olim7t, line 23: [thought] This is off-topic, but I don't understand why this function schedules a requeue after a successful creation.

pkg/result/result_helper.go
```diff
@@ -1,20 +1,54 @@
...
+// ReconcileResult represents the result of a step in the reconciliation process.
+//
+// We typically split the top-level Reconcile() method of a controller into multiple sub-functions. Each of these
+// functions uses ReconcileResult to communicate to its caller how the current iteration should proceed. There are 4
+// possible implementations: [Continue], [Done], [RequeueSoon], and [Error].
+//
+// The idiomatic way to handle a ReconcileResult in an intermediary sub-function is:
+//
+//	if recResult := callStep1(); recResult.Completed() {
+//	   // Global success, requeue or error: stop what we're doing and propagate up
```
Miles-Garnsey, line 22: Issue: based on the logic you've described [here](https://github.com/k8ssandra/k8ssandra-operator/pull/1022#discussion_r1777359540) it sounded like a requeue (which is not a success) will also be `Completed()`?

pkg/result/result_helper.go
```diff
@@ -121,18 +155,31 @@ func (r errorOut) GetError() error {
 	return r.err
 }
 
+// Continue indicates that the current step in the reconciliation is done. The caller should proceed with the next step.
 func Continue() ReconcileResult {
 	return continueReconcile{}
 }
 
+// Done indicates that the entire reconciliation loop was successful.
```
Miles-Garnsey, line 163: Issue: from a design perspective, this seems tricky, since only the last step should ever call `Done()` based on this description. No earlier step would ever know that the entire reconcilation was `Done()` from what I can see? 

I think if we're getting into this topic, we should make `Done()` more descriptive - perhaps it needs a rename to `AllReconciliationDone()` or something similar to that, which indicates that all reconciliation steps to reconcile a given object (from the `req` passed to the controller) is done. 

We need to be explicit about WHAT is done - that's the issue we're currently struggling with. Naming these implementations correctly might be a good starting point.

Miles-Garnsey (changes requested):
I've only offered a very preliminary review here. There are a few overarching considerations I'd like to bring to your attention:

1. Issue: whatever you decide `Completed()`, `Done()`, and the rest mean needs to align with the usage in ReconcileObject(). Based on that, I have concerns around the blast radius of this change. Have you done an analysis of how `ReconcileObject` is called throughout the codebase to ensure that changing it's return type in the case of success doesn't break any of the calls? As the core of the reconciliation process we need to treat this delicately.
2. Issue: across the board, the issue here is that we have never been explicit about WHAT particular reconciliation is `Done()` `Completed()` or needs requeue or is in an error state. We should be more explicit about this in the naming of these implementations to avoid problems in future.

pkg/reconciliation/generic.go
Miles-Garnsey: Because that's the pattern followed throughout the codebase. Object creation is asynchronous, acceptance by the API server does not necessarily mean that something won't go wrong downstream of that. As a result, we check again for the success of the creation operation before continuing, which indeed does mean a requeue.

There is stuff to do with optimistic locking and the way the caches work on the controller which make this necessary - try removing it if you're curious as to what breaks.

pkg/reconciliation/generic.go
Miles-Garnsey: Issue: I don't think it has to be controversial, but I'd actually like to see all of this in a design doc so that I can understand what you're trying to achieve with these changes. To me, they have a very wide blast radius, and there are some questions and themes in here which suggest that we probably need to align on certain topics (e.g. how caches work, how optimistic locking works, how the API server is asynchronous) before moving forward.

One of the main changes I think we need here is more descriptive naming for these implementations; and therefore alignment on what each of them is intended to signify. Once we have that it should be easy to figure out what the answer should be to this return type.

pkg/result/result_helper.go
```diff
@@ -1,20 +1,54 @@
...
+//	if recResult := callStep2(); recResult.Completed() {
+//	   // etc...
+//
+// The idiomatic way to handle a ReconcileResult in the top-level Reconcile() method is:
+//
+//	recResult := callSomeSubmethod()
+//	// Possibly inspect the result (e.g. to set an error field in the status)
+//	return recResult.Output()
 type ReconcileResult interface {
+	// Completed indicates that this result is *NOT* a [Continue].
```
Miles-Garnsey, line 35: Issue: Defining something in terms of what it is not is a large part of our issue here I think. Can we define `Completed()` as what it is, rather than what it is not? Would it be clearer if we renamed it to `BailToMainReconciliation()` or something like that? Is that what `Completed()` really means?

pkg/result/result_helper.go
olim7t: Yes, a requeue is `Completed()`:
```golang
func (c callBackSoon) Completed() bool {
	return true
}
```
But I don't see any issue here: if a step wants a requeue, we stop what we're doing and propagate up. The top-level Reconcile() will eventually return `recResult.Output()` (which is `ctrl.Result{Requeue: true, RequeueAfter: c.after}, nil`) to the controller runtime.

pkg/reconciliation/generic.go
olim7t: > Object creation is asynchronous, acceptance by the API server does not necessarily mean that something won't go wrong downstream of that.

Is that documented anywhere? The `client` package docs do not go into many details.

pkg/result/result_helper.go
Miles-Garnsey: Let me put my concern in another way. Right now, we have a struct implementation of this interface instantiated with `Done()` which is also `Completed()`. `Requeue()` and `Error()` are also `Completed()`. And yet, the English language meaning of the word `Completed()` is exactly the same as the meaning of the word `Done()`.

To me, having a getter function `Completed()` for a field across three different struct implementations ( `Done()`, `Requeue()` and `Error()` - some of which do not mean "Done" in the linguistic sense) is completely untenable. 

If you want to continue having something which means "this reconciliation run has progressed as far as possible, and needs to jump back to the top level of the reconciliation to be handled appropriately", we need a different word to `Completed()`. At that point, I think we can have a more productive conversation about what behaviours we are trying to model. But until we can express that intention in a function name, I would rather retain the explicit switch statements we have for `Done()`, `Requeue()` and `Error(), since this makes it explicit what behaviours we are handling.

pkg/reconciliation/generic.go
Miles-Garnsey: That's just how asynchronous HTTP APIs work. But I think the more critical thing here is that (from memory) due to some of the caching behaviour of these clients, the object will not appear if you try to go do a Get in some circumstances. I encourage you to try to remove the requeues and see if things break - I believe they do.

pkg/result/result_helper.go
Miles-Garnsey: This is kind of why I think this needs a design doc too - because at the moment it isn't clear to me what behaviour or states we are trying to model, and we need to discuss and agree a target state rather than just say that we're "documenting" existing behaviour but then making quite impactful changes to the core of the reconciliation logic.

Miles-Garnsey (dismissed):
Issue: this is potentially a big issue - when changing from `Done()` to `Continue()` in the ReconcileObject I think you're going to see bugs depending on the caching behaviour of the client. 

These errors might manifest in two ways:
1. In a caching client, where the return from this function is checked using `Completed()`, this will probably cause downstream failures, since (from memory) the cache refreshes only when the reconcile is restarting, and not while reconciliation is taking place. If you reference that object in downstream logic it may not exist (or be up to date). This is because `Continue()` is not `Completed()` while `Done()` is `Completed()`.
2. In all clients (caching and non-caching) this may cause failures when a switch statement is used to check the return result and expects to see a `Done()`. Mostly the switch statements should have `IsRequeue()`, `IsError()` with `Completed()` and `Done()` then falling through to the default case, but have we checked every call site to ensure this is the case?

You have a lot of unit/integration test failures here, which might be due to either of these issues - I'm not 100% sure.

pkg/medusa/refresh_secrets.go
Miles-Garnsey: I think you're right, we should drop the continue statements for this to make sense.

olim7t:
> You have a lot of unit/integration test failures here, which might be due to either of these issues - I'm not 100% sure.

Those failures are caused by a bug that was masked by the previous implementation.

The switch statement in `reconcileStargateConfigMap` used to return this for a requeue:
```go
	case recRes.IsRequeue():
		return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
```
I changed it to return `recResult.Output()`, which for a requeue is:
```go
	return ctrl.Result{Requeue: true, RequeueAfter: c.after}, nil
```
Setting `Requeue` should normally not make a difference, because the API docs indicate that it is implied by `RequeueAfter`.

However in `StargateReconciler.Reconcile()` we have this:
```go
		if stargateConfigResult.Requeue {
			return ctrl.Result{RequeueAfter: r.DefaultDelay}, nil
		}
```
Which is wrong, because it doesn't consider that `Requeue` could be implicitly set. The correct condition is:
```go
		if stargateConfigResult.Requeue || stargateConfigResult.RequeueAfter > 0 {
```
If I fix that condition on `main`, the tests fail in the same places. They need to be fixed to expect that requeue that was previously ignored.

pkg/reconciliation/generic.go
olim7t: I don't know, [this example in the controller-runtime codebase](https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/crd/main.go#L90) doesn't requeue. They reference `pod.CreationTimestamp`, a field that is set by the server during the creation operation, immediately after a successful `Create()` call.

pkg/result/result_helper.go
olim7t: > we need to discuss and agree a target state rather than just say that we're "documenting" existing behaviour

Yes, the goal of this PR is to have that discussion. I proposed an initial version, let's iterate on it with reviews until we reach a common understanding.

> this needs a design doc too

Are the go-docs not self-sufficient? Not that I mind having a separate document, but I wouldn't know what else to put in there.

> And yet, the English language meaning of the word `Completed()` is exactly the same as the meaning of the word `Done()`.

I've deliberately avoided going down the naming rabbithole so far... May I suggest that we keep that for a second step? I feel like it will get too messy if we rename as we go.
(but yes I agree with most of your comments)

pkg/reconciliation/generic.go
Miles-Garnsey: I'm not sure, my understanding was that at least caching meant that requeues were necessary. But I'm open to being wrong here :)

pkg/result/result_helper.go
Miles-Garnsey: We already chatted about this on the call, but I think a design doc is a better place to facilitate discussion than a PR since it has better threaded comments etc. PRs for discussion will get messy, but it is good indeed to have some code to point at, so maybe both is best for this case.

test/e2e/suite_test.go
```diff
@@ -842,7 +842,7 @@ func createSingleDatacenterCluster(t *testing.T, ctx context.Context, namespace
 	require.NoError(err, "failed to patch K8ssandraCluster in namespace %s", namespace)
 	checkStargateReady(t, f, ctx, stargateKey)
 	checkStargateK8cStatusReady(t, f, ctx, kcKey, dcKey)
-	checkContainerPresence(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
+	checkContainerPresenceEventually(t, ctx, f, stargateDeploymentKey, k8ssandra, getPodTemplateSpec, stargate.VectorContainerName)
```
olim7t, line 845: Same issue as in unit tests: because the requeue is not ignored anymore, this is taking longer than before and subject to timing issues.

pkg/reconciliation/generic.go
burmanm: No, if write is success, it really is written. Writing is not an asynchronous operation in Kubernetes, it's actually strictly consistent one (using Raft).

Reading is asynchronous however as we use a cache.

pkg/result/result_helper.go
burmanm: > Are the go-docs not self-sufficient? Not that I mind having a separate document, but I wouldn't know what else to put in there.

They are and should be the only place where API is documented. Everything else is out of the code and probably not maintained or reliable.

pkg/result/result_helper.go
burmanm: > No earlier step would ever know that the entire reconcilation was Done() from what I can see?

This is exactly what it is, the reconciling process is done. It's not even really a necessary process didn't end in a pkg, but in the Reconcile() function. Just return reconcile.Result{}, nil.

> We need to be explicit about WHAT is done - that's the issue we're currently struggling with. Naming these implementations correctly might be a good starting point.

Are we having issues? This API didn't even begin here and the original usage had no issues. I'm not convinced we need this ReconcileResult API at all, but then that's not what we're discussing here.

pkg/result/result_helper.go
olim7t: Exactly.
- loop: the entire process, from the moment we receive the first request to the moment we consider the object has reached the expected state.
- iteration: one call to the top-level `Reconcile()` method. One loop may consist of multiple iterations.
- step: an action in an iteration. Maps to sub-function calls for the sake of this API.

`Done` or `Error(terminalError)` ends the loop.
`Requeue` or `Error(nonTerminalError)` ends the current iteration.
`Continue` proceeds to the next step.

pkg/reconciliation/generic.go
olim7t: OK, based on this I've modified ReconcileObject to remove those extra requeues. It's in a separate commit if we decide to revisit it later.

Miles-Garnsey: > OK, based on this I've modified ReconcileObject to remove those extra requeues. It's in a separate commit if we decide to revisit it later.

That's incorrect @olim7t - based on what Micke has just said above `Reading is asynchronous however as we use a cache.` That means that (while I'm wrong on the consistency level of writes, I'm right on reads, and some of our logic will probably fail due to failure to refresh the cache.

pkg/result/result_helper.go
Miles-Garnsey: This isn't about documentation, it is about the development process. We do deep discussions in design docs, not on PRs. If we choose to do it differently here, I'm concerned that we will end up in the same situation that we ended up in when we started making changes to the name overrides in labels and downstream resources, where the blast radius of the change was not adequately considered and customers suffered issues as a result. 

I strongly advocate for the creation of a design doc for this piece of work to avoid customer impacts.

pkg/result/result_helper.go
Miles-Garnsey: > And this isn't a place to start discussing Done/Completed/Concluded/Finished/Ended

This is exactly the place to do so IMO because that's what we're having issues with. As I've said before, I would have serious concerns about any iteration of this PR which does not disambiguate between `Completed` and `Done` since they mean exactly the same thing in plain english and this is unclear from the way that these concepts are named in the codebase. 

> loop: the entire process, from the moment we receive the first request to the moment we consider the object has reached the expected state.
> iteration: one call to the top-level Reconcile() method. One loop may consist of multiple iterations.
> step: an action in an iteration. Maps to sub-function calls for the sake of this API.

While these concepts might form a useful basis for a future framework, and I think that bears discussion, I will once again re-iterate that they do not currently exist in the codebase. Also, once again, the design of a potential future framework is a matter for a design doc, not a PR.

pkg/result/result_helper.go
Miles-Garnsey: > Are we having issues? This API didn't even begin here and the original usage had no issues. I'm not convinced we need this ReconcileResult API at all, but then that's not what we're discussing here.

Maybe this should be something we're discussing. We are definitely having issues because there is no consistency in the usage of this API in the codebase and there are regular disagreements on what Done, Continue, and Completed mean, as there have been for several years.

pkg/reconciliation/generic.go
olim7t: By "extra" requeues I meant the one after a successful Create, and the one after a successful Update. I pushed the change so you can check the diff.

burmanm: > That means that (while I'm wrong on the consistency level of writes), I'm right on reads, and some of our logic will probably fail due to failure to refresh the cache.

This depends on what is read after the object was written. If it's something that the Kubernetes API returns, then it's consistent right away. For example, Olivier's example of using creationTimestamp, probably resourceVersion and uid etc are directly accessible in the object you just posted.

However, for many use cases, what happens is that there's other controllers doing some extra processing and then modifying the object on the Kubernetes side (as a lot of Kubernetes processing itself is async with controllers). Can't access or Update .Status for example for some objects, like Pods. Or in our runtimes, if k8ssandra-operator creates CassandraDatacenter, it probably can't Update the Status field of CassandraDatacenter even instantly after, because cass-operator has modified it and the resourceVersions no longer match.

The other thing that should be understood is that Read after Write is cached, but of course Write itself updates the object that was sent. So doing:

obj = Create(obj)
a := obj.ObjectMeta.uid
// and if nothing else is touching our object, we can patch it afterwards also.

Is valid. However, doing:

obj = Create(obj)
obj = Get(obj)
a := obj.ObjectMeta.uid

Is not valid. The Get() call is coming from cache and would probably not return an object at all. Using Reader interface from manager will allow doing this as it isn't cached, but we don't really use that interface a lot (I think only ClientConfigs does that).

pkg/result/result_helper.go
burmanm: > And yet, the English language meaning of the word Completed() is exactly the same as the meaning of the word Done().

What is the English language definition difference of while and while do then? Yet, semantics for those two are different and in every programming language. One has while executed after first execution, while one has while executed before execution.

This isn't English.

pkg/result/result_helper.go
olim7t: > No, there hasn't been

At the very least there's been some ambiguity, as evidenced by the changes we are doing in this PR to align different parts of the code.

> it had a clear definition there how things worked

This might be your longer experience of cass-operator speaking, but personally I wouldn't call this API clear, it took way more effort to understand than it should.
Anyway, whether it is clear or not, I don't see any downside to documenting the intent.

pkg/reconciliation/generic.go
adejanovski: What we're saying here is that a requeue may or may not be needed, depending on the case.
If the requeue itself is harmless, just adding a little delay, but is protecting the code from potential bugs we might as well keep it, don't we?
@olim7t, for the sake of moving this PR forward and keeping its scope to (mostly) the original one, I'd suggest to rollback this change and have a separate discussion for it.

pkg/result/result_helper.go
adejanovski: I agree with Miles, and let's not forget many of the people participating in these projects are not necessarily english native speakers. This API brings a level of confusion that is high enough to require having that conversation. There's overlap between `Completed()` (which should really be `IsCompleted()`), `Done()` and `Continue()` among other things which require to take a break and think of what this API is solving and how it should do it.
The fact that it's been used successfully (which is debatable) for the past few years isn't a valid reason to push back on the discussion.
I also think it's fair to have a separate doc, in the form of a markdown file, contributed as a PR so that the community can see it, to describe our view on the reconcile process and how `ReconcileResult` should be used.
"Design doc" might be an overstatement here and "doc" would probably be enough.

pkg/result/result_helper.go
adejanovski: > And this isn't a place to start discussing Done/Completed/Concluded/Finished/Ended

It's ok to discuss the names and agree on their meaning, which doesn't mean we're changing those at the moment. Getting a common understanding is the first step to drive some changes that would clarify things going forward.

I agree with Miles that we shouldn't define something as what it is not, but rather as what it is.
@olim7t, could you maybe rework this comment by removing the first line, and also `In other words, it means that`. The rest of your comment looks good to me.

olim7t: I agree too, but I think it's more of a reflection about the design rather than the wording of the comment. "Not continue" is an accurate description when you consider the idiomatic usage of this method: `if recResult.Completed() { return recResult }` -- "if I've been asked not to continue, return now".

Anyway I'll switch the two sentences if that helps.

pkg/reconciliation/generic.go
olim7t: > What we're saying here is that a requeue may or may not be needed, depending on the case.

This thread has turned into a generic conversation about requeues, but we are in fact looking at a very specific case. I've only made two changes related to requeues in this PR:
- ReconcileObject tries to create the object and succeeds
- ReconcileObject tries to update the object and succeeds

(it's all in [Don't requeue successful writes in ReconcileObject](https://github.com/k8ssandra/k8ssandra-operator/pull/1463/commits/5bbd6af0294d3b8a2072ab1c494773771019c433) if anyone wants to look at the diff)

If we accept the assertion that writes are synchronous then there's absolutely no need to requeue. ReconcileObject wanted to apply changes, those changes are applied, we're done.

Other controllers and cached reads are valid concerns that we encounter in other places, but not here.

pkg/result/result_helper.go
burmanm: > which should really be IsCompleted()

Ugly Java POJO, but this isn't a Java project. We don't need to indicate it's a boolean when a function has return value of boolean. Same as with "Get" when fetching an object.

> I also think it's fair to have a separate doc, in the form of a markdown file, contributed as a PR so that the community can see it

Why can't the users view what's defined in the godoc? It would be publicly available and generated already, like currently:

https://pkg.go.dev/github.com/k8ssandra/k8ssandra-operator@v1.20.3/pkg/reconciliation

What Olivier wrote here would appear there. We can add examples, descriptions etc. Adding markdown doc somewhere assumes users know to search for it (and hope it's not outdated) instead of the "standard" place.

pkg/medusa/refresh_secrets.go
adejanovski: While there's a bug, it seems to be out of the scope of this PR.
Can we revert the change and file another issue for this?

pkg/reconciliation/generic.go
adejanovski: As said in the previous comment, if we all agree that writes are synchronous, there's no need to force a requeue at this level. Creating a configmap for example won't require a requeue.

pkg/result/result_helper.go
adejanovski: In an effort to move forward again, let's just use these in code comments as docs on how to use this interface.
The comments will be available in the IDE and in the godocs, without requiring to look at an external doc.
We can add more information here if we feel there's not enough for anyone to build a good understanding of how to use it.
We can extract this into a doc if that makes sense to anyone, but let's not make that mandatory.

pkg/result/result_helper.go
adejanovski: Let's focus on whether or not we agree with the comments that @olim7t added on each function.
@burmanm @Miles-Garnsey, do these comments look good? 

Let's not change names for now and just build a common understanding.

pkg/reconciliation/generic.go
```diff
@@ -18,8 +18,18 @@ type Reconcileable[T any] interface {
...
+	recResult, _ := ReconcileAndGetObject[U, T](ctx, kClient, requeueDelay, desiredObject)
+	return recResult
+}
+
+// ReconcileAndGetObject ensures that desiredObject exists in the given state, either by creating it, or updating it if
+// it already exists. It returns the current state of the object on the server after the reconciliation.
+func ReconcileAndGetObject[U any, T Reconcileable[U]](
+	ctx context.Context, kClient client.Client, requeueDelay time.Duration, desiredObject U,
+) (result.ReconcileResult, *U) {
```
olim7t, line 32: I propose we add this variant, it's almost the same but returns the state of the object on the server after the reconciliation.
This is useful in two cases:
- the hash annotation matched so the object was not updated, but we're interested in a field that was updated independently by another controller. For example the IP of a LoadBalancer service in its status.
- we're interested in a field that is set by the write operation itself, for example `CreationTimestamp`.

Miles-Garnsey: I think this solves some problems but might be overoptimising - it introduces more complexity than I'd like and relies on folks understanding the caching behaviour. Based on the conversation here I think that is tricky for even experienced users and I'd rather avoid callers having to make more decisions about those subtleties.

olim7t: > relies on folks understanding the caching behaviour

How so? We're either returning the result of a read that observed matching hash annotations, or the result of a create or update (not subject to caching).

If we don't have this variant, we'll need to resort to `ReconcileObject` followed by a separate `Get`. If anything, it sounds like it could produce more caching issues (Get not seeing the result of a write in ReconcileObject).

burmanm: > If we don't have this variant, we'll need to resort to ReconcileObject followed by a separate Get. If anything, it sounds like it could produce more caching issues (Get not seeing the result of a write in ReconcileObject).

In that behavior, we would need to reconcile and wait for the cache to have that object (and potentially reconcile again, fail etc). Otherwise, we need to use the Create/Update one, but never Create+Get or Update+Get combo.

So it makes sense to get the object back if user knows what to do with it. Actually, why would we even have any non-returning version? Create/Update always return the object also.

adejanovski (approved):

pkg/reconciliation/generic.go
Miles-Garnsey: I'm with @burmanm on this. The Get call you're making here won't actually result in a cache update so this variant isn't going to solve the problem.

When I say that folks would need to understand the caching behaviour, what I mean is that they need to know that the object returned by this new variant is not necessarily up to date. They also need to reason about what fields are up to date, which is non-trivial in my view.

I am in favour of just always re-queuing so that we elide these complexities.

adejanovski: After chatting with @burmanm, it looks like all 3 of us (Olivier, Micke and myself) agree that a requeue shouldn't be done systematically. Let's then move forward with what @olim7t did here, and follow up with a refactoring that merges this new `ReconcileAndGetObject()` function into just `ReconcileObject()`, changing the signature to also return the object. That'll require a few changes in the code that calls this function to ignore the object when it's not needed.
If a requeue is needed, one should know and add it after the call to `ReconcileObject()`.
If this causes bugs, they should surface in our tests, otherwise it means we're missing some tests.

c3-amitsalunke (approved):
Changes look good, approved!

PR #1521, opened 2025-05-05
Reaper default security context

### LINKED ISSUE #1520: Reaper security context should be secure by default
**What is missing?**

Reaper should satisfy the following requirements 

Run as a non-root user
Leverage a read-only root filesystem
Disallow privilege escalation
Drop all capabilities
Comply with OpenShift nonroot-v2 Security Context Constraint

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-327) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-327

Miles-Garnsey:
**What this PR does**:

Fixes #1520 

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pkg/reaper/deployment.go
+++ b/pkg/reaper/deployment.go
@@ -2,12 +2,13 @@ package reaper
 
 import (
 	"fmt"
+	"math"
+	"strings"
+
 	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/encryption"
 	"k8s.io/utils/ptr"
-	"math"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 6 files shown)


pkg/reaper/deployment.go
```diff
@@ -303,6 +314,33 @@ func computePodSpec(reaper *api.Reaper, dc *cassdcapi.CassandraDatacenter, initC
...
 
+	// Default security context settings
+	defaultPodSecurityContext := &corev1.PodSecurityContext{
+		RunAsNonRoot: ptr.To(true),
+		FSGroup:      ptr.To[int64](1000),
+	}
+
+	defaultContainerSecurityContext := &corev1.SecurityContext{
+		RunAsNonRoot:             ptr.To(true),
+		RunAsUser:                ptr.To[int64](1000),
```
adejanovski, line 325: **Question:** was there a reason why you didn't set `RunAsGroup` as well here? Seems like something we want to set for better default security as well.

Miles-Garnsey: I"ve added that now, no reason why not. Although given that it has RunAsNonRoot I think that means it was all already non-root uids across user, group and fs.

PR #1526, opened 2025-05-07
Update to cass-operator v1.24.0, fix bunch of test framework issues

### LINKED ISSUE #1519: Upgrade cass-operator to v1.24.0 (0.57.0)
**What is missing?**
cass-operator v1.24.0 was released and needs to be upgraded in k8ssandra-operator

**Why do we need it?**
New features and bugfixes we need.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-326) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-326

burmanm:
**What this PR does**:
* Updates cass-operator version to 1.24.0
* Updates Kubernetes versions, controller-gen versions etc
* Fix CRD validations
* Fixes test framework issues
* Fix Medusa tests after they were incorrectly using the CRD
* Fix secretController tests to actually test something (they previously waited 3 nanoseconds and then said "everything is fine" without testing anything)
* Fix linting issues

Improve speed of certain envtests, such as replication from 193s -> 36s, stargate from 30s -> 19s, etc. Some larger, some smaller, in any case minutes of difference in total time to run make test. More improvements will be needed, but this is a first step and this PR also fixes some of the tests that dependent on running slow instead of correctly checking what's happening (such as Eventually not actually having Eventually semantics).

Fixes #1519 

**Checklist**
- [ ] Changes manually tested


controllers/medusa/medusabackupjob_controller_test.go
```diff
@@ -263,21 +266,36 @@ func createAndVerifyMedusaBackup(dcKey framework.ClusterKey, dc *cassdcapi.Cassa
...
+					return false
+				}
+			}
+			return true
+		}, timeout, interval)
+	} else {
+		return false
 	}
 
+	/*
```
rzvoncek, line 286: Unless there's a good reason leaving this commented, we should either uncomment or remove.

controllers/medusa/medusaconfiguration_controller_test.go
```diff
@@ -39,8 +39,8 @@ func testMedusaConfigurationOk(t *testing.T, ctx context.Context, f *framework.F
 	}
 
 	t.Log("Creating medusa bucket key secret")
-	err := f.Client.Create(ctx, bucketKeySecret)
-	require.NoError(err, "failed to create medusa bucket key secret")
+	// err := f.Client.Create(ctx, bucketKeySecret)
```
rzvoncek, line 42: Here, and few lines below, perhaps we should remove the commented lines in order to not confuse a future reader.

.github/workflows/version_tests.yaml
```diff
@@ -59,12 +59,11 @@ jobs:
     strategy:
       matrix:
         k8s_version:
-          - "1.21"
-          - "1.22"
-          - "1.23"
+          - "1.31"
         serverVersion:
-          - 3.11.14
```
rzvoncek, line 66: Here we stop running the kuttl multi-version test agains 3.11.
At the same time, we still have 3.11 featured in the fixtures for e2e tests, a whole lot. I'd not be surprised if it's actually all of them.
Shouldn't we bump the fixtures too (perhaps not in this PR) ?

test/testdata/fixtures/reaper-control-plane/reaper.yaml
```diff
@@ -10,8 +10,6 @@ spec:
     resources:
       requests:
         storage: 256Mi
-  httpManagement:
```
rzvoncek, line 13: This is making Reaper interact with the cluster is repairs via JMX. It's not a problem for this PR because the Control Plane Reaper test uses Cassandra 3.11. It will not work should we ever switch this to DSE and such. Although I don't see why we'd ever do that in the tests here.

Still, I'm wondering, Is there a particular reason you're removing this?

rzvoncek (approved):
I have a few small questions here and there, but not bloking requests.

test/testdata/fixtures/reaper-control-plane/reaper.yaml
burmanm: Ah, actually.. not in k8ssandra-operator. We made the default to management-api in the other project, not here (yet). Although I think we should consider doing it in the k8ssandra-operator also. So I'll return this back.

.github/workflows/version_tests.yaml
burmanm: We should remove all 3.11 tests, but I think that could happen in another PR.

controllers/medusa/medusaconfiguration_controller_test.go
burmanm: Ye.

controllers/medusa/medusabackupjob_controller_test.go
burmanm: I can remove it.

rzvoncek (approved):

PR #1606, opened 2025-08-27
Replace use of CRD image defaults and configuration with global configuration

### LINKED ISSUE #1605: Replace use of CRD defined image versions with ImageConfig
**What is missing?**
We currently define certain defaults for image versions to be used in the CRD itself. This causes unnecessary CRD updates as well as need to override fields when using private registry for example.

**Why do we need it?**

**Environment**

* K8ssandra Operator version:

burmanm:
**What this PR does**:
WIP: Replaces the use of CRD based defaults for a ConfigMap solution that the user can modify. This is a shared solution with what cass-operator will have in the next version.

This is still in progress, missing things:
* cass-operator upgrade to get the feature
* Helm chart changes to allow adding new image sections for Reaper and Medusa

Fixes #1605 

**Checklist**
- [ ] Changes manually tested


adejanovski (commented):
**Question:** how can one disable the new image config and fallback to the "old" one?
Do we see a use case for doing this?

**Todo:** Could you add some release notes for this, that's quite a breaking change for people with overrides. Ideally we'll need some docs as well, but that can be a follow up PR.

adejanovski (approved):
Very good stuff!

I love how configuration of images will be much easier and centralized with this.
We have to follow up with some proper docs to explain how to use that feature for both kustomize and helm based installations.

bradfordcp:
When the ImageConfig changes do we hot-reload the new values or restart the operator?

PR #1647, opened 2025-12-01
Add telemetry TLS support, disable MCAC, fix log parsing

### LINKED ISSUE #1643: Ability to configure metrics agent TLS configuration
**What is missing?**
If user wants to enable TLS configuration in the agents config, the current TelemetrySpec does not allow to change it. The supported top path elements in the mgmt-api configuration are:

endpoints
relabels
extended_metrics_disabled
labels

Under endpoints we have subset `tls` also, which isn't supported either. We need to expose all of these to allow proper generation of metrics-collector.yaml

**Why do we need it?**
User is unable to set Vector and /metrics endpoint of Cassandra to use TLS for communication, unlike rest of the mgmt-api endpoints.

**Environment**

* K8ssandra Operator version:

### LINKED ISSUE #1645: If parsing of Cassandra message fails, do not crash the server-system-logger
**What happened?**
If parsing the Cassandra log fails, the VRL program in Vector will panic and the panic will cause server-system-logger container to crash. Instead, we should gracefully just output whatever was the offending line to the logs (it could be something like JVM crash for example).

**Environment**

* K8ssandra Operator version:

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* K8ssandra Operator Logs:

```
```

### LINKED ISSUE #1646: Disable MCAC by default
**What is missing?**
This is to give users some time until https://github.com/k8ssandra/k8ssandra-operator/issues/1501 is implemented (removing MCAC support entirely). For now, we will disable MCAC by default and it has to be manually enabled.

Document the change to release notes.

**Why do we need it?**

**Environment**

* K8ssandra Operator version:

burmanm:
**What this PR does**:
* Disables MCAC by default
* Fixes log parsing to avoid panics
* Adds Endpoint support for telemetry to allow TLS configuration

Fixes #1643 
Fixes #1646 
Fixes #1645 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -832,16 +832,12 @@ func createSingleDatacenterClusterMgmtAuth(t *testing.T, ctx context.Context, na
 
 	t.Log("check that the K8ssandraCluster was created")
 	k8ssandra := &api.K8ssandraCluster{}
-	kcKey := types.NamespacedName{Namespace: namespace, Name: "test"}
+	kcKey := types.NamespacedName{Namespace: namespace, Name: "mgmt-auth"}
 	err := f.Client.Get(ctx, kcKey, k8ssandra)
 	require.NoError(err, "failed to get K8ssandraCluster in namespace %s", namespace)
 
 	dcKey := framework.ClusterKey{K8sContext: f.DataPlaneContexts[0], NamespacedName: types.NamespacedName{Namespace: namespace, Name: "dc1"}}
 	checkDatacenterReady(t, ctx, dcKey, f)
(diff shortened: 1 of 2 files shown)


controllers/reaper/vector.go
```diff
@@ -75,12 +74,18 @@ target = "stdout"
 		vectorConfigToml = telemetry.BuildCustomVectorToml(telemetrySpec)
 	}
 
-	var scrapeInterval int32 = vector.DefaultScrapeInterval
+	var scrapeInterval int32 = 30
```
adejanovski: **Question:** why switching from a constant to a hardcoded value?

PR #1721, opened 2026-04-27
Do not retain Reaper's PVC once if Reaper uses local storage and gets deleted

### LINKED ISSUE #1720: Delete Reaper's PVCs if Reaper runs as STS

rzvoncek:
**What this PR does**: 
Set the STS PVC deletion policy from retain to delete.

Fixes #1720

**Checklist**
- [x] Changes manually tested


CHANGELOG/CHANGELOG-1.32.md
```diff
@@ -15,4 +15,5 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
-* [CHANGE] (#1731)[https://github.com/k8ssandra/k8ssandra-operator/issues/1713] Bump Medusa to 0.28.0
\ No newline at end of file
+* [CHANGE] (#1731)[https://github.com/k8ssandra/k8ssandra-operator/issues/1713] Bump Medusa to 0.28.0
```
burmanm: issue: This link format is incorrect, it should `[#1731](..url..)`. I fixed this line on my other PR, but to avoid conflicts, these should be fixed here also.

pkg/reaper/deployment.go
```diff
@@ -462,6 +462,15 @@ func computeVolumeClaims(reaper *api.Reaper) []corev1.PersistentVolumeClaim {
 	return vcs
 }
 
+func computePvcRetentionPolicy(sts *appsv1.StatefulSet, reaper *api.Reaper) {
+	if reaper.Spec.StorageType == api.StorageTypeLocal {
```
burmanm: question: Do we need the filtering here? Do we have a case where we want to retain the PVC but still have a StatefulSet?

CHANGELOG/CHANGELOG-1.32.md
rzvoncek: Thanks, will fix.

pkg/reaper/deployment.go
rzvoncek: I don't think so. We also cannot have reaper as STS with other than local storage. I'll simplify this.

burmanm (approved):

PR #1744, opened 2026-06-10
Replicate medusa storage credentials to all contexts

ShreyashD333:
**What this PR does**:

- Create a replicated secret to replicate the medusa storage credentials to all contexts
- This medusa storage credentials secret should be referenced in MedusaConfiguration only then it will be replicated
- Mount the replicated secret on Cassandra pods
- Added validation to not specify storage secret ref in both MedusaConfiguration and K8ssandraCluster

**Checklist**
- [x] Changes manually tested


controllers/k8ssandra/secrets.go
```diff
@@ -103,3 +105,55 @@ func (r *K8ssandraClusterReconciler) reconcileReplicatedSecret(ctx context.Conte
...
+func (r *K8ssandraClusterReconciler) updateReplicatededMedusaSecretName(kc *api.K8ssandraCluster) {
+	if kc.Spec.UseExternalSecrets() {
+		return
+	}
+
+	if kc.Spec.Medusa.StorageProperties.CredentialsType == medusa.CredentialsTypeRoleBased {
+		return
+	}
+
+	kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name = kc.SanitizedName() + "-" + kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name
```
adejanovski, line 162: **Todo:** updateReplicatededMedusaSecretName does not check `StorageSecretRef.Name != ""`

If CredentialsType is not role-based and the secret ref is empty after a merge failure or edge case, the function will produce a name of <sanitizedName>- (the prefix only, no actual name). This could propagate a bogus volume name. A nil/empty guard should be added:

if kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name == "" {
    return
}

pkg/secret/replicated.go
```diff
@@ -220,3 +187,93 @@ func requiresUpdate(current, desired *replicationapi.ReplicatedSecret) bool {
...
+			replicationTargets = append(replicationTargets,
+				replicationapi.ReplicationTarget{
+					K8sContextName: dcTemplate.K8sContext,
+					Namespace:      dcTemplate.Meta.Namespace,
+					TargetPrefix:   kc.SanitizedName() + "-",
+					DropLabels:     []string{medusaApi.MedusaStorageSecretIdentifierLabel},
+				},
+			)
+		}
+	}
```
adejanovski, line 203: **Issue:** ReconcileMedusaReplicatedSecret adds DC-specific replication targets only when `dcTemplate.K8sContext != "" || dcTemplate.Meta.Namespace != ""`.
We can have a K8ssandraCluster that defines a single DC in a different namespace than the K8ssandraCluster namespace (or multiple ones), still in the control plane. In that case the k8sContext will not be defined, but the DC will live in a namespace that is different than the kc.Namespace.

It doesn't seem like we have the right coverage here in terms of replication to handle all cases:
- DCs in the same namespace as the kc namespace, running in the control plane (`k8sContext == ""`)
- DCs in a different namespace as the kc namespace, running in the control plane (`k8sContext == ""`)
- DCs in the same namespace as the kc namespace, running in a dataplane (`k8sContext != ""`)
- DCs in a different namespace as the kc namespace, running in a dataplane (`k8sContext != ""`)

One of the missing scenarios above is if `dcTemplate.Meta.Namespace == ""` (the DC goes into the same namespace as the kc) but the DC is running in a dataplane (`k8sContext != ""`).

In general, I think you should not specify the namespace using a single variable in the replication target but rather use something like this:
`utils.FirstNonEmptyString(dcTemplate.Meta.Namespace, kc.Namespace)`

If the dc doesn't have an explicit namespace defined, then use the namespace from the K8ssandraCluster object.

controllers/k8ssandra/secrets.go
```diff
@@ -103,3 +105,55 @@ func (r *K8ssandraClusterReconciler) reconcileReplicatedSecret(ctx context.Conte
...
+
+	if err := secret.ReconcileMedusaReplicatedSecret(ctx, r.Client, r.Scheme, desiredKc, logger); err != nil {
+		logger.Error(err, "Failed to reconcile Medusa ReplicatedSecret")
+		return result.Error(err)
+	}
+
+	return result.Continue()
+}
+
+func (r *K8ssandraClusterReconciler) updateReplicatededMedusaSecretName(kc *api.K8ssandraCluster) {
```
adejanovski:
```suggestion
func (r *K8ssandraClusterReconciler) updateReplicatedMedusaSecretName(kc *api.K8ssandraCluster) {
```

controllers/k8ssandra/medusa_reconciler.go
```diff
@@ -50,6 +50,11 @@ func (r *K8ssandraClusterReconciler) reconcileMedusa(
 			return result.Error(mergeResult.GetError())
 		}
 
+		// Update the storage secret reference name with replicated secret name
+		if kc.Spec.Medusa.MedusaConfigurationRef.Name != "" {
+			r.updateReplicatededMedusaSecretName(kc)
```
adejanovski:
```suggestion
			r.updateReplicatedMedusaSecretName(kc)
```

controllers/k8ssandra/secrets.go
```diff
@@ -103,3 +105,55 @@ func (r *K8ssandraClusterReconciler) reconcileReplicatedSecret(ctx context.Conte
...
+
+	// if MedusaConfigurationRef is not set, then we don't need to replicate the medusa secret
+	if kc.Spec.Medusa == nil || kc.Spec.Medusa.MedusaConfigurationRef.Name == "" {
+		return result.Continue()
+	}
+
+	if kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name != "" {
+		logger.Error(nil, "Both Medusa Configuration Reference and Storage Secret Reference cannot be specified at same time")
+		return result.Error(fmt.Errorf("both Medusa Configuration Reference and Storage Secret Reference cannot be specified at same time"))
+	}
```
adejanovski, line 122: **Todo:** The mutual exclusion of MedusaConfigurationRef and StorageSecretRef is enforced at reconcile time and returns a result.Error(...) which will cause repeated reconcile failures with an error in the controller logs. This constraint is arguably user-facing and should ideally be enforced at the admission webhook level. Without it, users will only discover the conflict when the cluster fails to reconcile, not when they apply the manifest.
New checks can be added [here](https://github.com/k8ssandra/k8ssandra-operator/blob/2386-medusa-configurations-replications/apis/k8ssandra/v1alpha1/k8ssandracluster_webhook.go#L288).

test/e2e/medusa_test.go
```diff
@@ -375,3 +485,19 @@ func createMedusaConfiguration(t *testing.T, ctx context.Context, namespace stri
...
+		Name:      kcName + "-medusa-storage-credentials",
+	}
+	replicatedSecretKey := framework.ClusterKey{
+		NamespacedName: replicatedSecretName,
+		K8sContext:     dcKey.K8sContext,
+	}
+	repSec := &replicationapi.ReplicatedSecret{}
+	err := f.Get(ctx, replicatedSecretKey, repSec)
+	require.NoError(t, err)
+	require.Equal(t, "test", repSec.Labels[k8ssandraapi.K8ssandraClusterNameLabel])
```
adejanovski: **Todo:** use `kcName` instead of a hardcode `"test"` here, that will make this test proof to future changes in the cluster name.

test/testdata/fixtures/multi-dc-medusa/k8ssandra.yaml
adejanovski, line 26: **todo:** we'll need to modify the namespace for this fixture so that we cover the case where the DC namespace is different than the KC one.
That'll probably require some changes in how k8ssandra-operator is deployed, which will need to be cluster scope.

adejanovski (changes requested):
Here are some initial comments to move things forward.
I'm currently testing this branch and will report back if I have new findings.

controllers/k8ssandra/secrets.go
ShreyashD333: done

pkg/secret/replicated.go
ShreyashD333: done, removed the k8scontext and Namespace check, So the replication targets include both CP and DP and all namespaces

controllers/k8ssandra/secrets.go
ShreyashD333: done

controllers/k8ssandra/medusa_reconciler.go
ShreyashD333: done

controllers/k8ssandra/secrets.go
ShreyashD333: done, moved to webhook

test/e2e/medusa_test.go
ShreyashD333: corrected

test/testdata/fixtures/multi-dc-medusa/k8ssandra.yaml
ShreyashD333: done, for this TC, k8ssandra-operator is deployed with cluster scope

controllers/k8ssandra/secrets.go
```diff
@@ -103,3 +105,55 @@ func (r *K8ssandraClusterReconciler) reconcileReplicatedSecret(ctx context.Conte
...
+	if kc.Spec.UseExternalSecrets() {
+		return
+	}
+
+	if kc.Spec.Medusa.StorageProperties.CredentialsType == medusa.CredentialsTypeRoleBased {
+		return
+	}
+
+	kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name = kc.SanitizedName() + "-" + kc.Spec.Medusa.StorageProperties.StorageSecretRef.Name
+}
```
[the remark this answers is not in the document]
ShreyashD333: not required as updateReplicatededMedusaSecretName is called if Medusa is not nil

controllers/k8ssandra/secrets.go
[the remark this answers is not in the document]
ShreyashD333: moved to webhook

controllers/k8ssandra/secrets.go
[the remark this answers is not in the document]
ShreyashD333: corrected

.github/workflows/kind_e2e_tests.yaml
```diff
@@ -148,7 +148,7 @@ jobs:
         run: make IMG="k8ssandra/k8ssandra-operator:${{ needs.build_image.outputs.image_tag }}" KIND_NODE_VERSION="${{ matrix.kind_node_version }}" create-kind-cluster kind-load-image cert-manager nginx-kind
       - name: Run e2e test ( ${{ matrix.e2e_test }} )
         run: |
-          e2e_test="TestOperator/${{ matrix.e2e_test }}"
+          e2e_test="TestOperator/${{ matrix.e2e_test }}$"
```
adejanovski, line 151: **question:** what's the `$` here for? Ensure we don't have issues with multiple tests having the same starting characters which could have them executed together?

adejanovski (approved):
LGTM, thanks!

.github/workflows/kind_e2e_tests.yaml
ShreyashD333: Correct, if you check the any workflow apart from this PR, the CreateSingleDatacenterCluster and CreateSingleDatacenterClusterMgmtAuth were running in same job, and CreateSingleDatacenterClusterMgmtAuth runs in its own separate job as well.

PR #1748, opened 2026-06-25
Set owner reference on co-located CassandraDatacenter (#952)

### LINKED ISSUE #952: Show resources created by the operator in ArgoCD
**What is missing?**
When using ArgoCD ideally we should see all resources created by the operator in argocd

**Why do we need it?**
Debugging

**Environment**

* K8ssandra Operator version:
1.5.1 

Related to https://github.com/zalando/postgres-operator/issues/1766

https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#applications

Thanks!

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/K8OP-241) by [Unito](https://www.unito.io)
┆Issue Number: K8OP-241

Ranexia04:
**What this PR does**:

When the operator creates a `CassandraDatacenter`, it now sets the owning `K8ssandraCluster` as an owner reference on the DC — but **only when the DC is co-located**: same namespace *and* same Kubernetes cluster as the `K8ssandraCluster`.

This lets GitOps tools such as ArgoCD render the `CassandraDatacenter` as a child of the `K8ssandraCluster` in their resource tree, instead of treating it as an orphaned, prunable resource (the out-of-sync / "trash-bin" behavior reported in the issue).

As discussed in #952, owner references cannot cross namespace or cluster boundaries — a cross-namespace owner reference is [treated as absent by the garbage collector](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications) and would make the dependent eligible for deletion. Since the operator supports cluster-scoped and multi-cluster deployments where the DC may live in a different namespace/cluster than its `K8ssandraCluster`, the reference is set only when both of these hold:
- **Same cluster** — the DC is created with the local client (`dcConfig.K8sContext == ""`).
- **Same namespace** — `dc.Namespace == kc.Namespace`.

For everything else, behavior is unchanged: cleanup keeps relying on the existing label- and finalizer-based mechanism.

**Implementation notes**:
- A non-controller owner reference (`controllerutil.SetOwnerReference`) is used rather than `SetControllerReference`. Both kinds of owner reference trigger GC cascade when the owner is deleted, so the choice is not about avoiding cascade; ArgoCD builds its tree from owner references regardless of the `controller` flag. The operator already owns the DC lifecycle (including graceful decommission) via the `K8ssandraClusterFinalizer`, so the non-controller reference avoids the extra `BlockOwnerDeletion` semantics and the single-controller constraint, which aren't needed here for a stateful resource.
- The owner reference is added before the DC's resource-hash annotation is computed, so existing co-located clusters pick it up automatically on the next reconcile (a metadata-only update — no pod restarts).

Fixes #952

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG/CHANGELOG-1.33.md
+++ b/CHANGELOG/CHANGELOG-1.33.md
@@ -15,5 +15,6 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+* [ENHANCEMENT] [#952](https://github.com/k8ssandra/k8ssandra-operator/issues/952) Set an owner reference on the CassandraDatacenter pointing to the K8ssandraCluster when they are co-located (same namespace and cluster). This improves visibility in tools like ArgoCD. Cross-namespace/cluster deployments are unaffected and keep relying on labels and finalizers.
 * [CHANGE] Update cass-operator to v1.31.0
 * [CHANGE] Bump k8ssandra-client to v0.8.13, medusa to v0.29.0 and reaper to v4.2.4
--- a/CHANGELOG/CHANGELOG-1.33.md
+++ b/CHANGELOG/CHANGELOG-1.33.md
@@ -19,6 +19,8 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 * [CHANGE] Bump k8ssandra-client to v0.8.13, medusa to v0.29.0 and reaper to v4.2.4
 * [CHANGE] Bump cassandra-medusa to v0.30.0
 * [ENHANCEMENT] [#1760](https://github.com/k8ssandra/k8ssandra-operator/issues/1760) Allow setting Medusa's chunk size
+* [ENHANCEMENT] [#952](https://github.com/k8ssandra/k8ssandra-operator/issues/952) Set an owner reference on the CassandraDatacenter pointing to the K8ssandraCluster when they are co-located (same namespace and cluster). This improves visibility in tools like ArgoCD. Cross-namespace/cluster deployments are unaffected and keep relying on labels and finalizers.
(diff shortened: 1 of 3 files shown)


Ranexia04:
Hi @adejanovski, can i get a review? Also, the link the signing the CLA is not working. Thanks in advance

vavsab:
@rzvoncek Could you please merge and release this PR?

CHANGELOG/CHANGELOG-1.33.md
```diff
@@ -15,5 +15,6 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 
 ## unreleased
 
+* [ENHANCEMENT] [#952](https://github.com/k8ssandra/k8ssandra-operator/issues/952) Set an owner reference on the CassandraDatacenter pointing to the K8ssandraCluster when they are co-located (same namespace and cluster). This improves visibility in tools like ArgoCD. Cross-namespace/cluster deployments are unaffected and keep relying on labels and finalizers.
```
rzvoncek, line 22: nit: please move this under the `[CHANGE]`s, so that it matches the order above.

rzvoncek (changes requested):
Hi, could you please polish the change log?

CHANGELOG/CHANGELOG-1.33.md
Ranexia04: Done - moved the entry under the `[CHANGE]` lines so it matches the order in the header (and added the missing trailing newline). Amended into the existing single commit.

Ranexia04:
@rzvoncek is it ok now?

rzvoncek:
Yes, okay now. Thank you. We'll ship this with the next release.

PR #1758, opened 2026-07-16
Trigger k8ssandrcluster reconcile on referenced MedusaConfiguration update

### LINKED ISSUE #1415: MedusaConfiguration changes aren't propagated to pods
When configuring Medusa through a MedusaConfiguration ref, the expectation is that any change to the referenced object will propagate to the pods which get configured with it.

In the latest versions, we're not propagating the changes to the config maps anymore, which makes it impossible to update the configurations after the initial creation of the cluster.

We need to get the configmaps updated when the referenced MedusaConfiguration gets updated.

ShreyashD333:
**What this PR does**:
Add watcher for medusaconfiguration in k8ssandracontroller , so that any change to referenced medusaconfiguration will trigger k8ssandracluster reconcile and it will update medua configurations configmap. Indexed k8ssandracluster againt medusaConfigurationReference for efficient filtering in watcher.

Fixes #1415 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/controllers/k8ssandra/k8ssandracluster_controller.go
+++ b/controllers/k8ssandra/k8ssandracluster_controller.go
@@ -27,6 +27,7 @@ import (
 	cassdcapi "github.com/k8ssandra/cass-operator/apis/cassandra/v1beta1"
 	cassimages "github.com/k8ssandra/cass-operator/pkg/images"
 	api "github.com/k8ssandra/k8ssandra-operator/apis/k8ssandra/v1alpha1"
+	medusaapi "github.com/k8ssandra/k8ssandra-operator/apis/medusa/v1alpha1"
 	reaperapi "github.com/k8ssandra/k8ssandra-operator/apis/reaper/v1alpha1"
 	stargateapi "github.com/k8ssandra/k8ssandra-operator/apis/stargate/v1alpha1"
 	"github.com/k8ssandra/k8ssandra-operator/pkg/cassandra"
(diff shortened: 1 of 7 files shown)


controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -51,6 +52,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+// indexK8ssandraClusterByMedusaConfigRef registers a field index on K8ssandraCluster so that
+// the controller can efficiently query which clusters reference a given MedusaConfiguration.
+func indexK8ssandraClusterByMedusaConfigRef(mgr ctrl.Manager) error {
+	return mgr.GetFieldIndexer().IndexField(
+		context.Background(),
```
burmanm: issue: This is wrong context. When creating these caches, you need to set them using the context you have available in the main.go (the main context) or a subcontext of that context. Otherwise errors and cancellations will not be applied correctly.

It might make sense to move this code to main.go or just call this code directly from main.go also if you need to more easily use that context. In that case just export the function.

controllers/k8ssandra/k8ssandracluster_controller.go
```diff
@@ -251,6 +292,8 @@ func (r *K8ssandraClusterReconciler) SetupWithManager(mgr ctrl.Manager, clusters
 		handler.EnqueueRequestsFromMapFunc(clusterLabelFilter))
 	cb = cb.Watches(&discoveryv1.EndpointSlice{},
 		handler.EnqueueRequestsFromMapFunc(endpointsFilter))
+	cb = cb.Watches(&medusaapi.MedusaConfiguration{},
```
burmanm, line 295: suggestion: Since MedusaConfiguration has Status field, you should have something like `predicate.GenerationChangedPredicate{}` here to prevent unnecessary reconciles.

controllers/k8ssandra/k8ssandracluster_controller.go
ShreyashD333: Thanks, I am now using context from main, did not move this function to main.go because it would create import cycle.

controllers/k8ssandra/k8ssandracluster_controller.go
ShreyashD333: done

rzvoncek (approved):
Manually tested, works as expected.

PR #1761, opened 2026-08-04
Allow setting Medusa's chunk size

### LINKED ISSUE #1760: Alow setting medusa's chunk size via k8ssandra-operator
We are improving medusa by fixing recognition of config.multipart_chunk_size in the other two storage implementations (non-s3 ones).

This setting is already accepted by medusa, it's just not applied to the providers. But the k8ssandra-operator does not support it at all - it's not present in the storage properties, there's only MultiPartUploadThreshold.

So we need to add the MultiPartChunkSize and make sure it goes through both the k8ssandra-cluster spec, and also through MedusaConfiguration.

rzvoncek:
**What this PR does**: Allows setting Medusa's chunk size via the `storage` CRD.

Fixes #1760

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/medusa/v1alpha1/medusa_types.go
+++ b/apis/medusa/v1alpha1/medusa_types.go
@@ -86,6 +86,11 @@ type Storage struct {
 	// +optional
 	MultiPartUploadThreshold int `json:"multiPartUploadThreshold,omitempty"`
 
+	// Chunk size in bytes for multipart uploads.
+	// Defaults to 0 (medusa uses its own default).
+	// +optional
+	MultiPartChunkSize int `json:"multiPartChunkSize,omitempty"`
+
 	// Host to connect to for the storage backend.
 	// +optional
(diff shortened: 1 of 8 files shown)


apis/medusa/v1alpha1/medusa_types.go
```diff
@@ -86,6 +86,11 @@ type Storage struct {
 	// +optional
 	MultiPartUploadThreshold int `json:"multiPartUploadThreshold,omitempty"`
 
+	// Chunk size in bytes for multipart uploads.
+	// Defaults to 0 (medusa uses its own default).
```
burmanm: suggestion: This `// Defaults to 0 (medusa uses its own default).` is slightly confusing. 0 is indeed the default value for int in Golang, but I don't see how it generates any value mentioning here. Could just be "leaving empty uses medusa's default value" or something. Or if unset..

burmanm (approved):

PR #1791, opened 2026-09-10
feat(chart): add affinity and topologySpreadConstraints support to the k8ssandra-operator chart

### LINKED ISSUE #1790: Helm chart: no way to set affinity or topologySpreadConstraints on the operator Deployment
### What is missing?

The `k8ssandra-operator` Helm chart exposes `nodeSelector` and `tolerations` for
the operator Deployment, but there is no way to set `affinity` or
`topologySpreadConstraints`.

The chart already supports running more than one operator pod via
`replicaCount`, and the operator uses leader election, so a standby replica is a
supported configuration. However, with no way to express anti-affinity, nothing
prevents the scheduler from placing all operator replicas on the same node.

In our clusters we run `replicaCount: 2` in all four environments, and in both
production clusters both operator pods are currently scheduled on the same node.
The `replicaCount: 2` therefore provides no protection against node loss: a
single node failure or drain takes down every operator replica at once. Because
the chart also installs fail-closed validating/mutating webhooks
(`k8ssandra-operator-validating-webhook-configuration` and the mutating
equivalent), losing all replicas simultaneously also blocks admission of
K8ssandraCluster resources until a pod is rescheduled.

### Why is this needed?

Without `affinity`, `replicaCount > 1` cannot be made genuinely
highly-available. `nodeSelector` and `tolerations` alone cannot express "keep
these replicas on different nodes".

This is about the **operator Deployment's own** scheduling. It is unrelated to
issues #74 and #445, which concern replacing affinity rules with topology spread
constraints for Cassandra **rack** topology in the CassandraDatacenter spec.

### Suggested solution

Add `affinity` and `topologySpreadConstraints` values, rendered into the
operator Deployment's pod spec in the same conditional style already used for
`nodeSelector` and `tolerations`. Both default to empty so existing
installations render byte-identical output.

Ranexia04:
**What this PR does**:

Adds `affinity` and `topologySpreadConstraints` values to the
`k8ssandra-operator` chart, rendered into the operator Deployment's pod spec.

The chart already exposes `nodeSelector` and `tolerations`, and already supports
`replicaCount > 1` (the operator uses leader election, so a standby replica is a
supported setup). But neither `nodeSelector` nor `tolerations` can express "keep
these replicas on different nodes", so nothing stops the scheduler from putting
every operator pod on the same node.

That is what happens in practice for us. We run `replicaCount: 2` in all four of
our environments, and in both production clusters both operator pods are
currently on the same node, so the second replica buys nothing against node
loss. It also means a single node drain can take out both replicas of the
chart's fail-closed webhooks at once, blocking K8ssandraCluster admission until
a pod is rescheduled elsewhere.

Both new values default to empty (`{}` / `[]`) and are wrapped in the same
`{{- if }}` conditional style as the existing `nodeSelector` and `tolerations`,
so existing installations render byte-identical output.

Example use:

```yaml
replicaCount: 2
affinity:
  podAntiAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        podAffinityTerm:
          labelSelector:
            matchLabels:
              app.kubernetes.io/name: k8ssandra-operator
          topologyKey: kubernetes.io/hostname
```

Note this concerns the **operator Deployment's own** scheduling. It is unrelated
to #74 and #445, which are about replacing affinity rules with topology spread
constraints for Cassandra **rack** topology inside the CassandraDatacenter spec.

Fixes #1790

**Checklist**
- [x] Changes manually tested

Notes on the unchecked boxes and on testing:

- **Manual testing**: verified with `helm template` that (a) with default values
  neither `affinity` nor `topologySpreadConstraints` appears in the rendered
  Deployment, (b) with both set they render correctly into the pod spec, and
  (c) the full chart still renders as valid YAML (31 manifests parsed). Rendered
  against chart deps `cass-operator` 0.67.1 and `k8ssandra-common` 0.29.2.
- **Automated tests**: left unchecked as I could not find an existing test layer
  that covers chart values plumbing — the e2e suites exercise the operator
  itself rather than rendered chart output. Happy to add coverage if you can
  point me at the right place.
- **Documentation**: the new values carry `# --` helm-docs annotations matching
  the surrounding entries. I deliberately did **not** regenerate
  `charts/k8ssandra-operator/README.md`: its values table is currently stale
  (it is missing `nodeSelector`, `tolerations`, `client.*` and
  `disableCrdUpgraderJob`, and its `image.tag` row says `v1.0.0` while
  `values.yaml` says `latest`), so a regeneration would produce a large diff
  unrelated to this change. Please tell me if you would rather I regenerate it
  here, or add just the two new rows by hand.
- **CLA**: not yet signed at the time of opening; I will get this done.

**Disclosure**: this change was written with AI assistance (Claude), reflected
in the commit's `Co-Authored-By` trailer. I have reviewed it and can speak to
the reasoning behind it.

### DIFF
--- a/CHANGELOG/CHANGELOG-1.34.md
+++ b/CHANGELOG/CHANGELOG-1.34.md
@@ -14,3 +14,5 @@ Changelog for the K8ssandra Operator, new PRs should update the `unreleased` sec
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
 ## unreleased
+
+* [ENHANCEMENT] Add `affinity` and `topologySpreadConstraints` support to the k8ssandra-operator chart, allowing the operator pods to be spread across nodes when `replicaCount` is greater than 1
--- a/charts/k8ssandra-operator/templates/deployment.yaml
+++ b/charts/k8ssandra-operator/templates/deployment.yaml
@@ -106,6 +106,12 @@ spec:
       {{- if .Values.tolerations }}
       tolerations: {{ toYaml .Values.tolerations | nindent 8 }}
       {{- end }}
+      {{- if .Values.affinity }}
+      affinity: {{ toYaml .Values.affinity | nindent 8 }}
+      {{- end }}
+      {{- if .Values.topologySpreadConstraints }}
+      topologySpreadConstraints: {{ toYaml .Values.topologySpreadConstraints | nindent 8 }}
+      {{- end }}
       terminationGracePeriodSeconds: 10
       volumes:
       - configMap:
--- a/charts/k8ssandra-operator/values.yaml
+++ b/charts/k8ssandra-operator/values.yaml
@@ -85,6 +85,16 @@ nodeSelector: {}
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
 ##
 tolerations: []
+# -- Affinity rules for operator pod assignment. Useful with replicaCount > 1 to
+# keep the operator pods on separate nodes.
+## Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+##
+affinity: {}
+# -- Topology spread constraints for operator pod assignment. Useful with
+# replicaCount > 1 to spread the operator pods across nodes or zones.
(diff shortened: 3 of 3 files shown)


bradfordcp:
Thank you for the contribution, this looks like a great addition to the project. I'll defer to some of the other maintainers to review, but I don't see any issues with my first glance.

xadips:
seems I duplicated this PR but then I found yours, closed mine but opened one for cass-operator as well https://github.com/k8ssandra/k8ssandra/pull/1728

PR #1792, opened 2026-09-13
Route jvmOptions.gc to jvm17-server-options on Cassandra 5.x

### LINKED ISSUE #1566: ZGC not being set when specified in manifest
**What happened?**
When I specify gc: ZGC in the K8ssandraCluster manifest, it looks like the GC is being set back to G1GC in the JVM options once the K8ssandraCluster comes up.

I expected the GC to be set to ZGC and see -XX:UseZGC in the JVM arguments instead of -XX:UseG1GC. 

Create a K8ssandraCluster using k8ssandra/cass-management-api:5.0.3-ubi-v0.1.104 as the server image and set gc: ZGC.

**Environment**

* K8ssandra Operator version::v.1.21.1
* Cass Operator version:v1.23.2
* K8ssandra-Client:v0.7.0

Client Version: v1.32.0
Kustomize Version: v5.5.0
Server Version:v1.29.4

* Manifests:

```
kind: K8ssandraCluster
metadata:
  name: c5jdk17-test
spec:
  cassandra:
    serverVersion: "5.0.3"
    serverImage: docker.io/k8ssandra/cass-management-api:5.0.3-ubi-v0.1.104
    config:
      jvmOptions:
        additionalOptions:
        - -Xmx12G # max heap size to 12G
        - -Xms12G # min heap size to 12G
        gc: ZGC
    resources:
      limits:
        cpu: 8000m
        memory: 24Gi
      requests:
        cpu: 2000m
        memory: 24Gi
    datacenters:
    - metadata:
        name: dc1
      size: 3
      storageConfig:
        cassandraDataVolumeClaimSpec:
          storageClassName: storageclass-cassandra
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 5Gi
```

### LINKED ISSUE #1766: `gc: Shenandoah` silently ignored on Cassandra 5.0 — written to `jvm11-server.options` not `jvm17-server.options`
(this was AI-assited but I think the finding is accurate)

### What happened

Setting `spec.cassandra.datacenters[].config.jvmOptions.gc: Shenandoah` with `serverVersion: 5.0.8` passes CRD validation and produces no warning, event, or log line — but the cluster runs G1.

The rendered config contains both collectors, in two different files:

```
/opt/cassandra/conf/jvm11-server.options:24:-XX:+UseShenandoahGC
/opt/cassandra/conf/jvm17-server.options:1:-XX:+UseG1GC
```

`bin/cassandra` selects exactly one JDK-specific options file based on the detected Java version. The image runs JDK 17:

```
openjdk version "17.0.20" 2026-07-21 LTS
```

so `jvm17-server.options` is selected and the Shenandoah flag is never read. Effective flags confirm it:

```
bool UseG1GC          = true   {product} {command line}
bool UseShenandoahGC  = false  {product} {default}
```

Note `UseShenandoahGC` is present in `VM.flags -all`, so the JVM does support the collector — this is not a missing-JDK-feature issue.

### Why it's silent

The `gc` field's own documentation scopes it to two targets: Cassandra 3.11 in `jvm.options`, Cassandra 4.0 in `jvm11-server.options`. There is no mapping for 5.0 / `jvm17-server.options`, but the enum still accepts `Shenandoah`. The value is written to the 4.0 target while the 5.0 target receives the `G1GC` default.

Because `G1GC` is also the field default, this produces **no diff and no rollout signal** — applying the change appears to succeed and nothing indicates the value was inert.

### Reproduction

```yaml
kind: K8ssandraCluster
spec:
  cassandra:
    serverVersion: "5.0.8"
    datacenters:
      - metadata:
          name: dc1
        size: 3
        config:
          jvmOptions:
            heap_initial_size: 30G
            heap_max_size: 30G
            gc: Shenandoah
```

Then, in the `cassandra` container:

```bash
CPID=$(pgrep -f CassandraDaemon)
java -version
grep -nE 'UseG1GC|Shenandoah' $CASSANDRA_CONF/jvm*-server.options
jcmd $CPID VM.flags -all | grep -E 'UseG1GC|UseShenandoahGC'
```

### Environment

- k8ssandra-operator CRDs v1.30
- `serverVersion: 5.0.8`, JDK 17.0.20
- Custom management-api image (AxonOps build), rebuilt on a current JDK base
- Config is rendered by `server-config-init` at pod start — all three `jvm*-server.options` files share an mtime, so they are generated rather than baked into the image

### Which layer

This should show whether k8ssandra-operator is targeting the options file itself, or forwarding `garbage_collector` for the config builder / cass-config-definitions to route.

### Expected

Either route `gc` to the version-appropriate options file for 5.x, or reject unsupported values at admission for the given `serverVersion`. Silently accepting an enum value that has no effect is the core problem — a validating webhook rejection would have surfaced this immediately.

### Additional notes

There is currently **no CRD field that can write to `jvm17-server.options` at all** (`additionalJvm8ServerOptions`, `additionalJvm11ServerOptions`, and `additionalJvmServerOptions` exist; there is no `additionalJvm17ServerOptions`). On Cassandra 5.x with JDK 17 this means there is no supported path to set any Java-17-specific JVM option, and no workaround for this bug that goes through the CRD's JVM options fields.

The same applies to every field documented against `jvm11-server.options` — `gc`, the five `gc_g1_*` fields, `jdk_allow_attach_self`, `netty_try_reflection_set_accessible` — all of which appear to be inert on 5.0 with JDK 17.

Secondary hazard: the stale `-XX:+UseShenandoahGC` left in `jvm11-server.options` is live config. If an operator later pins a JDK 11 image, the collector silently changes on a production DC with nothing in the manifest diff to explain it.

### cass-operator version

1.32.6

### Kubernetes version

1.36

### Method of installation

_No response_

### Anything else we need to know?

_No response_

KR-Ravindra:
**What this PR does**:

#### Problem
On Cassandra 5.x the JVM runs on Java 17 and reads `jvm17-server.options`, but `jvmOptions.gc` was rendered into `jvm11-server-options` for every Cassandra `>=4.x`. The chosen collector (ZGC, Shenandoah, ...) ended up in a file the JVM never reads, so the cluster silently ran G1GC while the CR claimed otherwise. Passes CRD validation with no event or warning.

#### Root cause
`apis/k8ssandra/v1alpha1/cassandraconfig_types.go:319` - the `cass-config` tag on `JvmOptions.GarbageCollector` maps `>=4.x` to `jvm11-server-options/garbage_collector`, as described by @burmanm in #1566.

#### Fix
Split the Cassandra constraint in that tag: `^4.x` keeps `jvm11-server-options`, `>=5.x` goes to `jvm17-server-options`. HCD and DSE mappings are unchanged. k8ssandra-client already renders `jvm17-server-options.garbage_collector` into `jvm17-server.options` (`pkg/config/builder.go`, `createServerJVMOptions`), so no change is needed on that side.

The `gc` field description (which #1766 pointed at, it only mentioned Cassandra 4.0 / `jvm11-server.options`) now lists Cassandra 5.0 / `jvm17-server.options`; the CRD in `config/crd/bases` and `charts/k8ssandra-operator/crds` is regenerated with `make manifests`. The `docs/` CRD reference is left to the nightly update workflow.

Not covered here: the G1 tuning fields (`gc_g1_*`), `netty_try_reflection_set_accessible` and `jdk_allow_attach_self` have the same `>=4.x -> jvm11-server-options` routing, but k8ssandra-client has no alias table for `jvm17-server.options`, so redirecting them would still be a no-op until the client learns those keys. Kept this PR to the `gc` field the issue reports.

#### How tested
Added `[5.0.0] GC` (and a `[4.1.0] GC` boundary case) to `TestCreateJsonConfig`, plus a tag-parsing case for the split constraint in `Test_parseCassConfigTag`.

Before the fix:
```
--- FAIL: TestCreateJsonConfig/[5.0.0]_GC (0.00s)
    -  (string) (len=20) "jvm17-server-options": (map[string]interface {}) (len=1) {
    +  (string) (len=20) "jvm11-server-options": (map[string]interface {}) (len=1) {
```
After the fix:
```
--- PASS: TestCreateJsonConfig/[4.0.0]_GC
--- PASS: TestCreateJsonConfig/[4.1.0]_GC
--- PASS: TestCreateJsonConfig/[5.0.0]_GC
--- PASS: Test_parseCassConfigTag
ok  	github.com/k8ssandra/k8ssandra-operator/pkg/cassandra
```
`go test ./apis/k8ssandra/v1alpha1/` (envtest), `go vet` and `go build ./...` pass; `gofmt -l` is clean. The existing `GCTests/4.0-jdk11-*` e2e jobs cover the `^4.x` half of the split.

Fixes #1566
Fixes #1766

**Checklist**
- [ ] Changes manually tested
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Checked against `main`: the `cass-config` tag on `JvmOptions.GarbageCollector` is the one burmanm pointed at in #1566; `pathForVersion` picks the first constraint that validates, so `^4.x` (>=4.0.0 <5.0.0) and `>=5.x` do not overlap; the HCD and DSE mappings are byte-identical to before; `config/crd/bases` and the chart CRD only carry the doc-comment change. On the client side `pkg/config/types.go` maps `jvm17-server-options` to `ServerOptions17`, and `createServerJVMOptions` renders `garbage_collector` for any `jvm*-server.options` file, so the new path is consumed. `go test ./pkg/cassandra/ -run 'TestCreateJsonConfig|Test_parseCassConfigTag'` passes at b49eaac. #1766 is the same bug (closed as a duplicate of #1566 in spirit), and no other PR references either issue.

One item before this leaves draft:
1. The SonarCloud quality gate fails on new-code duplication (36.1%, limit 3%). Per Sonar's per-file measures all 26 new lines in `pkg/cassandra/config_tag_test.go` are flagged as duplicated; the other files are clean. Either drop that parser case (the `[4.1.0] GC` / `[5.0.0] GC` cases in `TestCreateJsonConfig` already exercise both halves of the split constraint end to end) or reduce it to a `pathForVersion` assertion on the parsed tag, so the gate passes as it does on the recently merged PRs.

The unit/integration test job is still running.

PR #1798, opened 2026-09-21
chart: add affinity and topologySpreadConstraints support to k8ssandra-operator

### LINKED ISSUE #1790: Helm chart: no way to set affinity or topologySpreadConstraints on the operator Deployment

xadips:
**What this PR does**:

Adds `affinity` and `topologySpreadConstraints` values to the k8ssandra-operator Helm chart so the controller-manager Deployment can be given pod affinity/anti-affinity rules and topology spread constraints, the same way `nodeSelector` and `tolerations` already work.

Changes:
- `charts/k8ssandra-operator/templates/deployment.yaml`: render `affinity` and `topologySpreadConstraints` conditionally, following the existing `nodeSelector`/`tolerations` pattern.
- `charts/k8ssandra-operator/values.yaml`: add `affinity: {}` and `topologySpreadConstraints: []` defaults with helm-docs style comments and Kubernetes doc references.
- `CHANGELOG/CHANGELOG-1.34.md`: add an unreleased entry.

Default rendering is unchanged when these values are left unset.

Fixes #1790

**Checklist**
- [x] Changes manually tested


xadips:
Duplicate of https://github.com/k8ssandra/k8ssandra-operator/pull/1791, which was opened 11 days earlier against the same issue (#1790) with an equivalent fix. Closing this one in favor of that one.

